PR #5, opened 2020-06-08
add CI integration

matthiasdiener:


.github/workflows/tests.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - uses: goanpeca/setup-miniconda@v1
+      with:
+        auto-update-conda: true
+        channels: conda-forge
+    - name: Install prerequisites
+      shell: bash -l {0}
+      run: |
+        conda install pocl islpy pyopencl pytest pip
+        pip install git+https://gitlab.tiker.net/inducer/{loopy,meshmode,dagrt,leap,modepy,meshpy,grudge}
+        pip install -e .
```
inducer, line 45: This is pretty redundant with the case above. Is there a way to prevent these from becoming inconsistent?

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - name: Run tests
+      shell: bash -l {0}
+      run: |
+        pytest
```
inducer, line 49: I run `pytest` with [a few more options](https://gitlab.tiker.net/inducer/ci-support/-/blob/a531734d6ace98281a1e462f12147db56ca95ace/build-and-test-py-project.sh#L54). I feel that we'll want many of the same ones.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+  MacOS:
+    runs-on: macos-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: goanpeca/setup-miniconda@v1
+      with:
+        auto-update-conda: true
+        channels: conda-forge
```
inducer, line 39: I imagine that this will also source packages from the conda `defaults` channel, which we do not want. Maybe adding `nodefaults` might do the trick.

x-ref: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html

PR #6, opened 2020-06-09
Add unit tests

### LINKED ISSUE #4: Add unit tests
Need some example unit tests. Currently in progress in [wave-tests](https://github.com/illinois-ceesd/mirgecom/blob/wave-tests/test/test_wave.py) branch.

majosm:
Closes #4 

I found and fixed the bug that was causing inconsistency between the cosine test and the manufactured test. Now each test requires a `-1` factor in the RHS for both `u` and `v`. My algebra and implementation also now appear to agree with what I get from `sympy`.

Looks ready for review. One unresolved question: should it be merged with the tests passing (i.e., with fudge factor) or failing (without fudge factor)?

### DIFF
--- a/test/test_wave.py
+++ b/test/test_wave.py
@@ -140,14 +140,10 @@ def test_standing_wave(ctx_factory, dim, order):
         sym_v = [sym_c * sym_diff(sym_coords[i])(sym_phi) for i in range(dim)]
 
         # rhs(u part) = c*div(v)
-        # Why negative sign?
-        # sym_rhs_u = sym_c * sym_div(sym_v)
-        sym_rhs_u = -sym_c * sym_div(sym_v)
+        sym_rhs_u = sym_c * sym_div(sym_v)
         # rhs(v part) = c*grad(u)
         sym_grad_u = sym_grad(dim, sym_u)
-        # Why negative sign?
-        # sym_rhs_v = [sym_c*sym_grad_u[i] for i in range(dim)]
-        sym_rhs_v = [-sym_c*sym_grad_u[i] for i in range(dim)]
+        sym_rhs_v = [sym_c * sym_grad_u[i] for i in range(dim)]
         sym_rhs = [sym_rhs_u] + sym_rhs_v
 
         c = 2.
@@ -236,14 +232,10 @@ def test_wave_manufactured(ctx_factory, dim, order):
         sym_v = [sym_c * sym_diff(sym_coords[i])(sym_phi) for i in range(dim)]
 
         # rhs(u part) = c*div(v)
-        # Why negative sign?
-        # sym_rhs_u = sym_c * sym_div(sym_v)
-        sym_rhs_u = -sym_c * sym_div(sym_v)
+        sym_rhs_u = sym_c * sym_div(sym_v)
         # rhs(v part) = c*grad(u)
         sym_grad_u = sym_grad(dim, sym_u)
-        # Why negative sign?
-        # sym_rhs_v = [sym_c*sym_grad_u[i] for i in range(dim)]
-        sym_rhs_v = [-sym_c*sym_grad_u[i] for i in range(dim)]
+        sym_rhs_v = [sym_c * sym_grad_u[i] for i in range(dim)]
         sym_rhs = [sym_rhs_u] + sym_rhs_v
 
         c = 2.
--- a/test/test_wave.py
+++ b/test/test_wave.py
@@ -20,4 +20,262 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-# Tests go here
+import numpy as np
+import numpy.linalg as la
+import pyopencl as cl
+import pyopencl.array as cla  # noqa
+import pyopencl.clmath as clmath
+from pytools.obj_array import flat_obj_array
+
+from pyopencl.tools import (  # noqa
+        pytest_generate_tests_for_pyopencl
+        as pytest_generate_tests)
+
+import pytest
+
+import logging
+logger = logging.getLogger(__name__)
+
+@pytest.mark.parametrize("dim", [2, 3])
+@pytest.mark.parametrize("order", [2, 3, 4])
+def test_standing_wave(ctx_factory, dim, order):
+    cl_ctx = ctx_factory()
+    queue = cl.CommandQueue(cl_ctx)
+
+    from pytools.convergence import EOCRecorder
+    eoc_rec = EOCRecorder()
+
+    for n in [4, 8, 16]:
+        from meshmode.mesh.generation import generate_regular_rect_mesh
+        mesh = generate_regular_rect_mesh(
+                a=(-0.5*np.pi,)*dim,
+                b=(0.5*np.pi,)*dim,
+                n=(n,)*dim)
+
+        from grudge.eager import EagerDGDiscretization
+        discr = EagerDGDiscretization(cl_ctx, mesh, order=order)
+
+        nodes = discr.nodes().with_queue(queue)
+
+        # 2D: phi(x,y,t) = cos(omega*t-pi/4)*cos(x)*cos(y)
+        # 3D: phi(x,y,z,t) = cos(omega*t-pi/4)*cos(x)*cos(y)*cos(z)
+        # where omega = sqrt(#dims)*c
+
+        c = 2.
+        offset = np.pi/4.
+        omega = np.sqrt(dim) * c
+
+        # 2D: u = phi_t = -omega*sin(omega*t-pi/4)*cos(x)*cos(y)
+        # 3D: u = phi_t = -omega*sin(omega*t-pi/4)*cos(x)*cos(y)*cos(z)
+        u = discr.empty(queue, dtype=np.float64)
+        u[:] = -omega * np.sin(-offset)
+        for i in range(dim):
+            u = u * clmath.cos(nodes[i])
+
+        # 2D: v[0] = c*phi_x = -c*cos(omega*t-pi/4)*sin(x)*cos(y)
+        #     v[1] = c*phi_y = -c*cos(omega*t-pi/4)*cos(x)*sin(y)
+        # 3D: v[0] = c*phi_x = -c*cos(omega*t-pi/4)*sin(x)*cos(y)*cos(z)
+        #     v[1] = c*phi_y = -c*cos(omega*t-pi/4)*cos(x)*sin(y)*cos(z)
+        #     v[2] = c*phi_z = -c*cos(omega*t-pi/4)*cos(x)*cos(y)*sin(z)
+        v = []
+        for i in range(dim):
+            v_i = discr.empty(queue, dtype=np.float64)
+            v_i[:] = -c * np.cos(-offset)
+            for j in range(dim):
+                if i == j:
+                    v_i = v_i * clmath.sin(nodes[j])
+                else:
+                    v_i = v_i * clmath.cos(nodes[j])
+            v.append(v_i)
+        fields = flat_obj_array(u, v)
+
+        from mirgecom.wave import wave_operator
+        rhs = wave_operator(discr, c=c, w=fields)
+
+        # rhs(u part) = c*div(v)
+        # 2D: rhs(u part) = -2*c^2*cos(omega*t-pi/4)*cos(x)*cos(y)
+        # 3D: rhs(u part) = -3*c^2*cos(omega*t-pi/4)*cos(x)*cos(y)*cos(z)
+        expected_rhs_u = discr.empty(queue, dtype=np.float64)
+        # Why no negative sign?
+        # expected_rhs_u[:] = -omega**2 * np.cos(-offset)
+        expected_rhs_u[:] = omega**2 * np.cos(-offset)
+        for i in range(dim):
+            expected_rhs_u = expected_rhs_u * clmath.cos(nodes[i])
+
+        # rhs(v part) = c*grad(u)
+        # 2D: rhs(v part)[0] = omega*c*sin(omega*t-pi/4)*sin(x)*cos(y)
+        #     rhs(v part)[1] = omega*c*sin(omega*t-pi/4)*cos(x)*sin(y)
+        # 3D: rhs(v part)[0] = omega*c*sin(omega*t-pi/4)*sin(x)*cos(y)*cos(z)
+        #     rhs(v part)[1] = omega*c*sin(omega*t-pi/4)*cos(x)*sin(y)*cos(z)
+        #     rhs(v part)[2] = omega*c*sin(omega*t-pi/4)*cos(x)*cos(y)*sin(z)
+        expected_rhs_v = []
+        for i in range(dim):
+            rhs_v_i = discr.empty(queue, dtype=np.float64)
+            # Why negative sign?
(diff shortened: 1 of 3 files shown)


majosm:
Not sure if I should merge the CI branch into this so the tests will run, or merge this first and then merge the CI?

matthiasdiener:
> Not sure if I should merge the CI branch into this so the tests will run, or merge this first and then merge the CI?

I think the CI branch/PR depends on this being available, so I guess you should merge this first and then the CI branch.

majosm:
Also, now that the tests agree with each other, it may not be necessary to keep both. I can remove the manufactured one if it seems redundant?

test/test_wave.py
```diff
@@ -20,4 +20,266 @@
...
+            # Why negative sign?
+            # rhs_v_i[:] = omega*c * np.sin(-offset)
+            rhs_v_i[:] = -omega*c * np.sin(-offset)
+            for j in range(dim):
+                if i == j:
+                    rhs_v_i = rhs_v_i * clmath.sin(nodes[j])
+                else:
+                    rhs_v_i = rhs_v_i * clmath.cos(nodes[j])
+            expected_rhs_v.append(rhs_v_i)
+        expected_rhs = join_fields(expected_rhs_u, expected_rhs_v)
```
inducer, line 125: I'd be in favor of trying to:

- get `sympy` (or even `pymbolic`) to generate the same math
- using their output (directly, without copy-pasting) to check the solution

so that the only hardcoded formula in here is the original solution.

Why is this useful (IMO)?

Right now, I would have to go and check all the calculus here by hand. That's maybe manageable here, but for more complex manufactured solutions, it'll get old pretty quick. So if we build up some machinery here to make this type of thing easier, I think that has the potential to pay off big time later on. I'm happy to guide you through how to do this, maybe Thursday.

inducer:
> I can remove the manufactured one if it seems redundant?

Nah. Let's stick with "the more the merrier" for now.

majosm:
@matthiasdiener Do you know what this test failure means?

```pyopencl._cl.LogicError: clGetPlatformIDs failed: PLATFORM_NOT_FOUND_KHR```

**Comment by matthiasdiener** [comment]
> @matthiasdiener Do you know what this test failure means?
> 
> `pyopencl._cl.LogicError: clGetPlatformIDs failed: PLATFORM_NOT_FOUND_KHR`

I think it means that it can't find the installed pocl (e.g. missing export OCL_ICD_VENDORS). Let me check.

File: test/test_wave.py [hunk 2]
```diff
@@ -20,4 +20,277 @@
...
+    for i in range(dim):
+        grad_i = sym_diff(pmbl.var(coord_names[i]))(func)
+        grad.append(grad_i)
+    return grad
+
+
+def sym_eval_with(expr, values):
+    from pymbolic.mapper.evaluator import EvaluationMapper
+    import pymbolic.primitives as prim
+    class mapper(EvaluationMapper):
```
inducer, line 77: Classes want to be at the top level

test/test_wave.py
```diff
@@ -20,4 +20,277 @@
...
+                return np.sin(self.rec(par))
+            elif expr.function.name == "cos":
+                par, = expr.parameters
+                return np.cos(self.rec(par))
+            if expr.function.name == "clsin":
+                par, = expr.parameters
+                return clmath.sin(self.rec(par))
+            elif expr.function.name == "clcos":
+                par, = expr.parameters
+                return clmath.cos(self.rec(par))
```
inducer, line 91: Rather than distinguishing `cos` and `clcos`, you can just go

```python
par_val = self.rec(par)
from numbers import Number
if isinstance(par_val, Number):
   ...
else:
   ....
```

test/test_wave.py
```diff
@@ -20,4 +20,277 @@
...
+            sym_phi = sym_phi * sym_clcos(sym_coords[i])
+
+        # u = phi_t
+        sym_u = sym_diff(sym_t)(sym_phi)
+
+        # v_i = c*phi_{x_i}
+        sym_v = []
+        for i in range(dim):
+            sym_v_i = sym_c * sym_diff(sym_coords[i])(sym_phi)
+            sym_v.append(sym_v_i)
```
inducer, line 142: Check out list comprehensions.

inducer:
As for your error,
```
    CONDA_PKGS_DIR: /Users/runner/conda_pkgs_dir
ls: /usr/share/miniconda/envs/test/etc/OpenCL/vendors: No such file or directory
```
I'd suggest not building on the CI's conda (and simply installing our own from miniforge). This will make the path discrepancy go away.

matthiasdiener:
> As for your error,
> 
> ```
>     CONDA_PKGS_DIR: /Users/runner/conda_pkgs_dir
> ls: /usr/share/miniconda/envs/test/etc/OpenCL/vendors: No such file or directory
> ```
> 
> I'd suggest not building on the CI's conda (and simply installing our own from miniforge). This will make the path discrepancy go away.

Strange, this isn't the same error we saw yesterday, which was:
```
E           AttributeError: 'EagerDGDiscretization' object has no attribute '_memoize_dic_opposite_face_connection'
```

inducer:
For this error:
```
E AttributeError: 'EagerDGDiscretization' object has no attribute '_memoize_dic_opposite_face_connection'
```
There's usually another error further below, after a "While handling this... another error occurred:". (Memoize-related `KeyError` or `AttributeError` are just failed cache lookups, which means that it had to go and actually do some computation rather than pulling something out of cache.)

majosm:
Here's the real error (I think?):

```
TypeError: enqueue_nd_range_kernel(): incompatible function arguments. The following argument types are supported:
401
    1. (queue: pyopencl._cl.CommandQueue, kernel: pyopencl._cl.Kernel, global_work_size: object, local_work_size: object, global_work_offset: object = None, wait_for: object = None, g_times_l: bool = False) -> pyopencl._cl.Event
```

originating from this line in the test:
```python
nodes = discr.nodes().with_queue(queue)
```
Full text [here](https://github.com/illinois-ceesd/mirgecom/runs/789137720).

matthiasdiener:
Fixing the path error on MacOS made the CI tests run.

matthiasdiener:
Flake8 output (from #10):

```
./test/test_wave.py:45:5: E306 expected 1 blank line before a nested definition, found 0
./test/test_wave.py:83:5: E301 expected 1 blank line, found 0
./test/test_wave.py:89:5: E301 expected 1 blank line, found 0
./test/test_wave.py:113:9: F401 'meshmode.discretization.Discretization' imported but unused
./test/test_wave.py:206:9: F401 'meshmode.discretization.Discretization' imported but unused
```

matthiasdiener:
Flake8 tests pass now.

majosm:
Cleaned up the history a little. (Whoever merges: please rebase instead of squashing; some of these commits should remain separate.) Any final thoughts before merging @inducer, @matthiasdiener ?

Adding @inducer 's comment from #10 (sorry, I may have merged prematurely):

> Like this, Flake8 shows up as a "step" rather than a separate job. I'm OK with this, but I'm used to it being a separate job. It's also run across both OSs, although there's no expectation it would change its mind on anything based on what OS it runs on.

inducer:
> `TypeError: enqueue_nd_range_kernel(): incompatible function arguments.`

My fault. [loopy!396](https://gitlab.tiker.net/inducer/loopy/-/merge_requests/396) landed before a new release of pyopencl including [pyopencl!113](https://gitlab.tiker.net/inducer/pyopencl/-/merge_requests/113) propagated to conda. Should be all good now.

inducer:
> (sorry, I may have merged prematurely)

NP. We can fix the Flake8 thing if/whenever we feel like it. The important bit is that it gets run.

matthiasdiener:
> > (sorry, I may have merged prematurely)
> 
> NP. We can fix the Flake8 thing if/whenever we feel like it. The important bit is that it gets run.

I'm adding a small fix that should address your main issue with flake8.

test/test_wave.py
```diff
@@ -20,4 +20,262 @@
...
+        eval_values["c"] = c
+        eval_values["omega"] = np.sqrt(dim)*c
+        eval_values["pi"] = np.pi
+
+        def sym_eval(expr):
+            return EvaluationMapper(eval_values)(expr)
+
+        u = sym_eval(sym_u)
+        v = [sym_eval(sym_v[i]) for i in range(dim)]
+        fields = join_fields(u, v)
```
inducer, line 164: `join_fields` is deprecated. `pytools.obj_array.flat_obj_array` is the new hotness.

x-ref: https://gitlab.tiker.net/inducer/pytools/-/merge_requests/41

matthiasdiener:
> > > (sorry, I may have merged prematurely)
> > 
> > 
> > NP. We can fix the Flake8 thing if/whenever we feel like it. The important bit is that it gets run.
> 
> I'm adding a small fix that should address your main issue with flake8.

Done. Flake8 will now only be run on Linux, not MacOS.

test/test_wave.py
```diff
@@ -20,4 +20,262 @@
...
+
+        from mirgecom.wave import wave_operator
+        rhs = wave_operator(discr, c=c, w=fields)
+
+        expected_rhs = make_obj_array([sym_eval(sym_rhs_i) for sym_rhs_i in sym_rhs])
```
inducer, line 169: If you make `sym_rhs` an object array, you can call `sym_eval` on it directly.

test/test_wave.py
```diff
@@ -20,4 +20,262 @@
...
+        eval_values["t"] = 0.
+        eval_values["c"] = c
+        eval_values["omega"] = np.sqrt(dim)*c
+        eval_values["pi"] = np.pi
+
+        def sym_eval(expr):
+            return EvaluationMapper(eval_values)(expr)
+
+        u = sym_eval(sym_u)
+        v = [sym_eval(sym_v[i]) for i in range(dim)]
```
inducer, line 163: Here same as below: Make use of the fact that pymbolic recurses into object arrays on its own.

test/test_wave.py
```diff
@@ -20,4 +20,262 @@
...
+
+        c = 2.
+
+        eval_values = dict()
+        for i in range(dim):
+            eval_values[coord_names[i]] = nodes[i]
+        eval_values["t"] = 0.
+        eval_values["c"] = c
+        eval_values["omega"] = np.sqrt(dim)*c
+        eval_values["pi"] = np.pi
```
inducer, line 157: Use dict literal and dict comprehension.

test/test_wave.py
```diff
@@ -20,4 +20,262 @@
...
+
+        # v_i = c*phi_{x_i}
+        sym_v = [sym_c * sym_diff(sym_coords[i])(sym_phi) for i in range(dim)]
+
+        # rhs(u part) = c*div(v)
+        sym_rhs_u = sym_c * sym_div(sym_v)
+        # rhs(v part) = c*grad(u)
+        sym_grad_u = sym_grad(dim, sym_u)
+        sym_rhs_v = [sym_c * sym_grad_u[i] for i in range(dim)]
+        sym_rhs = [sym_rhs_u] + sym_rhs_v
```
inducer, line 147: I think `+` is an unfortunate operator here, since it's got nothing to do with arithmetic plus.

test/test_wave.py
```diff
@@ -20,4 +20,262 @@
...
+        sym_cos = pmbl.var("cos")
+        sym_phi = sym_cos(sym_omega*sym_t - sym_pi/4)
+        for i in range(dim):
+            sym_phi = sym_phi * sym_cos(sym_coords[i])
+
+        # u = phi_t
+        sym_u = sym_diff(sym_t)(sym_phi)
+
+        # v_i = c*phi_{x_i}
+        sym_v = [sym_c * sym_diff(sym_coords[i])(sym_phi) for i in range(dim)]
```
inducer, line 140: As below: `sym_diff` will happily differentiate object arrays.

test/test_wave.py
```diff
@@ -20,4 +20,262 @@
...
+
+        # rhs(u part) = c*div(v)
+        sym_rhs_u = sym_c * sym_div(sym_v)
+        # rhs(v part) = c*grad(u)
+        sym_grad_u = sym_grad(dim, sym_u)
+        sym_rhs_v = [sym_c * sym_grad_u[i] for i in range(dim)]
+        sym_rhs = [sym_rhs_u] + sym_rhs_v
```
inducer, line 239: Pretty redundant with the above: factor into common function?

test/test_wave.py
majosm: Having some trouble with this one. I think I know how to do dict comprehension and literals on their own, but not sure how to combine them. Is there a way to do this all in one line?

test/test_wave.py
majosm: Not sure how to do this, since it's the argument to DifferentiationMapper's constructor that is an array. If I make sym_coords an object array and try to do `sym_diff(sym_coords)(sym_phi)`, I get:

```python
    def __eq__(self, other):
        """Provides equality testing with quick positive and negative paths
        based on :func:`id` and :meth:`__hash__`.
    
        Subclasses should generally not override this method, but instead
        provide an implementation of :meth:`is_equal`.
        """
        if self is other:
            return True
>       elif hash(self) != hash(other):
E       TypeError: unhashable type: 'numpy.ndarray'
```

Is there a way I can modify the definition of `sym_diff` to make this work?

Also, can I multiply an object array by `sym_c`?

inducer: > Not sure how to do this, since it's the argument to DifferentiationMapper's constructor that is an array.

Ah, nvm. Spaced on that.

> Also, can I multiply an object array by `sym_c`?

Yes, I think so.

test/test_wave.py
inducer: https://www.python.org/dev/peps/pep-0584/

so... no, at least not yet. But the PEP contains some alternatives that are usable until then. Pick the one that you find the least bad.

test/test_wave.py
majosm: I couldn't get it to work with just `sym_c`, but `make_obj_array([sym_c])` did the trick.

majosm:
So close...

```
229
  File "/Users/runner/work/mirgecom/mirgecom/test/test_wave.py", line 166, in test_standing_wave
230
    rhs = wave_operator(discr, c=c, w=fields)
231
  File "/Users/runner/work/mirgecom/mirgecom/mirgecom/wave.py", line 76, in wave_operator
232
    dir_u = discr.interp("vol", BTAG_ALL, u)
233
  File "/usr/local/miniconda/envs/test/lib/python3.8/site-packages/grudge/eager.py", line 51, in interp
234
    +                DeprecationWarning, stacklevel=2)
235
TypeError: bad operand type for unary +: 'type'
```

Any ideas what this means?

majosm:
[The line in question](https://github.com/inducer/grudge/blob/master/grudge/eager.py#L51)

@inducer: stray `+`?

inducer:
> stray +?

Yes, definitely. I'm super confused how this made it past CI. I'll investigate and push a fix shortly.

inducer:
> I'm super confused how this made it past CI.

I get it. It's still valid syntax, but because it's in a deprecation path, it didn't run, so nobody noticed that it didn't work.

https://github.com/inducer/grudge/commit/d92e26e22ea8954e5fd1d3afafb02c4c13d4b6c5

inducer:
Sorry about that!

PR #10, opened 2020-06-24
Add Flake8 integration in CI

matthiasdiener:
Addresses https://github.com/illinois-ceesd/planning/issues/51

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -20,16 +20,17 @@ jobs:
           with:
             auto-update-conda: true
             channels: conda-forge, nodefaults
-        - name: Flake8
-          run: |
-            pip install flake8 pep8-naming
-            flake8
         - name: Install prerequisites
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -26,6 +26,11 @@ jobs:
             conda install pocl islpy pyopencl pytest pip clinfo
             pip install git+https://gitlab.tiker.net/inducer/{loopy,meshmode,dagrt,leap,modepy,meshpy,grudge}
             pip install -e .
+            pip install flake8 pep8-naming
+        - name: Flake8
+          shell: bash -l {0}
+          run: |
+            python -m flake8
```
inducer, line 33: Like this, Flake8 shows up as a "step" rather than a separate job. I'm OK with this, but I'm used to it being a separate job. It's also run across both OSs, although there's no expectation it would change its mind on anything based on what OS it runs on.

PR #16, opened 2020-06-29
Clean up wave tests

### LINKED ISSUE #17: Factor analytical solution tools into a submodule
@majosm, could you move [these guys](https://github.com/illinois-ceesd/mirgecom/blob/feebea60858b182504a7d47614b077883fc4872a/test/test_wave.py#L43-L98) into an importable submodule so that @MTCam can use them to check Euler solutions?

majosm:
Combines existing wave tests into one test parametrized on the problem being solved, and separates symbolic bits into a module for other tests to use.

Closes #17

### DIFF
--- a/mirgecom/symbolic.py
+++ b/mirgecom/symbolic.py
@@ -35,7 +35,8 @@ __doc__ = """
 .. autofunction:: diff
 .. autofunction:: div
 .. autofunction:: grad
-.. autofunction:: evaluate
+
+.. autoclass:: EvaluationMapper
 """
 
 
@@ -71,6 +72,8 @@ def grad(dim, func):
 
 
 class EvaluationMapper(ev.EvaluationMapper):
+    """Evaluates symbolic expressions given a mapping from variables to values"""
+
     def map_call(self, expr):
         assert isinstance(expr.function, prim.Variable)
         if expr.function.name == "sin":
@@ -95,10 +98,3 @@ class EvaluationMapper(ev.EvaluationMapper):
             return np.cos(val)
         else:
             return clmath.cos(val)
-
-
-def evaluate(expr, **kwargs):
-    """Return the value of a symbolic expression by substituting in keyword
-    argument values.
-    """
-    return EvaluationMapper(kwargs)(expr)
--- a/mirgecom/symbolic.py
+++ b/mirgecom/symbolic.py
@@ -1,0 +1,104 @@
+__copyright__ = """Copyright (C) 2020 CEESD Developers"""
+
+__license__ = """
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+"""
+
+import numpy as np
+import numpy.linalg as la # noqa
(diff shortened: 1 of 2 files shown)


test/test_wave.py
```diff
@@ -98,165 +98,104 @@ def _cos(self, val):
...
-        eval_values.update({coord_names[i]: nodes[i] for i in range(dim)})
+                    c * sym_div(sym_v) + sym_f,
+                    make_obj_array([c]) * sym_grad(dim, sym_u))
 
-        def sym_eval(expr):
+        def sym_eval(expr, t):
+            coord_names = ["x", "y", "z"]
+            eval_values = {coord_names[i]: nodes[i] for i in range(dim)}
+            eval_values["t"] = t
             return EvaluationMapper(eval_values)(expr)
```
inducer, line 188: - Make this take `**kwargs` instead of just `t`.
- Make it reusable along with #17.

test/test_wave.py
```diff
@@ -98,165 +98,104 @@ def _cos(self, val):
...
-        sym_c = pmbl.var("c")
-        sym_omega = pmbl.var("omega")
-        sym_pi = pmbl.var("pi")
-        sym_cos = pmbl.var("cos")
-        sym_phi = sym_cos(sym_omega*sym_t - sym_pi/4)
-        for i in range(dim):
-            sym_phi = sym_phi * (sym_coords[i]-1)**3 * (sym_coords[i]+1)**3
+
+        # f = phi_tt - c^2 * div(grad(phi))
+        sym_f = sym_diff(sym_t)(sym_diff(sym_t)(sym_phi)) - c**2\
```
inducer, line 169:
```suggestion
        sym_wave_op_residual = sym_diff(sym_t)(sym_diff(sym_t)(sym_phi)) - c**2\
```

test/test_wave.py
```diff
@@ -98,165 +98,104 @@ def _cos(self, val):
...
+        v = sym_eval(sym_v, 0.)
+        f = sym_eval(sym_f, 0.)
 
         from mirgecom.wave import wave_operator
-        rhs = wave_operator(discr, c=c, w=fields)
+        rhs = wave_operator(discr, c=c, w=flat_obj_array(u, v))
+        rhs[0] = rhs[0] + f
 
-        expected_rhs = sym_eval(sym_rhs)
+        expected_rhs = sym_eval(sym_rhs, 0.)
```
inducer, line 198: Try another time point. `t=0` is potentially special.

test/test_wave.py
inducer, line 169: Is there anyplace that checks that this is small?

test/test_wave.py
majosm: Trying to figure out how I could pass in the coordinates via `**kwargs` without requiring different call syntax for each problem dimension... (i.e., `sym_eval(expr, x=nodes[0], y=nodes[1])` for 2D vs. `sym_eval(expr, x=nodes[0], y=nodes[1], z=nodes[2])` for 3D.) Does pymbolic understand array variables? Could I define something along the lines of `pmbl.var('x[0]')`, `pmbl.var('x[1]')`, etc. rather than x,y,z and then do `sym_eval(expr, x=nodes)`?

test/test_wave.py
majosm: Yeah, this will be going away shortly. This PR is mainly a stepping stone towards adding time integration to the tests (wanted to combine first so I have less code to change).

test/test_wave.py
inducer: https://documen.tician.de/pymbolic/primitives.html#pymbolic.primitives.make_sym_vector

test/test_wave.py
majosm: So, I'm sort of looking at this from a manufactured solution point of view, where instead of solving `phi_tt = c^2 * div(grad(phi))` I'm solving `phi_tt = c^2 * div(grad(phi)) + f` where `f` comes from the solution being prescribed, and is not necessarily expected to be small. The machinery is set up here so that you can basically plug in any function you want (subject to boundary conditions), it will derive the `f` that makes the prescribed function a solution, compute numerical and symbolic RHSs for this modified equation, and then compare them. So in that sense it shouldn't really matter if `f` is small.

majosm:
Closes #17

test/test_wave.py
inducer: OIC, that makes sense, and it's fine by me. Could you add some comments that state this explicitly? I guess you have one data point in that I didn't manage to figure this out despite some staring. (But then I also stubbornly assumed I know what was going on. :)

mirgecom/symbolic.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def _cos(self, val):
+        from numbers import Number
+        if isinstance(val, Number):
+            return np.cos(val)
+        else:
+            return clmath.cos(val)
+
+
+def evaluate(expr, **kwargs):
```
inducer, line 100: Might as well expose the mapper rather than a wrapper.

mirgecom/symbolic.py
```diff
@@ -0,0 +1,104 @@
+__copyright__ = """Copyright (C) 2020 CEESD Developers"""
```
inducer, line 1:
```suggestion
__copyright__ = """Copyright (C) 2020 University of Illinois Board of Trustees"""
```

https://otm.illinois.edu/disclose-protect/open-source-licensing

inducer:
Thanks for working on this!

Issue #19, opened 2020-07-01
Make flake8, tests, and examples separate jobs


inducer:
So they can run in parallel.

matthiasdiener:
Flake8 already runs in parallel.

PR #21, opened 2020-07-08
Separate Flake8 test

matthiasdiener:
Partially addresses #19.

### DIFF
--- a/.github/workflows/flake8.yaml
+++ b/.github/workflows/flake8.yaml
@@ -11,5 +11,6 @@ jobs:
     steps:
         - name: Flake8 test
           run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/flake8.yaml
```diff
@@ -0,0 +1,16 @@
+name: Flake8
+
+# Trigger the workflow on push
+on:
+  push:
```
inducer, line 5: If you don't pin the flake8 version, you should have a scheduled (weekly?) run.

matthiasdiener: Fixed.

PR #7, opened 2020-06-18
Euler 

inducer:
cc @MTCam

PR #23, opened 2020-07-09
add profiling array context

matthiasdiener:
Continuation of https://gitlab.tiker.net/inducer/meshmode/-/merge_requests/89


examples/simple-dg.py
```diff
@@ -0,0 +1,527 @@
+from __future__ import division, print_function
```
inducer, line 1: Why is this duplicated here? Are there any changes wrt the original?

examples/eulerflow_profiling.py
```diff
@@ -0,0 +1,229 @@
+__copyright__ = """
```
inducer, line 1: Do you mean to have an Euler file here?

examples/eulerflow_profiling.py
```diff
@@ -0,0 +1,229 @@
...
+Copyright (C) 2020 University of Illinois Board of Trustees
+"""
+
+__license__ = """
+Permission is hereby granted, free of charge, to any person obtaining a copy
```
inducer, line 6: My understanding is that array context support for Euler will happen in #7, and `mtc/array-context` has sort of fallen by the wayside.

@MTCam, correct?

mirgecom/profiling.py
```diff
@@ -0,0 +1,201 @@
...
+        from statistics import mean
+
+        for key, value in self.profiling_data.items():
+            num_values = len(value)
+
+            times = [v.time for v in value]
+            flops = [v.flops for v in value]
+            mem_access = [v.mem_access for v in value]
+
+            print(format_str.format(key, num_values, min(times), int(mean(times)), max(times), min(flops), int(mean(flops)), max(flops), min(mem_access), int(mean(mem_access)), max(mem_access), round(mean(mem_access)/mean(times),3) ))
```
inducer, line 149: Flake8?

mirgecom/profiling.py
```diff
@@ -0,0 +1,201 @@
...
+            # get statistics
+            program = lp.add_and_infer_dtypes(program, types)
+            op_map = lp.get_op_map(program, subgroup_size='guess')
+            mem_map = lp.get_mem_access_map(program, subgroup_size='guess')
+
+            f32op_count = op_map.filter_by(dtype=[np.float32]).eval_and_sum(param_dict)
+            f64op_count = op_map.filter_by(dtype=[np.float64]).eval_and_sum(param_dict)
+
+            mem32_count = mem_map.filter_by(dtype=[np.float32]).eval_and_sum(param_dict)
+            mem64_count = mem_map.filter_by(dtype=[np.float64]).eval_and_sum(param_dict)
```
inducer, line 111: Doing this repeatedly will be horrendously expensive without caching.

examples/simple-dg.py
matthiasdiener: This was just due to merging with the other array context branch I believe. Removed.

examples/eulerflow_profiling.py
matthiasdiener: This was just due to merging with the other array context branch I believe. Removed.

examples/eulerflow_profiling.py
matthiasdiener: This was just due to merging with the other array context branch I believe. Removed.

mirgecom/profiling.py
matthiasdiener: Fixed.

mirgecom/profiling.py
matthiasdiener: Should be fixed in the new version.

matthiasdiener:
This currently crashes in get_memory_map (see https://github.com/illinois-ceesd/mirgecom/pull/23/checks?check_run_id=942573825) but I don't know why.

inducer:
https://github.com/inducer/loopy/pull/132 should address the crashes. This boiled down to an unhandled case in stride gathering.

matthiasdiener:
This gets much further now, but fails again in get_mem_access_map for the 'flatten' kernel. https://github.com/illinois-ceesd/mirgecom/pull/23/checks?check_run_id=949858833

matthiasdiener:
==================== ====== ======== ======== ======== ======== ======== ======== ======== ======== ======== ========
Function             Calls    T_min    T_avg    T_max    F_min    F_avg    F_max    M_min    M_avg    M_max   BW_avg
==================== ====== ======== ======== ======== ======== ======== ======== ======== ======== ======== ========
nodes                   10    11948   177663   521261     3840    60480    86400     2640    51840   100800    0.292
actx_special_exp         1   271568   271568   271568     7200     7200     7200     9000     9000     9000    0.033
resample_by_picking  75372     4678   143234  4385972        0        0        0      240     2290     3600    0.016
diff                  6854    62362   210808  2577113   288000   288000   288000   243000   243000   243000    1.153
grudge_assign_0          1    61451    61451    61451    79200    79200    79200    58500    58500    58500    0.952
actx_special_sqrt        4    29963   277521   436603      960    12600    21600      480     7650    10800    0.028
grudge_assign_0       6852    36632   156127  1708563    43200    43200    43200    45000    45000    45000    0.288
resample_by_mat      13875    31748    98802  1890861    53760    56646   288000    35280    38572   302400     0.39
diff                     6    14184   183721   457775     7680   115200   172800     5040    75600   113400    0.411
grudge_assign_0       6852    47686   229557  2018377    21600    21600    21600    16200    16200    16200    0.071
face_mass             6852    54913   205928  2251934   194400   194400   194400   144900   144900   144900    0.704
grudge_assign_0       6852    27210   129499  1911930    14400    14400    14400    13500    13500    13500    0.104
diff                  6852    52469   180302  1797766   144000   144000   144000   121500   121500   121500    0.674
grudge_assign_0         57    36206   106284  1213514     7200     7200     7200     9000     9000     9000    0.085
flatten                173    47250   271904  1537884        0        0        0    28800    28800    28800    0.106
==================== ====== ======== ======== ======== ======== ======== ======== ======== ======== ======== ========

matthiasdiener:
I think the overhead is manageable now (times for wave-eager):

- Baseline (no profiling whatsoever): 61.01s
- Creating the queue with cl.command_queue_properties.PROFILING_ENABLE: 64.88s
- Storing profiling events with this PR: 68.25s
- ... + synchronizing and printing profiling stats every 10 time steps: 72.85s

mirgecom/profiling.py
```diff
@@ -0,0 +1,208 @@
...
+
+                for d in data:
+                    if d.name not in param_dict:
+                        param_dict[d.name] = None
+
+                # Run the wrapper code, save argument values in param_dict
+                exec(gen.get(), param_dict)
+
+                # Get flops/memory statistics
+                kernel.options.ignore_boostable_into = True
```
inducer, line 180: Stateful modification is not allowed, for any data structure in loopy, and for most data structures in code I write.
```
kernel = kernel.copy(
   options=options.copy(ignore_boostable_into=True))
```
is the correct idiom.

But given https://github.com/inducer/loopy/pull/128, this should be a no-op anyway... so I wonder what this is doing here.

mirgecom/profiling.py
```diff
@@ -0,0 +1,208 @@
...
+                kernel = lp.add_and_infer_dtypes(kernel, types)
+                op_map = lp.get_op_map(kernel, subgroup_size='guess')
+                mem_map = lp.get_mem_access_map(kernel, subgroup_size='guess')
+
+                f32op = op_map.filter_by(dtype=[np.float32]).eval_and_sum(param_dict)
+                f64op = op_map.filter_by(dtype=[np.float64]).eval_and_sum(param_dict)
```
inducer, line 186: Why specialize to those two?

mirgecom/profiling.py
```diff
@@ -0,0 +1,208 @@
...
+
+                mem32 = mem_map.filter_by(dtype=[np.float32]).eval_and_sum(
+                    param_dict)
+                mem64 = mem_map.filter_by(dtype=[np.float64]).eval_and_sum(
+                    param_dict)
```
inducer, line 191: This is liable to miss memory accesses. There is a function to count total bytes. Use that instead.

mirgecom/profiling.py
```diff
@@ -0,0 +1,208 @@
...
+                if program not in self.kernel_stats:
+                    self.kernel_stats[program] = {}
+
+                self.kernel_stats[program][args_tuple] = (flops, bytes_accessed)
+
+        evt, result = program(self.queue, **kwargs, allocator=self.allocator)
+
+        if self.profiling_enabled:
+            flops = self.kernel_stats[program][args_tuple][0]
+            bytes_accessed = self.kernel_stats[program][args_tuple][1]
```
inducer, line 205: Why is this here? The variables are still set from above? Also, use tuple unpacking, not indexing.

mirgecom/profiling.py
matthiasdiener: This one? https://documen.tician.de/loopy/ref_other.html#loopy.ToCountMap.to_bytes

mirgecom/profiling.py
matthiasdiener: Changed to `np.float`.

mirgecom/profiling.py
matthiasdiener: Note that `kernel` is already a copy of `program`. As you said, this line seems to be not necessary anymore (but was necessary earlier when I wrote the code). Removed.

mirgecom/profiling.py
inducer: Yep.

mirgecom/profiling.py
inducer: Float is an alias for `float64`. Not better.

mirgecom/profiling.py
inducer: It's still invalid, despite being a copy. `LoopKernel` computes (and stores) a hash key. If you modify it in-place, you'll screw up the haashing logic royally.

mirgecom/profiling.py
matthiasdiener: Ok, what do you suggest? Add a line for `np.float16`?

mirgecom/profiling.py
matthiasdiener: Fixed.

mirgecom/profiling.py
inducer: On second thought, I think this is OK. Code review puts me in a pretty critical mood.

matthiasdiener: Ok, I simplified the code here a bit nevertheless.

mirgecom/profiling.py
```diff
@@ -0,0 +1,211 @@
...
+                if hasattr(value, 'shape'):
+                    args_tuple = args_tuple + (key, value.shape)
+                else:
+                    args_tuple = args_tuple + (key, value)
+
+            # Check if we need to get the invoker code to generate integer arguments
+            # N.B.: The invoker code might be different for the same program with
+            # different kwargs
+            if (program not in self.kernel_stats
+              or args_tuple not in self.kernel_stats[program]):
```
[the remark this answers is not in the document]
matthiasdiener: Fixed.

requirements.txt
```diff
@@ -6,7 +6,7 @@ pyvisfile
 git+https://github.com/inducer/pymbolic.git#egg=pymbolic
 git+https://github.com/inducer/islpy.git#egg=islpy
 git+https://github.com/inducer/pyopencl.git#egg=pyopencl
-git+https://github.com/inducer/loopy.git#egg=loo.py
+git+https://github.com/matthiasdiener/loopy.git@return_gen
```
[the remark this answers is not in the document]
matthiasdiener: Fixed.

mirgecom/profiling.py
matthiasdiener: Variables were set above only when the `if` was true. Clarified the code a bit by putting this in an `else`.

examples/wave-eager.py
```diff
@@ -54,8 +55,9 @@ def bump(actx, discr, t=0):
 
 def main():
     cl_ctx = cl.create_some_context()
-    queue = cl.CommandQueue(cl_ctx)
-    actx = PyOpenCLArrayContext(queue,
+    queue = cl.CommandQueue(cl_ctx,
+        properties=cl.command_queue_properties.PROFILING_ENABLE)
+    actx = PyOpenCLProfilingArrayContext(queue,
         allocator=cl_tools.MemoryPool(cl_tools.ImmediateAllocator(queue)))
```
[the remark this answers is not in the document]
matthiasdiener: Done.

mirgecom/profiling.py
```diff
@@ -0,0 +1,208 @@
...
+        self.flops = flops
+        self.bytes_accessed = bytes_accessed
+
+
+class TimingEvent:
+    def __init__(self, event, program, flops, bytes_accessed) -> None:
+        self.event = event
+        self.program = program
+        self.flops = flops
+        self.bytes_accessed = bytes_accessed
```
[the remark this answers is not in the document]
matthiasdiener: Done.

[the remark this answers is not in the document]
matthiasdiener: Done.

mirgecom/profiling.py
```diff
@@ -0,0 +1,208 @@
...
+from pytools.py_codegen import PythonCodeGenerator
+import loopy as lp
+import numpy as np
+
+
+class ProfileData:
+    def __init__(self, time, flops, bytes_accessed) -> None:
+        self.time = time
+        self.flops = flops
+        self.bytes_accessed = bytes_accessed
```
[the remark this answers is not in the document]
matthiasdiener: Done.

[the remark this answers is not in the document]
matthiasdiener: Done.

mirgecom/profiling.py
```diff
@@ -0,0 +1,208 @@
...
+
+        if not queue.properties & cl.command_queue_properties.PROFILING_ENABLE:
+            from warnings import warn
+            warn("Profiling was not enabled in the command queue."
+                 "Timing data will not be collected.")
+            self.profiling_enabled = False
+        else:
+            self.profiling_enabled = True
+
+        self.events = []
```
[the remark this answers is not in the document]
matthiasdiener: renamed to `profile_events`.

mirgecom/profiling.py
```diff
@@ -0,0 +1,208 @@
...
+"""
+
+from meshmode.array_context import PyOpenCLArrayContext
+import pyopencl as cl
+from pytools.py_codegen import PythonCodeGenerator
+import loopy as lp
+import numpy as np
+
+
+class ProfileData:
```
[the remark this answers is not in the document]
matthiasdiener: Renamed to ProfileResult. Not sure if it is better.

mirgecom/profiling.py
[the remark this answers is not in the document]
matthiasdiener: Fixed by removing the extra class.

mirgecom/profiling.py
```diff
@@ -0,0 +1,208 @@
...
+    def __init__(self, time, flops, bytes_accessed) -> None:
+        self.time = time
+        self.flops = flops
+        self.bytes_accessed = bytes_accessed
+
+
+class TimingEvent:
```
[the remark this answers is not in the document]
matthiasdiener: Renamed to ProfileEvent.

mirgecom/profiling.py
```diff
@@ -0,0 +1,208 @@
...
+            evt = t.event
+
+            time = evt.profile.end - evt.profile.start
+
+            if program in self.profiling_data:
+                self.profiling_data[program].append(
+                    ProfileData(time, flops, bytes_accessed))
+            else:
+                self.profiling_data[program] = [
+                    ProfileData(time, flops, bytes_accessed)]
```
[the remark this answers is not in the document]
matthiasdiener: Fixed.

mirgecom/profiling.py
```diff
@@ -0,0 +1,208 @@
...
+            flops = [v.flops for v in value]
+            bytes_accessed = [v.bytes_accessed for v in value]
+
+            print(format_str.format(key.name, num_values, min(times),
+                int(mean(times)), max(times), min(flops), int(mean(flops)),
+                max(flops), min(bytes_accessed), int(mean(bytes_accessed)),
+                max(bytes_accessed), round(mean(bytes_accessed)/mean(times), 3)))
+
+        print(format_str.format('='*20, '='*6, '='*8, '='*8, '='*8, '='*8, '='*8,
+            '='*8, '='*8, '='*8, '='*8, '='*8))
```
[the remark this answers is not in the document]
matthiasdiener: Changed to pytools.Table.

mirgecom/profiling.py
```diff
@@ -0,0 +1,208 @@
...
+        if not self.profiling_enabled:
+            return
+
+        if self.events:
+            cl.wait_for_events([t.event for t in self.events])
+
+        for t in self.events:
+            program = t.program
+            flops = t.flops
+            bytes_accessed = t.bytes_accessed
```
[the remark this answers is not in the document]
matthiasdiener: Fixed.

mirgecom/profiling.py
```diff
@@ -0,0 +1,208 @@
...
+
+class TimingEvent:
+    def __init__(self, event, program, flops, bytes_accessed) -> None:
+        self.event = event
+        self.program = program
+        self.flops = flops
+        self.bytes_accessed = bytes_accessed
+
+
+class PyOpenCLProfilingArrayContext(PyOpenCLArrayContext):
```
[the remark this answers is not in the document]
matthiasdiener: Added.

matthiasdiener:
Thank you for the detailed review @inducer! I think I addressed all of your comments.

matthiasdiener:
Updated requirements.txt to reflect recently merged loopy patch.

matthiasdiener:
Ready for review.

mirgecom/profiling.py
```diff
@@ -0,0 +1,175 @@
...
+            for key, value in kernel.arg_dict.items():
+                if key not in param_dict:
+                    param_dict[key] = None
+
+            for d in data:
+                if d.name not in param_dict:
+                    param_dict[d.name] = None
+
+            # Run the wrapper code, save argument values in param_dict
+            exec(gen.get(), param_dict)
```
inducer, line 157: If this code were stuck into a function, it
* would execute faster (global var retrievals use `dict` lookups, function-local variables use a more efficient, numbering-based mechanism)
* could be cached.

This is particularly important in a setting where size arguments to kernels can change rapidly.

mirgecom/profiling.py
```diff
@@ -0,0 +1,175 @@
...
+
+            # Get flops/memory statistics
+            kernel = lp.add_and_infer_dtypes(kernel, types)
+            op_map = lp.get_op_map(kernel, subgroup_size='guess')
+            bytes_accessed = lp.get_mem_access_map(kernel, subgroup_size='guess') \
+              .to_bytes().eval_and_sum(param_dict)
```
inducer, line 163: We should also gather the size of the accessed footprint.

mirgecom/profiling.py
```diff
@@ -0,0 +1,175 @@
...
+                max(flops), min(bytes_accessed), int(mean(bytes_accessed)),
+                max(bytes_accessed), round(mean(bytes_accessed)/mean(times), 3)])
+
+        if hasattr(pytools.Table, 'github_markdown'):
+            print(tbl.github_markdown())
+        else:
+            print(tbl)
+
+    def call_loopy(self, program, **kwargs) -> dict:
+        program = self.transform_loopy_program(program)
```
inducer, line 110: We had discussed gathering stats on the untransformed program. (Perhaps in addition to the transformed program.) What do you think?

mirgecom/profiling.py
```diff
@@ -0,0 +1,175 @@
...
+
+            tbl.add_row([key.name, num_values, min(times),
+                int(mean(times)), max(times), min(flops), int(mean(flops)),
+                max(flops), min(bytes_accessed), int(mean(bytes_accessed)),
+                max(bytes_accessed), round(mean(bytes_accessed)/mean(times), 3)])
+
+        if hasattr(pytools.Table, 'github_markdown'):
+            print(tbl.github_markdown())
+        else:
+            print(tbl)
```
inducer, line 107: Add a version constraint on `pytools` in `setup.py` to make sure you can rely on the new functionality being there. You'll probably need to submit a version bump PR to `pytools`. I think that's preferable to this silent behavior change based on versions.

mirgecom/profiling.py
```diff
@@ -0,0 +1,175 @@
...
+        self.kernel_stats = {}
+
+    def finish_profile_events(self) -> None:
+        # First, wait for event completion
+        if self.profile_events:
+            cl.wait_for_events([t.cl_event for t in self.profile_events])
+
+        for t in self.profile_events:
+            program = t.program
+            (flops, bytes_accessed) = self.kernel_stats[program][t.args_tuple]
```
inducer, line 72: Not a fan of this being an unlabeled tuple. Not extensible, and the code dealing with that data isn't super readable.

mirgecom/profiling.py
```diff
@@ -0,0 +1,208 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+"""
+
+from meshmode.array_context import PyOpenCLArrayContext
+import pyopencl as cl
+from pytools.py_codegen import PythonCodeGenerator
+import loopy as lp
+import numpy as np
+
```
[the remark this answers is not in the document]
matthiasdiener: Added to `doc/running.rst`.

mirgecom/profiling.py
matthiasdiener: In other words, you would prefer a dataclass here?

mirgecom/profiling.py
matthiasdiener: This was just here to test the pytools stuff. Changed to just do `print(tbl)`.

mirgecom/profiling.py
inducer: Yeah.

mirgecom/profiling.py
```diff
@@ -0,0 +1,175 @@
...
+                    types[key] = value.dtype
+                param_dict[key] = value
+
+            for key, value in kernel.arg_dict.items():
+                if key not in param_dict:
+                    param_dict[key] = None
+
+            for d in data:
+                if d.name not in param_dict:
+                    param_dict[d.name] = None
```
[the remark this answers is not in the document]
matthiasdiener: Done.

mirgecom/profiling.py
matthiasdiener: Running

```python
footprint_bytes = lp.gather_access_footprint_bytes(kernel, True)
```

at this point gives the following error:

```
/Users/mdiener/Work/emirge/loo-py/loopy/statistics.py:1217: LoopyWarning: in kernel resample_by_picking: Barvinok wrappers are not installed. Counting routines have overestimated the number of integer points in your loop domain. (add 'count_overestimate' to silenced_warnings kernel attribute to disable)
  warn_with_kernel(kernel, "count_overestimate",
  File "/Users/mdiener/Work/emirge/pymbolic/pymbolic/mapper/__init__.py", line 126, in __call__
    method = getattr(self, expr.mapper_method)
AttributeError: 'tuple' object has no attribute 'mapper_method'

During handling of the above exception, another exception occurred:

  File "/Users/mdiener/Work/emirge/loo-py/loopy/symbolic.py", line 1667, in get_access_range
    idx_aff = guarded_aff_from_expr(access_map.space, subscript[idim])
  File "/Users/mdiener/Work/emirge/loo-py/loopy/symbolic.py", line 1429, in guarded_aff_from_expr
    return with_aff_conversion_guard(aff_from_expr, space, expr, vars_to_zero)
  File "/Users/mdiener/Work/emirge/loo-py/loopy/symbolic.py", line 1420, in with_aff_conversion_guard
    raise ExpressionToAffineConversionError(
loopy.diagnostic.ExpressionToAffineConversionError: could not convert expression 'to_element_indices[iel]' to affine representation: UnknownVariableError: to_element_indices

During handling of the above exception, another exception occurred:

  File "examples/wave-eager.py", line 124, in <module>
    main(True)
  File "examples/wave-eager.py", line 107, in main
  File "/Users/mdiener/Work/emirge/mirgecom/mirgecom/integrators.py", line 40, in rk4_step
    k1 = rhs(t, state)
  File "examples/wave-eager.py", line 101, in rhs
    return wave_operator(discr, c=1, w=w)
  File "/Users/mdiener/Work/emirge/mirgecom/mirgecom/wave.py", line 79, in wave_operator
    dir_u = discr.project("vol", BTAG_ALL, u)
  File "/Users/mdiener/Work/emirge/grudge/grudge/eager.py", line 73, in project
    return self.connection_from_dds(src, tgt)(vec)
  File "/Users/mdiener/Work/emirge/meshmode/meshmode/discretization/connection/direct.py", line 403, in __call__
    actx.call_loopy(pick_knl(),
  File "/Users/mdiener/Work/emirge/mirgecom/mirgecom/profiling.py", line 160, in call_loopy
    footprint_bytes = lp.gather_access_footprint_bytes(kernel, True)
  File "/Users/mdiener/Work/emirge/loo-py/loopy/statistics.py", line 1851, in gather_access_footprint_bytes
    fp = gather_access_footprints(kernel,
  File "/Users/mdiener/Work/emirge/loo-py/loopy/statistics.py", line 1815, in gather_access_footprints
    write_footprints.append(afg(insn.assignees))
  File "/Users/mdiener/Work/emirge/pymbolic/pymbolic/mapper/__init__.py", line 132, in __call__
    return self.map_foreign(expr, *args, **kwargs)
  File "/Users/mdiener/Work/emirge/pymbolic/pymbolic/mapper/__init__.py", line 182, in map_foreign
    return self.map_tuple(expr, *args, **kwargs)
  File "/Users/mdiener/Work/emirge/pymbolic/pymbolic/mapper/__init__.py", line 297, in map_list
    return self.combine(self.rec(child, *args, **kwargs) for child in expr)
  File "/Users/mdiener/Work/emirge/loo-py/loopy/statistics.py", line 1090, in combine
    return reduce(merge_dicts, values)
  File "/Users/mdiener/Work/emirge/pymbolic/pymbolic/mapper/__init__.py", line 297, in <genexpr>
    return self.combine(self.rec(child, *args, **kwargs) for child in expr)
  File "/Users/mdiener/Work/emirge/pymbolic/pymbolic/mapper/__init__.py", line 134, in __call__
    return method(expr, *args, **kwargs)
  File "/Users/mdiener/Work/emirge/loo-py/loopy/statistics.py", line 1107, in map_subscript
    access_range = get_access_range(self.domain, subscript,
  File "/Users/mdiener/Work/emirge/loo-py/loopy/symbolic.py", line 1679, in get_access_range
    raise UnableToDetermineAccessRange(
loopy.symbolic.UnableToDetermineAccessRange: unable to determine access range of subscript: [to_element_indices[iel], idof_inner + idof_outer*16] (encountered ExpressionToAffineConversionError: could not convert expression 'to_element_indices[iel]' to affine representation: UnknownVariableError: to_element_indices)
```

inducer: That's expected (as in, catch that exception and record no value). Projection operators by necessity include indirect (i.e. data-dependent) addressing. In that situation (and without further information), it's simply not possible to statically determine the access footprint.

matthiasdiener: OIC. I thought that `ignore_uncountable` would already take care of that.

mirgecom/profiling.py
matthiasdiener: Fixed.

mirgecom/profiling.py
matthiasdiener: Added.

mirgecom/profiling.py
matthiasdiener: I don't remember exactly what our conclusion was but I'm not sure gathering stats on the untransformed program makes too much sense. The performance of the application depends on the performance of the transformed kernels, so I'm not sure if the untransformed kernels add a lot if info.

inducer: >  The performance of the application depends on the performance of the transformed kernels, 

That's true, but this is about operation counting. And pre-transform kernels:
* often have rectangular loops that are easier to count with primitive (non-barvinok) tools
* have operation counts that are at least not unrelated to the number of net operations performed by the transformed kernel. Gross disagreements between transformed and untransformed could help raise red flags.

mirgecom/profiling.py
matthiasdiener: I'm having trouble putting this into a function (related to arg passing). let's discuss in the meeting today.

mirgecom/profiling.py
matthiasdiener: let's discuss in the meeting today.

mirgecom/profiling.py
```diff
@@ -0,0 +1,175 @@
...
+
+        args_tuple = tuple(
+            (key, value.shape) if hasattr(value, 'shape') else (key, value)
+            for key, value in kwargs.items())
+
+        # Check if we need to get the invoker code to generate integer arguments
+        # N.B.: The invoker code might be different for the same program with
+        # different kwargs
+        if (program not in self.kernel_stats
+                or args_tuple not in self.kernel_stats[program]):
```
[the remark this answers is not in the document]
matthiasdiener: I've implemented something like it, but it fails with `TypeError: pyopencl arrays are not hashable.`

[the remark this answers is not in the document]
matthiasdiener: Per our discussion today, I extract the argument calculation into its own method, but cache by hand.

mirgecom/profiling.py
matthiasdiener: Per our discussion today, will postpone this for a future PR.

mirgecom/profiling.py
matthiasdiener: Idea: Put `param_dict` into `args`, make the function return the modified `param_args`.

matthiasdiener: Done.

matthiasdiener:
Addressed your comments @inducer. This is ready for another review.

mirgecom/profiling.py
```diff
@@ -0,0 +1,198 @@
...
+            bytes_accessed = [v.bytes_accessed for v in value]
+            footprint_bytes = [v.footprint_bytes for v in value]
+
+            tbl.add_row([key.name, num_values, min(times),
+                int(mean(times)), max(times), min(flops), int(mean(flops)),
+                max(flops), min(bytes_accessed), int(mean(bytes_accessed)),
+                max(bytes_accessed), round(mean(bytes_accessed)/mean(times), 3),
+                mean(footprint_bytes)])
+
+        print(tbl)
```
inducer, line 116: Could return the table instead of printing it.

mirgecom/profiling.py
```diff
@@ -0,0 +1,198 @@
...
+            times = [v.time for v in value]
+            flops = [v.flops for v in value]
+            bytes_accessed = [v.bytes_accessed for v in value]
+            footprint_bytes = [v.footprint_bytes for v in value]
+
+            tbl.add_row([key.name, num_values, min(times),
+                int(mean(times)), max(times), min(flops), int(mean(flops)),
+                max(flops), min(bytes_accessed), int(mean(bytes_accessed)),
+                max(bytes_accessed), round(mean(bytes_accessed)/mean(times), 3),
+                mean(footprint_bytes)])
```
inducer, line 114: I think it might be useful to apply some number formatting, e.g. to get everything to the "Giga" range.

mirgecom/profiling.py
```diff
@@ -0,0 +1,198 @@
...
+                param_dict)
+
+            footprint_bytes = 0
+            try:
+                footprint = lp.gather_access_footprint_bytes(kernel)
+                for k, v in footprint.items():
+                    footprint_bytes += footprint[k].eval_with_dict(param_dict)
+
+            except lp.symbolic.UnableToDetermineAccessRange:
+                pass
```
inducer, line 181: From this point on, `footprint_bytes` is just wrong, so I would prefer `footprint_bytes = None`.

mirgecom/profiling.py
```diff
@@ -0,0 +1,198 @@
...
+
+            self.kernel_stats.setdefault(program, {})[args_tuple] = ProfileResult(
+                time=0, flops=flops, bytes_accessed=bytes_accessed,
+                footprint_bytes=footprint_bytes)
+
+        return self.kernel_stats[program][args_tuple]
```
inducer, line 187: In the "cache miss" case, this is a redundant lookup.

examples/wave-eager.py
```diff
@@ -111,6 +121,12 @@ def rhs(t, w):
 
 
 if __name__ == "__main__":
-    main()
+    import argparse
+    parser = argparse.ArgumentParser(description='Wave-eager (non-MPI version)')
+    parser.add_argument('--profile', action='store_true',
+        help='enable kernel profiling')
+    args = parser.parse_args()
```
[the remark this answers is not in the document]
matthiasdiener: Fixed (also in profiling.py)

mirgecom/profiling.py
matthiasdiener: Done.

mirgecom/profiling.py
matthiasdiener: Fixed.

mirgecom/profiling.py
matthiasdiener: Done. Everything is now Giga (or seconds).

mirgecom/profiling.py
matthiasdiener: Fixed.

matthiasdiener:
@inducer Ready for review.

Not sure what this error means:
`
/home/runner/work/mirgecom/mirgecom/.miniforge3/envs/testing/lib/python3.8/site-packages/mirgecom-2020.1-py3.8.egg/mirgecom/profiling.py:docstring of mirgecom.profiling.PyOpenCLProfilingArrayContext.table_profiling_data:: WARNING: py:class reference target not found: pytools.Table
`

 pytools.Table seems to have a correct docstring.

inducer:
> Edit: Added pytools.Table to nitpick ignore list.

Nah. Probably pytools is missing from the intersphinx list.

matthiasdiener:
> > Edit: Added pytools.Table to nitpick ignore list.
> 
> Nah. Probably pytools is missing from the intersphinx list.

Ok, that did it.

doc/running.rst
```diff
@@ -18,3 +18,21 @@ cache files in directories that are private to each rank by using the ``XDG_CACH
...
+
+You can use :class:`mirgecom.profiling.PyOpenCLProfilingArrayContext` instead of
+:class:`meshmode.array_context.PyOpenCLArrayContext` to profile kernel executions.
+In addition to using this array context, you also need to enable profiling in the underlying
+:class:`pyopencl.CommandQueue`, like this:
+
+.. code:: python
+
+   queue = cl.CommandQueue(cl_ctx,
+            properties=cl.command_queue_properties.PROFILING_ENABLE)
```
inducer, line 34:
```suggestion
:class:`pyopencl.CommandQueue`, like this::

   queue = cl.CommandQueue(cl_ctx,
            properties=cl.command_queue_properties.PROFILING_ENABLE)
```

doc/running.rst
```diff
@@ -18,3 +18,21 @@ cache files in directories that are private to each rank by using the ``XDG_CACH
 environment variable, such as in the following example::
 
    $ srun -n 512 bash -c 'XDG_CACHE_HOME=/p/lscratchh/diener3/xdg-scratch$SLURM_PROCID python examples/wave-eager-mpi.py'
+
+
+Profiling kernel execution
+--------------------------
+
+You can use :class:`mirgecom.profiling.PyOpenCLProfilingArrayContext` instead of
+:class:`meshmode.array_context.PyOpenCLArrayContext` to profile kernel executions.
```
inducer, line 27:
```suggestion
:class:`~meshmode.array_context.PyOpenCLArrayContext` to profile kernel executions.
```

mirgecom/profiling.py
```diff
@@ -0,0 +1,202 @@
...
+                footprint_bytes = None
+
+            res = ProfileResult(
+                time=0, flops=flops, bytes_accessed=bytes_accessed,
+                footprint_bytes=footprint_bytes)
+
+            self.kernel_stats.setdefault(program, {})[args_tuple] = res
+            return res
+
+        return self.kernel_stats[program][args_tuple]
```
inducer, line 191: I find this hard to read, because you have to read past a whole bunch of stuff to see what happens in the "cache hit" case. I'd prefer if this were structured as

```python
if CACHE_HIT:
    return

# cache miss
# stuff stuff stuff
return res
```

mirgecom/profiling.py
```diff
@@ -0,0 +1,202 @@
...
+            flops = [v.flops / 1e9 for v in value]
+            bytes_accessed = [v.bytes_accessed / 1e9 for v in value]
+            footprint_bytes = [v.footprint_bytes / 1e9
+                if v.footprint_bytes is not None else 0 for v in value]
+
+            tbl.add_row([key.name, num_values, min(times),
+                mean(times), max(times), min(flops), mean(flops),
+                max(flops), min(bytes_accessed), mean(bytes_accessed),
+                max(bytes_accessed), round(mean(bytes_accessed)/mean(times), 3),
+                mean(footprint_bytes)])
```
inducer, line 115: Will `mean` tolerate `footprint_bytes is None`?

inducer, line 115: Nobody needs more than three or four digits in these. `"%.4g"`?

mirgecom/profiling.py
```diff
@@ -0,0 +1,202 @@
...
+        for t in self.profile_events:
+            program = t.program
+            res = self._get_kernel_stats(program, t.kwargs)
+            res.time = t.cl_event.profile.end - t.cl_event.profile.start
+
+            self.profile_results.setdefault(program, []).append(res)
+
+        self.profile_events = []
+
+    def table_profiling_data(self) -> pytools.Table:
```
inducer, line 89:
```suggestion
    def tabulate_profiling_data(self) -> pytools.Table:
```
https://www.merriam-webster.com/dictionary/table (verb)

mirgecom/profiling.py
```diff
@@ -0,0 +1,202 @@
...
+        """Returns a :class:`pytools.Table` with the profiling results."""
+        self._finish_profile_events()
+
+        tbl = pytools.Table()
+
+        tbl.add_row(["Function", "Calls", "Time_min (s)", "Time_avg (s)",
+            "Time_max (s)", "GFlops_min", "GFlops_avg", "GFlops_max", "GMemAcc_min",
+            "GMemAcc_avg", "GMemAcc_max", "Bandwidth (GByte/s)",
+            "Footprint (GByte)"])
```
inducer, line 98: Unit specification is conventionally done with square brackets.

mirgecom/profiling.py
```diff
@@ -0,0 +1,202 @@
...
+                {d.name: None for d in idi if d.name not in param_dict})
+
+            # Generate the wrapper code
+            wrapper = executor.get_wrapper_generator()
+
+            gen = PythonFunctionGenerator("my_gen_args", ["param_dict"])
+
+            # Unpack dict items to local variables
+            for k, v in param_dict.items():
+                gen(f"{k}=param_dict['{k}']")
```
inducer, line 153: Is it possible to just add these as function arguments, rather than doing the arg unpacking yourself?

mirgecom/profiling.py
```diff
@@ -0,0 +1,202 @@
...
+            param_dict.update({k: None for k, value in kernel.arg_dict.items()
+                if k not in param_dict})
+
+            param_dict.update(
+                {d.name: None for d in idi if d.name not in param_dict})
+
+            # Generate the wrapper code
+            wrapper = executor.get_wrapper_generator()
+
+            gen = PythonFunctionGenerator("my_gen_args", ["param_dict"])
```
inducer, line 149: Note that this code runs in a namespace that is under user (and hence not under your) control. That means all identifiers you introduce must be prefixed somehow.

mirgecom/profiling.py
matthiasdiener: No, that's why it is filtered out above (line 108-109).

mirgecom/profiling.py
matthiasdiener: renamed to `_my_gen_args_profiling`.

mirgecom/profiling.py
matthiasdiener: Done.

mirgecom/profiling.py
matthiasdiener: Done.

mirgecom/profiling.py
matthiasdiener: Done.

mirgecom/profiling.py
matthiasdiener: I haven't found a way.

mirgecom/profiling.py
matthiasdiener: Fixed.

mirgecom/profiling.py
```diff
@@ -0,0 +1,207 @@
...
+        from statistics import mean
+
+        for key, value in self.profile_results.items():
+            num_values = len(value)
+
+            times = [v.time / 1e9 for v in value]
+            flops = [v.flops / 1e9 for v in value]
+            bytes_accessed = [v.bytes_accessed / 1e9 for v in value]
+            footprint_bytes = [v.footprint_bytes / 1e9
+                if v.footprint_bytes is not None else 0 for v in value]
```
inducer, line 109: This will throw off your min, max, mean. "0" is a really poor thing to use for "invalid value".

mirgecom/profiling.py
```diff
@@ -0,0 +1,207 @@
...
+
+            g = ".4g"
+
+            tbl.add_row([key.name, num_values, f"{min(times):{g}}",
+                f"{mean(times):{g}}", f"{max(times):{g}}", f"{min(flops):{g}}",
+                f"{mean(flops):{g}}", f"{max(flops):{g}}",
+                f"{min(bytes_accessed):{g}}", f"{mean(bytes_accessed):{g}}",
+                f"{max(bytes_accessed):{g}}",
+                f"{round(mean(bytes_accessed)/mean(times), 3):{g}}",
+                f"{mean(footprint_bytes):{g}}"])
```
inducer, line 119: Absolute gflops/gbyte numbers tell me very little. I would rather see:

* flop rate (min/max/mean)
* bandwidth (according to access count, min/max/mean)
* bandwidth (according to footprint, min/max/mean)

mirgecom/profiling.py
inducer: `_` doesn't suffice as a prefix. `_mcom_`? Also, in the previous version of the code `param_dict` was a name to which this would apply.

mirgecom/profiling.py
inducer: FTFY 1444a96

mirgecom/profiling.py
inducer: What I mean by that is that you code can crash if no value is left to take the mean of.

mirgecom/profiling.py
matthiasdiener: Fixed by only storing the non-None values (and printing "n/a" as the mean if there aren't any values at all)

mirgecom/profiling.py
matthiasdiener: Added the values you indicated, and removed the raw flops, bytes numbers.

mirgecom/profiling.py
matthiasdiener: Should be fixed now.

mirgecom/profiling.py
matthiasdiener: Thanks!

mirgecom/profiling.py
matthiasdiener: Changed prefix to `_mcom_`.

matthiasdiener:
Here is how the table looks like now:

Function            | Calls | Time_min [s] | Time_avg [s] | Time_max [s] | GFlops/s_min | GFlops/s_avg | GFlops/s_max | BWAcc_min [GByte/s] | BWAcc_mean [GByte/s] | BWAcc_max [GByte/s] | BWFoot_min [GByte/s] | BWFoot_mean [GByte/s] | BWFoot_max [GByte/s]
:-------------------|:------|:-------------|:-------------|:-------------|:-------------|:-------------|:-------------|:--------------------|:---------------------|:--------------------|:---------------------|:----------------------|:---------------------
nodes               | 8     | 4.153e-05    | 6.865e-05    | 8.172e-05    | 0.09245      | 0.7137       | 1.127        | 0.5085              | 4.167                | 6.199               | 0.07088              | 0.5808                | 0.8462
actx_special_exp    | 1     | 9.495e-05    | 9.495e-05    | 9.495e-05    | 0.07583      | 0.07583      | 0.07583      | 0.7583              | 0.7583               | 0.7583              | 0.7583               | 0.7583                | 0.7583
resample_by_picking | 132   | 3.659e-05    | 5.011e-05    | 5.936e-05    | 0            | 0            | 0            | 0.1828              | 1.83                 | 3.72                | n/a                  | n/a                   | n/a
diff                | 14    | 8.503e-05    | 8.503e-05    | 8.503e-05    | 3.387        | 3.387        | 3.387        | 22.86               | 22.86                | 22.86               | 1.289                | 1.289                 | 1.289
grudge_assign_0     | 1     | 7.333e-05    | 7.333e-05    | 7.333e-05    | 1.08         | 1.08         | 1.08         | 6.382               | 6.382                | 6.382               | 2.455                | 2.455                 | 2.455
actx_special_sqrt   | 4     | 3.186e-05    | 5.172e-05    | 6.789e-05    | 0.03013      | 0.2277       | 0.4436       | 0.1205              | 1.089                | 1.774               | 0.1205               | 1.089                 | 1.774
grudge_assign_0     | 12    | 7.496e-05    | 7.496e-05    | 7.496e-05    | 0.5763       | 0.5763       | 0.5763       | 4.803               | 4.803                | 4.803               | 3.842                | 3.842                 | 3.842
resample_by_mat     | 24    | 5.237e-05    | 5.237e-05    | 5.237e-05    | 1.027        | 1.027        | 1.027        | 8.469               | 8.469                | 8.469               | n/a                  | n/a                   | n/a
diff                | 6     | 3.344e-05    | 5.931e-05    | 8.63e-05     | 0.2296       | 1.69         | 2.838        | 1.206               | 8.872                | 14.9                | 0.1186               | 0.8475                | 1.421
grudge_assign_0     | 12    | 6.988e-05    | 6.988e-05    | 6.988e-05    | 0.3091       | 0.3091       | 0.3091       | 1.855               | 1.855                | 1.855               | 1.855                | 1.855                 | 1.855
face_mass           | 12    | 5.921e-05    | 5.921e-05    | 5.921e-05    | 3.283        | 3.283        | 3.283        | 19.58               | 19.58                | 19.58               | 1.354                | 1.354                 | 1.354
grudge_assign_0     | 12    | 5.055e-05    | 5.055e-05    | 5.055e-05    | 0.2848       | 0.2848       | 0.2848       | 2.136               | 2.136                | 2.136               | 2.136                | 2.136                 | 2.136
diff                | 12    | 6.841e-05    | 6.841e-05    | 6.841e-05    | 2.105        | 2.105        | 2.105        | 14.21               | 14.21                | 14.21               | 1.064                | 1.064                 | 1.06

inducer:
> Here is how the table looks like now:

:+1: Nice!

matthiasdiener:
Fixed pydocstyle, ready for review.

mirgecom/profiling.py
```diff
@@ -0,0 +1,229 @@
...
+          .to_bytes().eval_and_sum(domain_params)
+
+        flops = op_map.filter_by(dtype=[np.float32, np.float64]).eval_and_sum(
+            domain_params)
+
+        footprint_bytes = 0
+        try:
+            footprint = lp.gather_access_footprint_bytes(kernel)
+            for k, v in footprint.items():
+                footprint_bytes += footprint[k].eval_with_dict(domain_params)
```
inducer, line 207:
```suggestion
        try:
            footprint = lp.gather_access_footprint_bytes(kernel)
            footprint_bytes = sum(footprint[k].eval_with_dict(domain_params) for k in footprint)
```

mirgecom/profiling.py
```diff
@@ -0,0 +1,229 @@
...
+        """Return a :class:`pytools.Table` with the profiling results."""
+        self._finish_profile_events()
+
+        tbl = pytools.Table()
+
+        tbl.add_row(["Function", "Calls",
+            "Time_min [s]", "Time_avg [s]", "Time_max [s]",
+            "GFlops/s_min", "GFlops/s_avg", "GFlops/s_max",
+            "BWAcc_min [GByte/s]", "BWAcc_mean [GByte/s]", "BWAcc_max [GByte/s]",
+            "BWFoot_min [GByte/s]", "BWFoot_mean [GByte/s]", "BWFoot_max [GByte/s]"])
```
inducer, line 107: I know this is already a pretty column-heavy table, but I *think* it could be useful to also print estimates of intensity (i.e. flops/byte).

mirgecom/profiling.py
```diff
@@ -0,0 +1,229 @@
...
+
+            flops_per_sec = [f / t for f, t in zip(flops, times)]
+
+            bytes_accessed = [v.bytes_accessed / 1e9 for v in value]
+            footprint_bytes = [v.footprint_bytes / 1e9 for v in value
+                if v.footprint_bytes is not None]
+
+            bandwidth_access = [b / t for b, t in zip(bytes_accessed, times)]
+
+            if len(footprint_bytes) > 0:
```
inducer, line 128:
```suggestion
            if footprint_bytes:
```

mirgecom/profiling.py
```diff
@@ -0,0 +1,229 @@
...
+                bandwidth_footprint = [b / t for b, t in zip(footprint_bytes, times)]
+                bandwidth_footprint_val = {"min": f"{min(bandwidth_footprint):{g}}",
+                    "mean": f"{mean(bandwidth_footprint):{g}}",
+                    "max": f"{max(bandwidth_footprint):{g}}"}
```
inducer, line 132: Numpy's masked arrays do this with less fuss:
```py
>>> ary = np.ma.masked_equal([12, 13, None], None)
>>> ary
masked_array(data=[12, 13, --],
             mask=[False, False,  True],
       fill_value='?',
            dtype=object)
>>> np.mean(ary)
12.5
>>> ary = np.ma.masked_equal([None,None, None], None)
>>> f"{np.mean(ary):.4g}"
'--'
>>> 
```

mirgecom/profiling.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def _get_kernel_stats(self, program, kwargs: dict) -> ProfileResult:
+        # We need a tuple to index the cache
+        args_tuple = tuple(
+            (key, value.shape) if hasattr(value, "shape") else (key, value)
+            for key, value in kwargs.items())
+
+        # Are kernel stats already in the cache?
+        if program in self.kernel_stats and args_tuple in self.kernel_stats[program]:
+            return self.kernel_stats[program][args_tuple]
```
inducer, line 156: IMO the Python-idiomatic way would be like this:
```py
try:
    return self.kernel_stats[program][args_tuple]
except KeyError:
```
Then the fast (cache hit) path does not involve redundant lookups.

mirgecom/profiling.py
```diff
@@ -0,0 +1,229 @@
...
+
+        kernel = executor.get_typed_and_scheduled_kernel(
+            executor.arg_to_dtype_set(kwargs))
+
+        idi = info.implemented_data_info
+
+        types = {k: v for k, v in kwargs.items()
+            if hasattr(v, "dtype") and not v.dtype == object}
+
+        param_dict = {**kwargs}
```
inducer, line 171:
```suggestion
        param_dict = kwargs.copy()
```
?

mirgecom/profiling.py
```diff
@@ -0,0 +1,229 @@
...
+        param_dict.update({k: None for k, value in kernel.arg_dict.items()
```
inducer, line 172:
```suggestion
        param_dict.update({k: None for k in kernel.arg_dict.keys()
```

inducer (commented):
Sorry, spotted even more. Mostly simple stuff. Thanks for your patience.

mirgecom/profiling.py
matthiasdiener: Fixed.

mirgecom/profiling.py
matthiasdiener: Added (just the mean for now).

mirgecom/profiling.py
matthiasdiener: Good point, fixed.

mirgecom/profiling.py
matthiasdiener: Fixed. (np.min/np.max are unfortunately a bit harder than np.mean)

matthiasdiener (commented):
Thanks, should be ready for another review.

inducer (approved):

PR #31, opened 2020-07-23
Workaround for file locking issues with large numbers of ranks

matthiasdiener:
Fixes https://github.com/illinois-ceesd/planning/issues/27


mirgecom/__init__.py
```diff
@@ -25,26 +25,23 @@
 import sys
 from warnings import warn
 
-try:
-    from mpi4py import MPI
+from mpi4py import MPI
```
inducer, line 28: We shouldn't unconditionally import `mpi4py`. Running mirgecom sequentially without MPI is entirely legitimate IMO (but let's chat about that in the dev meeting tomorrow, added to agenda).

I get that that makes parallel runs tougher to detect, but, as a fallback not requiring MPI, we could also fish around for some env vars set by a queueing system for evidence of large runs.

mirgecom/__init__.py
```diff
@@ -25,26 +25,23 @@
...
+    if "XDG_CACHE_HOME" in os.environ:
+        warn("XDG_CACHE_HOME already set, not adjusting it.")
+    else:
+        rank = MPI.COMM_WORLD.Get_rank()
+        username = getpass.getuser()
 
-            os.environ["XDG_CACHE_HOME"] = "/tmp/{}/xdg-cache-r{}".format(username, rank)
-except:
-    pass
+        os.environ["XDG_CACHE_HOME"] = "/tmp/{}/xdg-cache-r{}".format(username, rank)
```
inducer, line 47: We should conceivably create this directory.

mirgecom/__init__.py
```diff
@@ -25,26 +25,23 @@
...
+size = MPI.COMM_WORLD.Get_size()
 
-    if size >= 128:
-        for mod in ["pyopencl", "pytools", "loopy"]:
-            if mod in sys.modules:
-                warn("{} already loaded, not adjusting XDG_CACHE_HOME.".format(mod))
+if size >= 128:
+    for mod in ["pyopencl", "pytools", "loopy"]:
+        if mod in sys.modules:
+            warn("{} already loaded, not adjusting XDG_CACHE_HOME.".format(mod))
```
inducer, line 39: This warning is too opaque. Write for a  hapless student who might have no idea what any of it means, but whose runs are dog-slow.

mirgecom/__init__.py
```diff
@@ -25,26 +25,23 @@
...
 
-        if "XDG_CACHE_HOME" in os.environ:
-            warn("XDG_CACHE_HOME already set, not adjusting it.")
-        else:
-            rank = MPI.COMM_WORLD.Get_rank()
-            username = getpass.getuser()
+    if "XDG_CACHE_HOME" in os.environ:
+        warn("XDG_CACHE_HOME already set, not adjusting it.")
```
inducer, line 42: IMO, the expectation should be that a user should set this variable in their job script for large runs, and the warning above should teach them to do so. So this code path here IMO is the expected case and should not raise a warning.

mirgecom/__init__.py
matthiasdiener: From my tests, these directories seem to be created automatically. Is it necessary to create them ourselves?

inducer: Don't know. Check the XDG spec?

mirgecom/__init__.py
matthiasdiener: I think the consensus in the meeting was to always require mpi4py?

mirgecom/__init__.py
matthiasdiener: Made the warning more verbose.

mirgecom/__init__.py
matthiasdiener: Removed the warning.

mirgecom/__init__.py
matthiasdiener: XDG spec is unclear, creating the dir just to be safe.

mirgecom/__init__.py
inducer: Yes, agreed.

mirgecom/__init__.py
```diff
@@ -19,3 +19,30 @@
...
+# large numbers of ranks. See https://github.com/illinois-ceesd/planning/issues/27
+# for details.
+
+size = MPI.COMM_WORLD.Get_size()
+
+if size >= 128:
+    for mod in ["pyopencl", "pytools", "loopy"]:
+        if mod in sys.modules:
+            warn("{} already loaded, not adjusting XDG_CACHE_HOME. Please "
+                 "import mirgecom before this module.".format(mod))
```
inducer, line 40: Use an f-string.

mirgecom/__init__.py
```diff
@@ -19,3 +19,30 @@
...
+
+    if "XDG_CACHE_HOME" not in os.environ:
+        rank = MPI.COMM_WORLD.Get_rank()
+        username = getpass.getuser()
+
+        xdg_cache_path = "/tmp/{}/xdg-cache-r{}".format(username, rank)
+        os.makedirs(xdg_cache_path, exist_ok=True)
+        os.environ["XDG_CACHE_HOME"] = xdg_cache_path
```
inducer, line 48: If this is not set, raise a warning to teach the user to set `XDG_CACHE_HOME` in their job script.

mirgecom/__init__.py
inducer, line 40: - I don't think relying on this import side effect is a good default policy. It can help catch mistakes, but I think generally folks should set the env var in their job script. So this warning should teach them to do that.
- Also, if the env var is set, we should assume that the user has taken care of the issue.

mirgecom/__init__.py
```diff
@@ -19,3 +19,30 @@
...
+    for mod in ["pyopencl", "pytools", "loopy"]:
+        if mod in sys.modules:
+            warn("{} already loaded, not adjusting XDG_CACHE_HOME. Please "
+                 "import mirgecom before this module.".format(mod))
+
+    if "XDG_CACHE_HOME" not in os.environ:
+        rank = MPI.COMM_WORLD.Get_rank()
+        username = getpass.getuser()
+
+        xdg_cache_path = "/tmp/{}/xdg-cache-r{}".format(username, rank)
```
inducer, line 46: This is a predictable pattern and hence insecure. Use secure temp file creation.

mirgecom/__init__.py
matthiasdiener: done.

matthiasdiener: Fixed.

mirgecom/__init__.py
matthiasdiener: Fixed.

mirgecom/__init__.py
matthiasdiener: Fixed.

mirgecom/__init__.py
```diff
@@ -19,3 +19,43 @@
...
+rank = MPI.COMM_WORLD.Get_rank()
+
+if size >= 256 and rank == 0:
+    if "XDG_CACHE_HOME" not in os.environ:
+        skip = False
+
+        for mod in ["pyopencl", "pytools", "loopy"]:
+            if mod in sys.modules:
+                warn(f"{mod} already loaded, not adjusting XDG_CACHE_HOME. Please "
+                  "see https://github.com/illinois-ceesd/planning/issues/27 for "
```
inducer, line 45: This is not a publicly accessible URL. Could you add a section to the docs explaining this and link to that?

(same above)

mirgecom/__init__.py
```diff
@@ -19,3 +19,43 @@
...
+                  "information on how to avoid file system overheads by setting "
+                  "the XDG_CACHE_HOME variable in your job scripts.")
+                skip = True
+                break
+
+        if not skip:
+            username = getpass.getuser()
+            dir_prefix = f"/tmp/{username}/"
+            os.makedirs(dir_prefix, exist_ok=True)
```
inducer, line 54: This is still predictable, and I think it can still be used for arbitrary file creation across accounts (by planting a symlink) in `/tmp`. Plus `/tmp` isn't always the directory for temporary files. Just leave as much as possible of this to `tempfile`.

matthiasdiener: Instead of attempting to work around this issue here, would it be enough to just warn the user and point them how to fix it themselves?

mirgecom/__init__.py
matthiasdiener: Added to docs.

mirgecom/__init__.py
matthiasdiener: Implemented the warning only.

doc/running.rst
```diff
@@ -0,0 +1,25 @@
...
+The PyOpenCL package used in MirgeCOM caches results of compilation on the hard
+disk in order to speed up multiple executions of the same kernel. This can lead
+to slowdowns on startup when executing on many ranks and/or nodes due to concurrent
+hard disk accesses. We have seen these slowdowns starting with 256 ranks/nodes.
+An indicator of concurrency issues are warnings like these:
+
+.. code::
+
+  .conda/envs/dgfem/lib/python3.8/site-packages/pyopencl/cache.py:101: UserWarning:
+  could not obtain cache lock--delete '.cache/pyopencl/pyopencl-compiler-cache-v2-py3.8.3.final.0/lock' if necessary
```
inducer, line 16: 4-wide indent (below too)

doc/running.rst
```diff
@@ -0,0 +1,25 @@
+Running MirgeCOM
+================
+
+Running with large numbers of ranks and nodes
+---------------------------------------------
+
+The PyOpenCL package used in MirgeCOM caches results of compilation on the hard
```
inducer, line 7: Not just PyOpenCL

doc/running.rst
```diff
@@ -0,0 +1,25 @@
...
+disk in order to speed up multiple executions of the same kernel. This can lead
+to slowdowns on startup when executing on many ranks and/or nodes due to concurrent
+hard disk accesses. We have seen these slowdowns starting with 256 ranks/nodes.
```
inducer, line 10: I would say this is good practice for any distributed-memory run.

mirgecom/__init__.py
```diff
@@ -19,3 +19,22 @@
...
+
+# This code avoids slow startups due to file system locking when running with
+# large numbers of ranks. See
+# https://mirgecom.readthedocs.io/en/latest/running.html#running-with-large-numbers-of-ranks-and-nodes
+# for more details
+
+size = MPI.COMM_WORLD.Get_size()
+rank = MPI.COMM_WORLD.Get_rank()
+
+if size >= 256 and rank == 0 and "XDG_CACHE_HOME" not in os.environ:
```
inducer, line 36:
```suggestion
if size > 1 and rank == 0 and "XDG_CACHE_HOME" not in os.environ:
```

doc/running.rst
matthiasdiener: Isn't the offical indentation for rst 3 spaces? Changed to 3.

doc/running.rst
matthiasdiener: fixed.

doc/running.rst
matthiasdiener: removed the 256 ranks phrase

PR #36, opened 2020-07-24
Array context

inducer:
cc @matthiasdiener

### DIFF
--- a/mirgecom/wave.py
+++ b/mirgecom/wave.py
@@ -23,11 +23,11 @@ THE SOFTWARE.
 import numpy as np
 import numpy.linalg as la  # noqa
 from pytools.obj_array import (
-    join_fields, make_obj_array,
-    with_object_array_or_scalar)
+    flat_obj_array, make_obj_array)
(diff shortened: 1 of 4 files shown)


mirgecom/wave.py
```diff
@@ -23,70 +23,71 @@
...
         0.5*normal_times(v_jump),
         )
 
-    return discr.interp(w_tpair.dd, "all_faces", c*flux_weak)
+    return discr.project(w_tpair.dd, "all_faces", c*flux_weak)
+
 
+def wave_operator(actx, discr, c, w):
+    """
+    Args:
```
majosm, line 65: Probably want to add argument documentation for `actx` here?

Would it make any sense to store a reference to the array context inside the discretization object, in order to avoid always having to pass both? (I don't have a good sense for exactly how these two relate yet...)

inducer: Good catch! Passing an (extra) array context really doesn't make any sense, as the state vector/trace pair already have one on board. Fixed in dde43d4.

Issue #18, opened 2020-06-30
Should run examples as part of CI


inducer:
cc @MTCam 

Created because #7 creates examples. (which is a good thing, but if nothing ensures they keep running, they won't)

[Example CI script](https://github.com/inducer/meshmode/blob/7826fa5e13854bf1dae425b4226865acc10ee01f/.github/workflows/ci.yml#L41-L54)

MTCam:
I added a runner and a CI for running the example.  The example will fail if the solution diverges from the expected/exact solution.  In the next iteration I would like to add command line arguments to set which case to run, and other params.

MTCam:
#55 adds examples running to CI for mirgecom
[This PR](https://github.com/illinois-ceesd/emirge/pull/68) adds examples running to CI for emirge

matthiasdiener:
Isn't this done already?

MTCam:
Examples run as a part of CI, for both mirgecom, and emrige.  That is, emirge CI runs mirgecom examples to verify correct installation.

PR #37, opened 2020-07-27
fix wave-eager arraycontext

matthiasdiener:


examples/wave-eager.py
```diff
@@ -94,7 +96,7 @@ def rhs(t, w):
         fields = rk4_step(fields, t, dt, rhs)
 
         if istep % 10 == 0:
-            print(istep, t, la.norm(fields[0].get()))
+            print(istep, t)
```
inducer, line 99: You could call `discr.norm(fields[0], np.inf)` here.

examples/wave-eager.py
```diff
@@ -24,35 +24,37 @@
...
 from grudge.eager import EagerDGDiscretization
 from grudge.shortcuts import make_visualizer
 from mirgecom.wave import wave_operator
 from mirgecom.integrators import rk4_step
+from meshmode.array_context import PyOpenCLArrayContext
+from meshmode.dof_array import thaw
 
 
-def bump(discr, queue, t=0):
+def bump(discr, actx, t=0):
```
inducer, line 36: I've stuck to the convention `actx, discr` in the calling sequence in grudge otherwise.

.github/workflows/tests.yaml
```diff
@@ -68,3 +68,4 @@ jobs:
             . "$MINIFORGE_INSTALL_DIR/bin/activate" testing
             cd test
             python -m pytest --durations=10 --tb=native --junitxml=pytest.xml --doctest-modules -rxsw .
+            python examples/wave-eager.py
```
inducer, line 71: We should really run *all* examples, as a separate job. (cf #18 )

matthiasdiener: There is only one example currently (?). I separated that out as test step for now until we decide how to proceed with #18.

examples/wave-eager.py
matthiasdiener: Fixed, thanks.

examples/wave-eager.py
matthiasdiener: Fixed, thanks.

PR #39, opened 2020-07-27
add readthedocs integration

matthiasdiener:
Here is how it looks like: https://mirgecom.readthedocs.io/en/rtd/

Fixes https://github.com/illinois-ceesd/planning/issues/62

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -14,8 +14,6 @@
 # import sys
 # sys.path.insert(0, os.path.abspath('.'))
 
-autodoc_mock_imports = ["mpi4py"]
-
 
 # -- Project information -----------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


doc/conf.py
```diff
@@ -61,7 +61,7 @@
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = 'alabaster'
+html_theme = 'sphinx_rtd_theme'
```
inducer, line 64: Is this required?

matthiasdiener: No, but doesn't it look nicer? sphinx_rtd_theme is the "standard" RTD theme, alabaster is here: https://alabaster.readthedocs.io/en/latest/

inducer: I kind of like alabaster better. :)

inducer: I've added that to the agenda for Friday. :P

matthiasdiener: > I've added that to the agenda for Friday. :P

:cold sweat:

I'll set it back to alabaster 😃

inducer:
I didn't see a strong reason not to, so I just went and merged this. Shout if that was a bad idea.

PR #43, opened 2020-07-30
Work around pip breakage

inducer:
Use 'python setup.py install' rather than 'pip install .' to avoid spurious installation of old loopy.

x-ref: https://gitlab.tiker.net/inducer/ci-support/-/commit/fc8c4b779b34404832b0426aa8f49c445d3e3a2a

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -60,7 +60,7 @@ jobs:
             MINIFORGE_INSTALL_DIR=.miniforge3
             . "$MINIFORGE_INSTALL_DIR/bin/activate" testing
             pip install -r requirements.txt
-            pip install -e .
+            python setup.py install
 
         - name: Run tests
           shell: bash -l {0}


MTCam:
Will this change our development workflow?  I am frequently `pip install -e .` in all the relevant packages to switch back and forth between pre-array-context and master.

inducer:
> Will this change our development workflow? I am frequently pip install -e . in all the relevant packages to switch back and forth between pre-array-context and master.

No, in principle it shouldn't. A new pip version of pip was [recently released](https://pip.pypa.io/en/stable/news/#id1), and I think it broke something. If you look at [this failing run](https://github.com/illinois-ceesd/mirgecom/runs/928818261), it installs loopy from git just fine, and then, during `pip install .` it goes and downloads an old-as-dirt version of loopy from the package index and overwrites the existing once, hence all the recent CI failures. Entertainingly, as it does so, it loses track of what it installed and claims the new (git) version is still there.

cc @matthiasdiener for good measure, in case we need to update the install docs to work around this nonsense.

inducer:
(Could one of you OK this? Without it, all of our CIs are broken.)

inducer:
Thanks!

PR #44, opened 2020-07-30
Fix euler array context

inducer:
If you're trying to run this at home, note that this relies on:
- https://github.com/inducer/meshmode/pull/22
- https://github.com/inducer/grudge/pull/10

### DIFF
--- a/test/test_euler.py
+++ b/test/test_euler.py
@@ -423,7 +423,7 @@ def test_facial_flux(ctx_factory, order, dim):
         # The off-diagonals are zero. We get a {ndim}-vector for each
         # dimension, the flux for the x-component of momentum (for example) is:
         # f_momx = < 1.0, 0 , 0> , then we return f_momx .dot. normal, which
-        # can have negative components values.
+        # can introduce negative values.
         #
         # (Explanation courtesy of Mike Campbell,
         # https://github.com/illinois-ceesd/mirgecom/pull/44#discussion_r463304292)
--- a/test/test_euler.py
+++ b/test/test_euler.py
@@ -395,15 +395,11 @@ def test_facial_flux(ctx_factory, order, dim):
 
         discr = EagerDGDiscretization(actx, mesh, order=order)
 
-        mass_input = discr.zeros(actx)
-        energy_input = discr.zeros(actx)
+        mass_input = discr.zeros(actx) + 1.0
+        energy_input = discr.zeros(actx) + 2.5
         mom_input = flat_obj_array(
             [discr.zeros(actx) for i in range(discr.dim)]
         )
-
-        # This sets p = 1
-        mass_input[:] = 1.0
-        energy_input[:] = 2.5
 
         fields = flat_obj_array(
             mass_input, energy_input, mom_input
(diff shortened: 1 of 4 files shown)


inducer:
> This is so great. Thanks a million! I am surprised that this didn't need updating.

That was https://github.com/inducer/meshmode/pull/22. `DOFArray < DOFArray` (really `object_array < object_array`) produces some nonsensical numpy weirdness. That PR papers over that by hand-implementing the desired semantics.

test/test_euler.py
```diff
@@ -395,58 +395,33 @@ def test_facial_flux(ctx_factory, order, dim):
...
-            if err > maxmomerr:
-                maxmomerr = err
+        iff_split = split_conserved(dim, interior_face_flux)
+        assert fnorm(iff_split.mass) < tolerance
+        assert fnorm(iff_split.energy) < tolerance
 
-        assert err < tolerance
+        # FIXME (AK -> MC): explain why this is right
+        momerr = fnorm(iff_split.momentum) - p0
+        assert momerr < tolerance
```
inducer, line 422: This stuff (and the equivalent bit below) I wasn't feeling too confident about. @MTCam Could you double check that I did this right?

test/test_euler.py
```diff
@@ -395,58 +395,33 @@ def test_facial_flux(ctx_factory, order, dim):
...
-            err = np.abs(err - p0)
-            if err > maxmomerr:
-                maxmomerr = err
+        iff_split = split_conserved(dim, interior_face_flux)
+        assert fnorm(iff_split.mass) < tolerance
+        assert fnorm(iff_split.energy) < tolerance
 
-        assert err < tolerance
+        # FIXME (AK -> MC): explain why this is right
+        momerr = fnorm(iff_split.momentum) - p0
```
inducer, line 421: This particularly seemed weird, given that the array seemed to contain values of 0, 1, and -1.

inducer: (I.e. could you explain, maybe by adding a comment there?)

MTCam:
> > This is so great. Thanks a million! I am surprised that this didn't need updating.
> 
> That was [inducer/meshmode#22](https://github.com/inducer/meshmode/pull/22). `DOFArray < DOFArray` (really `object_array < object_array`) produces some nonsensical numpy weirdness. That PR papers over that by hand-implementing the desired semantics.

COol cool.   Thanks again - this is huge for me. When this gets merged-->merged-->mtc/euler, then mtc/euler will be on *master* for grudge, meshmode, et al.   A giant reduction in my rate of hair loss.

test/test_euler.py
MTCam: Oh sure.  It is because the expected pressure 1.0 (by design).   And the flux diagonal is `[rho(v x v) + p]` since we have zero velocities it's just `p`.

inducer: But what about the zero and the sign flip?

MTCam: The off-diagonals are zero. We get a {ndim}-vector for each dimension, the flux for the x-component of momentum (for example) is:
f_momx = < 1.0, 0 , 0> , then we return f_momx .dot. normal, which can have negative components.

MTCam: I dunno what happened there.  I hit the wrong button when trying to comment and it sent me into review mode.

inducer: Thanks! Added your explanation in dfa591f.

test/test_euler.py
MTCam: It looks right to me.   You replaced some giant ugly loops with succinct norms that do a better job.  As long as the norms don't just zero out the arrays, I suspect it's fine :)

MTCam: Actually, even if they did just zero out these arrays, some tests would be failing due to that.

MTCam:
Can this get merged?

MTCam:
FYI, after merging this to mtc/array-context, and mtc/array-context to mtc/euler.  All merges are clean, and all tests pass.  Just tried this out on my laptop in a test branch - didn't push the button yet.

inducer:
Btw, thanks for the wording fix!

PR #45, opened 2020-07-31
Propagate failure in `run_examples.sh`

inducer:


inducer:
> Any chance you'd mind if we just capture it, but let all the examples run before returning the error?

Totally fine. This was just the quick and lazy way.

PR #47, opened 2020-08-03
Slipwall

MTCam:
This is implementation-only at this point.  Just trying to cover our bases for the upcoming demo on the Y0 grid.  This PR is associated with [this planning issue](https://github.com/illinois-ceesd/planning/issues/82).


MTCam:
Ported to master+, reopened in #85

PR #51, opened 2020-08-07
Add fparser to requirements.txt

inducer:


requirements.txt
```diff
@@ -12,3 +12,6 @@ git+https://github.com/inducer/meshpy.git
 git+https://github.com/inducer/modepy.git
 git+https://github.com/inducer/meshmode.git
 git+https://github.com/inducer/grudge.git
+
+# Toy Fortran parser for Loopy
+git+https://github.com/inducer/f2py.git
```
inducer, line 17: @matthiasdiener Does this mean that emirge will produce a checkout of this? I'd say that's not necessary.

matthiasdiener: Yes, it would. We could just write `f2py` instead to use pip.

matthiasdiener:
can you add pudb too?

requirements.txt
inducer: Unfortunately, `f2py` doesn't exist as a package on the package index: https://pypi.org/search/?q=f2py

To explain the context of this a bit: f2py is a Fortran->Python wrapper generator that ships with numpy. Due to limitations of the old code base (no F90), Pearu (Peterson, original author of f2py) decided to restart from scratch, starting with a Fortran parser. That's as far as he got, and a forked version of this is in that repo.

So:
- From git is the only way to install it
- But I don't think we want a checkout of it by default

-> https://github.com/illinois-ceesd/emirge/issues/58

requirements.txt
```diff
@@ -12,3 +13,6 @@ git+https://github.com/inducer/meshpy.git
 git+https://github.com/inducer/modepy.git
 git+https://github.com/inducer/meshmode.git
 git+https://github.com/inducer/grudge.git
+
+# Toy Fortran parser for Loopy
+f2py
```
inducer: I don't think this exists. And even if, we want extra changes that I've made on my fork.

requirements.txt
```diff
@@ -1,6 +1,7 @@
+mpi4py
 numpy
+pudb
```
inducer, line 3: I don't think this quite belongs here, since it's not a functional requirement of mirgecom.

Counter-proposal: https://github.com/illinois-ceesd/emirge/pull/59

requirements.txt
matthiasdiener: Yeah, I changed it back already.

requirements.txt
matthiasdiener: Good point, will remove it here.

inducer:
Added some `#egg=` annotations prompted by https://github.com/illinois-ceesd/emirge/issues/58#issuecomment-670589502. Good to merge from my end if CIs pass.

matthiasdiener:
.... and we broke emirge ... :-(

inducer:
Argh, sorry! Maybe it'd be a good idea to have a CI that tries emirge? Or trigger an emirge build for each change in mirgecom? (Can Github do that? Gitlab can, sort of)

inducer:
Should we revert this for the time being?

matthiasdiener:
> Should we revert this for the time being?

I think just removing the comments would be fine.

PR #55, opened 2020-08-09
Add example runner script

MTCam:
I would like to add this examples runner script so that teesd can expect its presence regardless of branch.

### DIFF
--- a/examples/run_examples.sh
+++ b/examples/run_examples.sh
@@ -2,11 +2,8 @@
 
 # set -e
 
-examples_dir=${1}
-if [ -z ${examples_dir} ]
-then
-    examples_dir=`pwd`
-fi
+origin=$(pwd)
+examples_dir=${1-$origin}
 declare -i exitcode=0
 printf "Running examples in ${examples_dir}...\n"
 for example in ${examples_dir}/*.py
@@ -19,13 +16,12 @@ do
         printf "Running serial example: ${example}\n"        
         python ${example}
     fi
-    printf "Example ${example} "
     if [ $? -eq 0 ]
     then
-        printf "succeeded.\n"
+        printf "Example ${example} succeeded.\n"
     else
         ((exitcode=exitcode+1))
-        printf "failed.\n"
+        printf "Example ${example} failed.\n"
     fi
 done
 printf "Done running examples!\n"
--- a/examples/run_examples.sh
+++ b/examples/run_examples.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+# set -e
+
+examples_dir=${1}
+if [ -z ${examples_dir} ]
+then
+    examples_dir=`pwd`
+fi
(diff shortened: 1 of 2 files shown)


examples/run_examples.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+# set -e
+
```
matthiasdiener, line 4:
```suggestion
set -o nounset

```

examples/run_examples.sh
```diff
@@ -0,0 +1,35 @@
...
+do
+    if [[ "${example}" == *"mpi"* ]]
+    then
+        printf "Running parallel example: ${example}\n"        
+        mpiexec -n 2 python ${example}
+    else
+        printf "Running serial example: ${example}\n"        
+        python ${example}
+    fi
+    if [ $? -eq 0 ]
```
matthiasdiener, line 19:
```suggestion
    if [[ $? -eq 0 ]]
```

examples/run_examples.sh
```diff
@@ -0,0 +1,36 @@
...
+# set -e
+set -o nounset
+
+origin=$(pwd)
+examples_dir=${1-$origin}
+declare -i exitcode=0
+echo "Running examples in ${examples_dir}..."
+for example in ${examples_dir}/*.py
+do
+    if [[ "${example}" == *"mpi"* ]]
```
matthiasdiener, line 12:
```suggestion
    if [[ "$example" == *"mpi"* ]]
```

examples/run_examples.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+# set -e
+set -o nounset
+
+origin=$(pwd)
+examples_dir=${1-$origin}
+declare -i exitcode=0
+echo "Running examples in ${examples_dir}..."
+for example in ${examples_dir}/*.py
```
matthiasdiener, line 10:
```suggestion
for example in $examples_dir/*.py
```

examples/run_examples.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+# set -e
+set -o nounset
+
+origin=$(pwd)
+examples_dir=${1-$origin}
+declare -i exitcode=0
+echo "Running examples in ${examples_dir}..."
```
matthiasdiener, line 9:
```suggestion
echo "Running examples in $examples_dir ..."
```

examples/run_examples.sh
matthiasdiener: Not sure if this `if` should be a bit more strict here (maybe `*-mpi*`?). It currently matches the word `compile` for example.

MTCam: Yeah, it could more strict for sure.  We could just require that MPI examples are `\<whatever\>-mpi.py`.  Does that sound reasonable?

PR #61, opened 2020-08-11
Add wave-eager-mpi

### LINKED ISSUE #41: Add MPI example
MPI-enabled version of [wave-eager.py](https://github.com/illinois-ceesd/mirgecom/blob/master/examples/wave-eager.py).

matthiasdiener:
Fixes #41.

Mostly a copy of https://github.com/inducer/grudge/blob/master/examples/wave/wave-eager-mpi.py

### DIFF
--- a/examples/wave-eager-mpi.py
+++ b/examples/wave-eager-mpi.py
@@ -1,5 +1,3 @@
-from __future__ import division, print_function
-
 __copyright__ = "Copyright (C) 2020 Andreas Kloeckner"
 
 __license__ = """
--- a/examples/wave-eager-mpi.py
+++ b/examples/wave-eager-mpi.py
@@ -1,4 +1,4 @@
-__copyright__ = "Copyright (C) 2020 Andreas Kloeckner"
+__copyright__ = "Copyright (C) 2020 University of Illinois Board of Trustees"
 
 __license__ = """
 Permission is hereby granted, free of charge, to any person obtaining a copy
--- a/examples/wave-eager-mpi.py
+++ b/examples/wave-eager-mpi.py
@@ -27,83 +27,18 @@ import numpy as np
 import numpy.linalg as la  # noqa
 import pyopencl as cl
 
-from pytools.obj_array import flat_obj_array, make_obj_array
+from pytools.obj_array import flat_obj_array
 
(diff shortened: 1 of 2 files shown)


examples/wave-eager-mpi.py
```diff
@@ -0,0 +1,139 @@
+__copyright__ = "Copyright (C) 2020 Andreas Kloeckner"
```
majosm, line 1:
```suggestion
__copyright__ = "Copyright (C) 2020 University of Illinois Board of Trustees"
```

matthiasdiener: @inducer Ok for you?

inducer: Yes, fine.

examples/wave-eager-mpi.py
```diff
@@ -0,0 +1,139 @@
...
+from meshmode.array_context import PyOpenCLArrayContext
+from meshmode.dof_array import thaw
+
+from meshmode.mesh import BTAG_ALL, BTAG_NONE  # noqa
+
+from grudge.eager import EagerDGDiscretization
+from grudge.shortcuts import make_visualizer
+from mirgecom.integrators import rk4_step
+from mirgecom.wave import wave_operator
+from mpi4py import MPI
```
inducer, line 39: I would suggest we delete the non-MPI version then.

matthiasdiener: Imho, we could keep both. MPI/mpi4py has some additional challenges compared to serial mirgecom so that the non-MPI variant could be useful.

inducer: Maintaining code is not free. The more code we produce, the more there is to maintain.

MTCam: I agree with both of you.  Not free to maintain, but I would argue that it is worth it.  It is instructive to compare the non-mpi version to the mpi version.  If nothing else, this side-by-side comparison is some testament to the advantages of the general approach.

Additionally, it allows someone who does not have MPI (god help them) still run some examples.

inducer: :shrug: OK, let's see how it plays out. :)

Issue #58, opened 2020-08-11
Figure out GPU assignment strategy


inducer:
I suggest we do this in code, having all MPI ranks on a node figure out what GPUs are available and making sure each rank gets one. This is a simple bit of MPI code that we can write that works generically.

MTCam:
I think this is related to #56, figuring out platform-specific parallel spawn.

matthiasdiener:
> code

Should this go into `mirgecom/__init__.py` ?

inducer:
> Should this go into `mirgecom/__init__.py`?

IMO it should just be a utility that simulations call that uses MPI collectives internally. In case that's what you meant, I'm not a fant of doing things as a side effect of importing a module. Doing that has come back to bite me every time.

inducer:
We also need GPU names that are valid across ranks. We can try GPU indices at first, but I'm not sure we're guaranteed that each rank will enumerate GPUs in the same order.

Nvidia has a CL extension that [exposes PCI bus/slot IDs](https://github.com/inducer/pyopencl/blob/d4bbc613318c81936d1df2aea1bf5909e75b0258/src/wrap_cl.hpp#L782-L789). If needed, we can have POCL expose the same thing. (But even then, this is only going to be Nv-specific.)

matthiasdiener:
> We also need GPU names that are valid across ranks. We can try GPU indices at first, but I'm not sure we're guaranteed that each rank will enumerate GPUs in the same order.

But why would we need that?

matthiasdiener:
> > Should this go into `mirgecom/__init__.py`?
> 
> IMO it should just be a utility that simulations call that uses MPI collectives internally. In case that's what you meant, I'm not a fant of doing things as a side effect of importing a module. Doing that has come back to bite me every time.

But should this be part of mirgecom, like `mirgecom.select_device()`, or a standalone utility?

inducer:
In principle, I'd be happy to have that in PyOpenCL even.

matthiasdiener:
> In principle, I'd be happy to have that in PyOpenCL even.

Wouldn't PyOpenCL have a dependency on mpi4py then?

MTCam:
Just for our collective reference, this strategy appears to work on Lassen.  According to their documentation, cores and GPUs must be named consistently across nodes or jsrun (and its ilk) will not work as expected.  Here's the dumb thing I'm doing on Lassen:
```
#!/bin/bash --login                                                                                      

[...]

PR #65, opened 2020-08-12
emirgecom

### LINKED ISSUE #52: Add mirgecom CI job to try emirge
Needed because emirge parses `requirements.txt` in this repo. cf. https://github.com/illinois-ceesd/mirgecom/pull/51#issuecomment-670613629

matthiasdiener:
Merge emirge into mirgecom. Leads to imho much simpler (and fewer) scripts, not to mention one less repository. No need for manual parsing of requirements.txt. Main difference: dependent dev packages are now installed into `src/<package>` by default, but we can avoid that if desired.

Fixes https://github.com/illinois-ceesd/emirge/issues/58, fixes https://github.com/illinois-ceesd/emirge/issues/46, fixes #52.

---

Compared to the current situation, this has the following advantages:

- Avoids cross-repo dependency between emirge and mirgecom:
  1. Changes to emirge might break mirgecom installation (since mirgecom is only partially tested on emirge changes).
  2. Changes to mirgecom might break emirge (since emirge isn't tested at all when changing mirgecom).

  These happen quite regularly (ii. about once a week), and it's probably not feasible to test changes comprehensively in both directions. 

- No reimplementation of pip's requirements.txt parsing in shell scripts, ie. can use all requirements.txt features directly, same behavior as pip's documentation.

Potential disadvantages:
- Might discourage people from working on the dependencies (e.g. loopy) directly, since mirgecom appears more like a top-level package.


inducer:
I think this is not a good idea. Unlike software stuff that can be fixed with refactoring, this IMO would do damage inside people's heads by setting incorrect expectations. That tends to be a lot harder to fix.

- Instead of conveying that "the code" consists of "these packages" (including mirgecom), it conveys that mirgecom *is* "the code" (and the other things are "just some dependencies)").
- This reinforces that mirgecom is "the top level package", which I strongly disagree with. IMO, the intention is for it to be a *library* and *reusable*. There could very validly be other packages that use/sit on top of mirgecom. This change creates inertia that ensures this will never happen, i.e. it creates an incentive *against* modularity and *towards* a "big ball of mud".
- To add insult to injury, it seems like the other packages are not even checked out right now. (`pip install --editable git+https://...` does not give me a checkout)
- But suppose that this were fixed (and checkouts of the lower-level packages appeared), hiding them inside `src` *will* make people hesitant to touch them. I don't know if this is clear, but at least from my perspective, the expectation is that developers have a view of and will manipulate the entire stack. This creates incentives *against* that.

The only technical reason that favors this change that I have heard so far revolves around testing. "mirgecom" and "emirge" need to test for mutual compatibility. This is incrementally easier in the single-repo setting, but the disadvantages I list above to me *by far* outweigh this minor benefit.

inducer:
> Fixes illinois-ceesd/emirge#58

It doesn't. It has nothing to do with this issue.

matthiasdiener:
> * To add insult to injury, it seems like the other packages are not even checked out right now. (`pip install --editable git+https://...` does not give me a checkout)

What do you mean? With this PR, loopy etc. are git cloned inside src/ and checked out there.

matthiasdiener:
> > Fixes [illinois-ceesd/emirge#58](https://github.com/illinois-ceesd/emirge/issues/58)
> 
> It doesn't. It has nothing to do with this issue.

From what I understood, it does. Loopy etc. are checked out inside the src/ folder, while f2py (the focus of illinois-ceesd/emirge#58) is not.

matthiasdiener:
> I think this is not a good idea. Unlike software stuff that can be fixed with refactoring, this IMO would do damage inside people's heads by setting incorrect expectations. That tends to be a lot harder to fix.

I'd say that having an extra repo with a bunch of shell scripts to install another repo does more damage to people's heads ;-)

> * Instead of conveying that "the code" consists of "these packages" (including mirgecom), it conveys that mirgecom _is_ "the code" (and the other things are "just some dependencies)").
> * This reinforces that mirgecom is "the top level package", which I strongly disagree with. IMO, the intention is for it to be a _library_ and _reusable_. There could very validly be other packages that use/sit on top of mirgecom. This change creates inertia that ensures this will never happen, i.e. it creates an incentive _against_ modularity and _towards_ a "big ball of mud".

It's not clear to me how this affects mirgecom being a library or reusable. From mirgecom's point-of-view, it is a top-level package (i.e., above loopy etc.). This doesn't mean that there could not be other packages on top of mirgecom.

Other packages that want to build on top of mirgecom just add it to their requirements.txt, and will get mirgecom's requirements when installing the other package, in the same way it is done now. No need to run install.sh or anything else.

> * But suppose that this were fixed (and checkouts of the lower-level packages appeared), hiding them inside `src` _will_ make people hesitant to touch them. I don't know if this is clear, but at least from my perspective, the expectation is that developers have a view of and will manipulate the entire stack. This creates incentives _against_ that.

Like I mentioned above, we can put those packages wherever we like, src/ is just the default. 

> The only technical reason that favors this change that I have heard so far revolves around testing. "mirgecom" and "emirge" need to test for mutual compatibility. This is incrementally easier in the single-repo setting, but the disadvantages I list above to me _by far_ outweigh this minor benefit.

It also avoids the dirty bash gymnastics we have to do in order to parse requirements.txt (which we can never do 100% correctly, even in theory, as the requirements.txt format changes between pip versions). We could add this method to emirge as well, but would have ever fewer reasons to have emirge at all.

majosm:
> * Instead of conveying that "the code" consists of "these packages" (including mirgecom), it conveys that mirgecom _is_ "the code" (and the other things are "just some dependencies)").
> * This reinforces that mirgecom is "the top level package", which I strongly disagree with. IMO, the intention is for it to be a _library_ and _reusable_. There could very validly be other packages that use/sit on top of mirgecom. This change creates inertia that ensures this will never happen, i.e. it creates an incentive _against_ modularity and _towards_ a "big ball of mud".

This is compelling, and I'm leaning towards the emirge approach a little... But at the moment I would say we're already treating mirgecom as the top-level package by using its `requirements.txt` directly in emirge. If we want to treat mirgecom as just another package, shouldn't emirge have its own list of dependencies separate from mirgecom? Say we add some other packages that are needed for simulation drivers but we don't want mirgecom to explicitly depend on them (maybe visualization, post-processing, etc.)?

majosm:
>It's not clear to me how this affects mirgecom being a library or reusable. From mirgecom's point-of-view, it is a top-level package (i.e., above loopy etc.). This doesn't mean that there could not be other packages on top of mirgecom.
>
>Other packages that want to build on top of mirgecom just add it to their requirements.txt, and will get mirgecom's requirements when installing the other package, in the same way it is done now. No need to run install.sh or anything else.

Doesn't the `--editable` stuff kind of break this?

MTCam:
fwiw, and I've said this before in some long slack convos;   I like emirge as a separate package.  I don't use it as a _mirgecom_ installer as much as a general "development environment" installer.    For me, it installs the development environment I require and then it installs mirgecom into it as one of the packages that is more-or-less equal to all the other packages it installs.

That said, I think it could be split into a mirgecom-specific piece that lives inside mirgecom, and an environment install piece that lives outside.  The environment install piece could be responsible for selecting whatever packages you want and git cloning them to the installation dir.

matthiasdiener:
> > It's not clear to me how this affects mirgecom being a library or reusable. From mirgecom's point-of-view, it is a top-level package (i.e., above loopy etc.). This doesn't mean that there could not be other packages on top of mirgecom.
> > Other packages that want to build on top of mirgecom just add it to their requirements.txt, and will get mirgecom's requirements when installing the other package, in the same way it is done now. No need to run install.sh or anything else.
> 
> Doesn't the `--editable` stuff kind of break this?

I'm not sure it would break that, but it should be easy to workaround (by e.g., only adding `--editable` when running from install.sh).

matthiasdiener:
> fwiw, and I've said this before in some long slack convos; I like emirge as a separate package. I don't use it as a _mirgecom_ installer as much as a general "development environment" installer. For me, it installs the development environment I require and then it installs mirgecom into it as one of the packages that is more-or-less equal to all the other packages it installs.
> 
> That said, I think it could be split into a mirgecom-specific piece that lives inside mirgecom, and an environment install piece that lives outside. The environment install piece could be responsible for selecting whatever packages you want and git cloning them to the installation dir.

Isn't the environment part already done (as the extra parameters to install.sh)?

MTCam:
> Isn't the environment part already done (as the extra parameters to install.sh)?

Yes, it is. But I think where our expectations on the functionality and scope of _emirge_ diverges is in the DMZ between where it is treated as a conda+env installer+setter upper vs. mirgecom installer and setter upper.  I could be missing the mark - but factoring in this separation of concerns may help us sort out the pieces.

matthiasdiener:
> If we want to treat mirgecom as just another package, shouldn't emirge have its own list of dependencies separate from mirgecom? Say we add some other packages that are needed for simulation drivers but we don't want mirgecom to explicitly depend on them (maybe visualization, post-processing, etc.)?

I think the dependencies should be declared in the same package as where the simulation drivers are. We are not restricted to using a single requirements.txt, we could easily have a requirements.txt with necessary dependencies and a requirements_sim.txt with optional dependencies for a particular use case.

MTCam:
> This is compelling, and I'm leaning towards the emirge approach a little... But at the moment I would say we're already treating mirgecom as the top-level package by using its `requirements.txt` directly in emirge. If we want to treat mirgecom as just another package, shouldn't emirge have its own list of dependencies separate from mirgecom? Say we add some other packages that are needed for simulation drivers but we don't want mirgecom to explicitly depend on them (maybe visualization, post-processing, etc.)?

@majosm: The need of extra packages above-and-beyond _mirgecom_ requirements.txt is exactly the situation I have and why I added the --pip-pkgs=<requirements.txt> --conda-pkgs=<list>.   I use these to install the stuff that _testy_ needs to run the tests, parse, and plot the test results.  

Nominally, I think that _mirgecom_ could be installed as Matthias says just with `pip install -r mirgecom/requirements.txt`, and the part that does that could be slurped up into _mirgecom_ itself, but from my perspective it would still be useful to have an _emirge_ package that can set up a development environment with the desired packages checked out and installed in develop/editable mode at the desired location.

inducer:
>     * To add insult to injury, it seems like the other packages are not even checked out right now. (`pip install --editable git+https://...` does not give me a checkout)

When @matthiasdiener and I met earlier, we discovered that `pip install --editable git+https://...` seems to find and overwrite existing git checkouts linked to the same environment. That made it look like no checkout was being generated (because it overwrote an unrelated one). I'm not a huge fan of that behavior...

matthiasdiener:
>> Isn't the environment part already done (as the extra parameters to install.sh)?

> Yes, it is. But I think where our expectations on the functionality and scope of emirge diverges is in the DMZ between where it is treated as a conda+env installer+setter upper vs. mirgecom installer and setter upper. I could be missing the mark - but factoring in this separation of concerns may help us sort out the pieces.

Doesn't this separation of concerns exist in this PR? requirements.txt as the mirgecom installer/setter upper, and install.sh as the conda + env installer ?

> but from my perspective it would still be useful to have an _emirge_ package that can set up a development environment with the desired packages checked out and installed in develop/editable mode at the desired location.

Is there a feature/use case that is not supported, or is this related to the separation of concerns?

matthiasdiener:
> > * To add insult to injury, it seems like the other packages are not even checked out right now. (`pip install --editable git+https://...` does not give me a checkout)
> 
> When @matthiasdiener and I met earlier, we discovered that `pip install --editable git+https://...` seems to find and overwrite existing git checkouts linked to the same environment. That made it look like no checkout was being generated (because it overwrote an unrelated one). I'm not a huge fan of that behavior...

Yeah, this is certainly weird. I could not reproduce this:

```bash
$ pwd
/g/g91/diener3/Work

$ pip --version
pip 20.2.2 from /g/g91/diener3/Work/emirge-lassen/miniforge3/envs/dgfem/lib/python3.8/site-packages/pip (python 3.8)

$ python
>>> import pymbolic
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'pymbolic'
>>>

$ pip install --editable git+https://github.com/inducer/pymbolic.git#egg=pymbolic
Obtaining pymbolic from git+https://github.com/inducer/pymbolic.git#egg=pymbolic
  Cloning https://github.com/inducer/pymbolic.git to ./src/pymbolic
[...]

$ python
Python 3.8.5 | packaged by conda-forge | (default, Jul 31 2020, 01:57:54)
>>> import pymbolic
>>> pymbolic.__path__
['/usr/WS1/diener3/Work/src/pymbolic/pymbolic']
>>>

$ cd ..

$  pip install --editable git+https://github.com/inducer/pymbolic.git#egg=pymbolic
Obtaining pymbolic from git+https://github.com/inducer/pymbolic.git#egg=pymbolic
  Cloning https://github.com/inducer/pymbolic.git to ./src/pymbolic
[...]

$ python
>>> import pymbolic
>>> pymbolic.__path__
['/g/g91/diener3/src/pymbolic/pymbolic']
>>>

$ pip uninstall pymbolic
$ python
>>> import pymbolic
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'pymbolic'
```

I.e., it does a fresh checkout and modifies the link to that new checkout.

MTCam:
> Doesn't this separation of concerns exist in this PR? requirements.txt as the mirgecom installer/setter upper, and install.sh as the conda + env installer ?
> Is there a feature/use case that is not supported, or is this related to the separation of concerns?

Yes, the functionality exists in the current _emirge_, and I can't think of a use case that isn't covered.  (Alas, some are more convenient than others) It's just that they are covered under the same umbrella right now (_emirge_), and to me they seem like two different tools - one that installs conda and/or sets up conda env, and another that installs mirgecom into an existing environment (conda or no conda).

majosm:
Some follow-up thoughts from our meeting discussion...

Matthias mentioned two issues with the emirge approach:
1. The need to test emirge against changes in mirgecom
2. The variability of the `requirements.txt` format breaking the parser in emirge

To me it looks like both of these stem from using mirgecom's `requirements.txt` directly. If we instead listed dependencies separately inside emirge, I think that would fix both issues. Plus then we could use our own file format, giving us more flexibility to specify what gets installed with conda vs. pip, etc.

This would also contribute to treating mirgecom as "just another package".

The only real downside I can think of is that it adds some degree of duplication. Seems worth it though?

inducer:
Agree with @majosm. In a way, https://github.com/illinois-ceesd/emirge/issues/58 is already an indication that we were perhaps looking for a file format that's different from `requirements.txt`.

matthiasdiener:
> Some follow-up thoughts from our meeting discussion...
> 
> Matthias mentioned two issues with the emirge approach:
> 
> 1. The need to test emirge against changes in mirgecom
> 2. The variability of the `requirements.txt` format breaking the parser in emirge
> 
> To me it looks like both of these stem from using mirgecom's `requirements.txt` directly. If we instead listed dependencies separately inside emirge, I think that would fix both issues. Plus then we could use our own file format, giving us more flexibility to specify what gets installed with conda vs. pip, etc.
> 
> This would also contribute to treating mirgecom as "just another package".
> 
> The only real downside I can think of is that it adds some degree of duplication. Seems worth it though?

I'm not sure about that.
The CI failures aren't just related to changes in requirements.txt. Duplicating (parts of) the file just gives us more opportunity to get out of sync.

MTCam:
Just want to bounce this off the group.  If emirge does this (and only this):
```
install conda (optional)
conda install <list of packages> (optional)
pip install -r <requirements.txt>
```
And the `<requirements.txt>` is arbitrary (except for having the correct format), then isn't that independent of _mirgecom_?  Since `pip` can already parse a `requirements.txt` and install the packages (optionally downloaded) to a user-specified path, then _emirge_ can easily be used as a stand-alone tool that really just helps a user install conda and a bunch of python packages.  We don't really need _emirge_ to do any parsing of the `requirementst.txt` at all; we can leave that to the proper tool: `pip`.

The only real touch-point with _mirgecom_ is the `requirements.txt` file.   If in _emirge_ CI we run some test installs against _mirgecom_'s `requirements.txt` , and _mirgecom_ CI runs _emirge_ to install _mirgecom_, then doesn't that cover all the bases and prevent the two from missing changes in the other that prevent a working install procedure?

matthiasdiener:
> Just want to bounce this off the group. If emirge does this (and only this):
> 
> ```
> install conda (optional)
> conda install <list of packages> (optional)
> pip install -r <requirements.txt>
> ```

So, essentially not downloading/installing mirgecom itself?

> And the `<requirements.txt>` is arbitrary (except for having the correct format), then isn't that independent of _mirgecom_? 

I think the dependence comes from that emirge is supposed to install an environment that can run mirgecom. Other packages wouldn't usually need this set of scripts since their installation is easier.

> Since `pip` can already parse a `requirements.txt` and install the packages (optionally downloaded) to a user-specified path, then _emirge_ can easily be used as a stand-alone tool that really just helps a user install conda and a bunch of python packages. We don't really need _emirge_ to do any parsing of the `requirementst.txt` at all; we can leave that to the proper tool: `pip`.

Agreed. 

> 
> The only real touch-point with _mirgecom_ is the `requirements.txt` file. If in _emirge_ CI we run some test installs against _mirgecom_'s `requirements.txt` , and _mirgecom_ CI runs _emirge_ to install _mirgecom_, then doesn't that cover all the bases and prevent the two from missing changes in the other that prevent a working install procedure?

Essentially yes, at the price of having to duplicate and synchronize the CI tests in both repos.

majosm:
> The CI failures aren't just related to changes in requirements.txt. Duplicating (parts of) the file just gives us more opportunity to get out of sync.

Can you give some examples of the other CI issues you were running into?

> Just want to bounce this off the group. If emirge does this (and only this):
> 
> ```
> install conda (optional)
> conda install <list of packages> (optional)
> pip install -r <requirements.txt>
> ```
> 
> And the `<requirements.txt>` is arbitrary (except for having the correct format), then isn't that independent of _mirgecom_? Since `pip` can already parse a `requirements.txt` and install the packages (optionally downloaded) to a user-specified path, then _emirge_ can easily be used as a stand-alone tool that really just helps a user install conda and a bunch of python packages. We don't really need _emirge_ to do any parsing of the `requirementst.txt` at all; we can leave that to the proper tool: `pip`.
> 
> The only real touch-point with _mirgecom_ is the `requirements.txt` file. If in _emirge_ CI we run some test installs against _mirgecom_'s `requirements.txt` , and _mirgecom_ CI runs _emirge_ to install _mirgecom_, then doesn't that cover all the bases and prevent the two from missing changes in the other that prevent a working install procedure?

If I understand the situation with `pip` correctly, the trouble is that we don't have a good way to tell it whether a given package should produce a checkout in the install directory or not (I think that's what #58 is about? And the `--editable` stuff above).

Having a parsing capability on hand should also make it easier for us to add some convenience features in the future ([*cough*](https://github.com/illinois-ceesd/emirge/issues/53)).

MTCam:
> If I understand the situation with `pip` correctly, the trouble is that we don't have a good way to tell it whether a given package should produce a checkout in the install directory or not (I think that's what #58 is about? And the `--editable` stuff above).

True, but `pip` can already perform the task already of doing the download/clone/checkout to a user-specified path if the `--editable` flag is used. So if _emirge_ only did the basic things:
```
install conda (optional)
conda --yes install <package list>
pip install -r <requirements.txt>
```
Then, getting it to install a _mirgecom_-compatible environment would consist only of passing it the right `<package list>` (if required), and `<requirements.txt>` file.  The `<requirements.txt>` file only needs to include the packages for which we need the `--editable` flag.  Everything else would already be automatically handled by `pip` since it will install *any* package and all the ones listed in each package-specific `requirementst.txt` automatically.

majosm:
> True, but `pip` can already perform the task already of doing the download/clone/checkout to a user-specified path if the `--editable` flag is used. So if _emirge_ only did the basic things:

Can it? From the discussion above it looked like `--editable` doesn't work (or at least has some pretty undesirable behavior).
 
> ```
> install conda (optional)
> conda --yes install <package list>
> pip install -r <requirements.txt>
> ```
> 
> Then, getting it to install a _mirgecom_-compatible environment would consist only of passing it the right `<package list>` (if required), and `<requirements.txt>` file. The `<requirements.txt>` file only needs to include the packages for which we need the `--editable` flag. Everything else would already be automatically handled by `pip` since it will install _any_ package and all the ones listed in each package-specific `requirementst.txt` automatically.

Not sure I'm 100% following what you're describing... but I think we want to avoid putting `--editable` (or equivalent) inside the `requirements.txt` of any subpackages (mirgecom included) as that will impact their installation when used outside of emirge.

MTCam:
> Not sure I'm 100% following what you're describing... but I think we want to avoid putting `--editable` (or equivalent) inside the `requirements.txt` of any subpackages (mirgecom included) as that will impact their installation when used outside of emirge.

Exactly.  I'm suggesting that _emirge_ could use a run-of-the-mill `requirements.txt` that specifies only the packages the present user wants to install as `--editable`.  All other packages are automatically picked up by `pip` when it parses the package-specific `requirements.txt`.

majosm:
> Exactly. I'm suggesting that _emirge_ could use a run-of-the-mill `requirements.txt` that specifies only the packages the present user wants to install as `--editable`. All other packages are automatically picked up by `pip` when it parses the package-specific `requirements.txt`.

I thiiiink I'm on board with that... 🙂 If we can get `--editable` to work, at least.

matthiasdiener:
> > Not sure I'm 100% following what you're describing... but I think we want to avoid putting `--editable` (or equivalent) inside the `requirements.txt` of any subpackages (mirgecom included) as that will impact their installation when used outside of emirge.
> 
> Exactly. I'm suggesting that _emirge_ could use a run-of-the-mill `requirements.txt` that specifies only the packages the present user wants to install as `--editable`. All other packages are automatically picked up by `pip` when it parses the package-specific `requirements.txt`.

Not all packages in our mirgecom/requirements.txt are currently installed as dependencies of the other packages though, e.g. pytest.

majosm:
> Not all packages in our mirgecom/requirements.txt are currently installed as dependencies of the other packages though, e.g. pytest.

I think in this scenario mirgecom itself would be one of the subpackages inside `emirge/requirements.txt`, so pytest, etc. should get installed.

matthiasdiener:
Ok, let me try to implement that.

MTCam:
> Ok, let me try to implement that.

In this particular instance, I think to make _emirge_ install a working version of _mirgecom_, then the `requirements.txt` file passed to _emirge_ would only need to include an entry for _mirgecom_.   If you want any development/downloaded/cloned packages, then listing them in the _emirge_ `requirements.txt` with `--editable` flag should do the trick.

majosm:
> In this particular instance, I think to make _emirge_ install a working version of _mirgecom_, then the `requirements.txt` file passed to _emirge_ would only need to include an entry for _mirgecom_. If you want any development/downloaded/cloned packages, then listing them in the _emirge_ `requirements.txt` with `--editable` flag should do the trick.

In principle... but I think we should default to editable installs of all of the frequently-modified packages (in keeping with @inducer's proposed "mirgecom is not *the* code" philosophy).

MTCam:
> In principle... but I think we should default to editable installs of all of the frequently-modified packages (in keeping with @inducer's proposed "mirgecom is not _the_ code" philosophy).

For our own purposes, sure.  This would amount to not much more than a `user.requirements.txt` and  `developer.requirements.txt`.   The user could choose which one he needs, and *could* manually tweak his own `requirements.txt` if he wants to switch out any particular package or branch.

A collection of `requirements.txt` files living in _emirge_ or some central place could provide examples of installing _mirgecom_ in various ways, because all you'd really need to replicate or create an environment is the proper `requirements.txt` file.

MTCam:
Harsh yo.

Issue #66, opened 2020-08-13
Avoid doing so many device array allocations


inducer:
According to @matthiasdiener's measurement, the code spends a lot of time allocating/deallocating. By creating an array context subclass that uses a [memory pool](https://documen.tician.de/pyopencl/tools.html#pyopencl.tools.MemoryPool), we might be able to avoid them.

matthiasdiener:
Patch has been merged.

PR #68, opened 2020-08-13
use pooled allocator

matthiasdiener:
Reduces memory allocations from 300,000 to 160.
15% speedup.
Addresses #66.

### DIFF
--- a/examples/wave-eager.py
+++ b/examples/wave-eager.py
@@ -31,6 +31,7 @@ from mirgecom.wave import wave_operator
 from mirgecom.integrators import rk4_step
 from meshmode.array_context import PyOpenCLArrayContext
 from meshmode.dof_array import thaw
+import pyopencl.tools as cl_tools
 
 
 def bump(actx, discr, t=0):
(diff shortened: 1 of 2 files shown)


examples/wave-eager.py
```diff
@@ -54,7 +55,8 @@ def bump(actx, discr, t=0):
 def main():
     cl_ctx = cl.create_some_context()
     queue = cl.CommandQueue(cl_ctx)
-    actx = PyOpenCLArrayContext(queue)
+    actx = PyOpenCLArrayContext(queue,
+        allocator=cl_tools.MemoryPool(cl_tools.ImmediateAllocator(queue)))
```
inducer, line 59: It'll be pretty easy to forget that, and as a result get rotten performance. Could you submit a patch to meshmode that produces a warning if a `PyOpenCLArrayContext` is created without an allocator? (where the warning can be silenced by passing `allocator=False`)

matthiasdiener: Done here: https://github.com/inducer/meshmode/pull/34

matthiasdiener:
@inducer Should I do the same for wave-eager-mpi?

inducer:
Yeah

inducer:
Added a bit to #7 to make sure the Euler examples start life with pools enabled as well.

cc @MTCam

PR #75, opened 2020-08-23
Multicomponent infrastructure

MTCam:
Sticking this here for now. TODO: refactor this so that it does not repeat any of the Euler module code.


MTCam:
Superseded by #95 where this is reported to current main.

PR #83, opened 2020-09-01
Refactor some common utilities into simutils

MTCam:
* Adds parallel grid generator common to all examples
* Generalizes checkpoint to be used even for sims without exact soln (needed for Y0/Y1 )
* Adds a placeholder for profiling util (also used in Y0/Y1 example)

Note: Y0/Y1 Euler example will follow in a separate PR

### DIFF
--- a/examples/lump-mpi.py
+++ b/examples/lump-mpi.py
@@ -89,9 +89,9 @@ def main(ctx_factory=cl.create_some_context):
     rank = comm.Get_rank()
 
     from meshmode.mesh.generation import generate_regular_rect_mesh
-    grid_generator = partial(generate_regular_rect_mesh, a=(box_ll,) * dim,
-                             b=(box_ur,) * dim, n=(nel_1d,) * dim)
-    local_mesh, global_nelements = create_parallel_grid(comm, grid_generator)
+    generate_grid = partial(generate_regular_rect_mesh, a=(box_ll,) * dim,
+                            b=(box_ur,) * dim, n=(nel_1d,) * dim)
+    local_mesh, global_nelements = create_parallel_grid(comm, generate_grid)
     local_nelements = local_mesh.nelements
 
     discr = EagerDGDiscretization(
--- a/examples/lump-mpi.py
+++ b/examples/lump-mpi.py
@@ -39,7 +39,7 @@ from mirgecom.euler import inviscid_operator
 from mirgecom.simutil import (
     inviscid_sim_timestep,
     sim_checkpoint,
-    create_parallel_box_grid,
+    create_parallel_grid,
     ExactSolutionMismatch
 )
 from mirgecom.io import make_init_message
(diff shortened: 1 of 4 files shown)


examples/lump-mpi.py
```diff
@@ -85,34 +86,10 @@ def main(ctx_factory=cl.create_some_context):
...
-
-        part_per_element = get_partition_by_pymetis(mesh, num_parts)
-
-        local_mesh = mesh_dist.send_mesh_parts(mesh, part_per_element, num_parts)
-        del mesh
-
-    else:
-        local_mesh = mesh_dist.receive_mesh_part()
+    local_mesh, global_nelements = create_parallel_box_grid(comm, dim, box_ll,
+                                                            box_ur, nel_1d)
```
inducer, line 92: I think I'd prefer something a bit more general that gets the mesh by a callback. (rather than being specific to boxes)

MTCam: Yeah, ok, try this on.   I *think* this is something like what you are getting at.

mirgecom/simutil.py
```diff
@@ -104,46 +104,94 @@ def exact_sim_checkpoint(discr, exact_soln, visualizer, eos, q,
...
         return 0
 
-    actx = q[0].array_context
-    nodes = thaw(actx, discr.nodes())
-    rank = 0
-
-    if comm is not None:
-        rank = comm.Get_rank()
+    if profiler is not None:
+        profiler.StartTimer(f"checkpoint_{step}")
```
inducer, line 108: I'd be more comfortable leaving these out. For one, this seems to preempt https://github.com/illinois-ceesd/planning/issues/57, and for another, the non-PEP8 interface sticks out to me like a sore thumb.

mirgecom/simutil.py
```diff
@@ -104,46 +104,94 @@ def exact_sim_checkpoint(discr, exact_soln, visualizer, eos, q,
...
-            ("dv", dependent_vars),
-            ("exact_soln", expected_state),
-            ("residual", q - expected_state)
+            ("cv", cv),
+            ("dv", dependent_vars)
         ]
+        if exact_soln is not None:
+            exact_list = [
+                ("exact_soln", expected_state),
+                ("residual", exp_resid)
```
inducer, line 141: I'm kind of skeptical about adding this to the dump. At least paraview can straightforwardly calculate this internally.

mirgecom/simutil.py
```diff
@@ -104,46 +104,94 @@ def exact_sim_checkpoint(discr, exact_soln, visualizer, eos, q,
...
-        if maxerr > exittol:
-            raise ExactSolutionMismatch(step, t=t, state=q)
+    if profiler is not None:
+        profiler.EndTimer(f"checkpoint_{step}")
+
+    if maxerr > exittol:
+        raise ExactSolutionMismatch(step, t=t, state=q)
+
+
+def create_parallel_grid(comm, grid_generator):
```
inducer, line 174: Documentation?

inducer, line 174: Callables should be named with verbs.

mirgecom/simutil.py
MTCam: This is removed now, however...

I don't think it duplicates https://github.com/illinois-ceesd/planning/issues/57.   That Issue does not mention timing user-defined constructs - which I need.   Timing kernels only does not give the information that I need to profile the simulation.   For example, how do I use anything from https://github.com/illinois-ceesd/planning/issues/57 to time the I/O?  What if I want to time how long I spend inside the checkpoint routine (which I do)?

mirgecom/simutil.py
MTCam: It was only convenience.  It is gone.

mirgecom/simutil.py
MTCam: Added - but possibly all messed up, heh.  We'll see.

MTCam: Fixed.

MTCam: I actually think it is OK, except I didn't figure out how to document the MPI communicator in a great way.

inducer: See 83bb205.

mirgecom/simutil.py
inducer: (see the discussion there)

inducer (approved):

PR #87, opened 2020-09-03
Install pymetis from conda

inducer:


requirements.txt
```diff
@@ -2,7 +2,6 @@ mpi4py
 numpy
 pytest
 pyvisfile
-pymetis
```
matthiasdiener, line 5: Shouldn't we keep this here still, if someone installs mirgecom without using our conda scripts?

inducer: Fair, added back.

matthiasdiener (approved):

PR #88, opened 2020-09-03
Add pydocstyle linter

inducer:
This finds things like
```
        D417: Missing argument descriptions in the docstring (argument(s) discr are missing descriptions in 'inviscid_operator' docstring)
         D401: First line should be in imperative mood; try rephrasing (found 'The')
         D100: Missing docstring in public module
```
as well as heaps and heaps of formatting stuff.

@lukeolson @majosm @matthiasdiener @MTCam Thoughts?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,8 +21,31 @@ jobs:
             python3 -m venv myenv
             source myenv/bin/activate
             python -m pip install wheel
-            python -m pip install flake8 pep8-naming
-            python -m flake8 --show-source --statistics "$(basename $GITHUB_REPOSITORY)" test examples && echo "Flake8 found no errors."
+            python -m pip install flake8 pep8-naming flake8-quotes
+            python -m flake8 --show-source --statistics "$(basename $GITHUB_REPOSITORY)" test examples setup.py doc/conf.py && echo "Flake8 found no errors."
+
+    pylint:
+        name: Pylint
+        runs-on: ubuntu-latest
+        steps:
+        - uses: actions/checkout@v2
+        -
+          uses: actions/setup-python@v1
+          with:
+            python-version: '3.x'
+        - name: Install
+          shell: bash -l {0}
+          run: |
+            . .ci-support/install.sh
+
+        - name: Pylint
+          run: |
+            MINIFORGE_INSTALL_DIR=.miniforge3
+            . "$MINIFORGE_INSTALL_DIR/bin/activate" testing
+            python -m pip install wheel
+            python -m pip install pylint pyyaml
+            curl -L -O https://raw.githubusercontent.com/inducer/ci-support/master/run-pylint.py
+            python run-pylint.py --yaml-rcfile=.pylintrc.yml "$(basename $GITHUB_REPOSITORY)" test examples && echo "Pylint found no errors."
 
     pydocstyle:
         runs-on: ubuntu-latest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,6 +24,22 @@ jobs:
             python -m pip install flake8 pep8-naming
             python -m flake8 --show-source --statistics "$(basename $GITHUB_REPOSITORY)" test examples && echo "Flake8 found no errors."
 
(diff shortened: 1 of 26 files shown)


MTCam:
I like this, in principal.  flake8/editor integration changed my life. I have a similar hope for this tool.

Do we intend on fixing the gobs of stuff it finds?  If we can make it shut up about the stuff we don't care about and fix all the stuff we do, then it could be very useful.    On the other hand, if whatever new violations I produce with my new code are buried in a mountain of other stuff and CI passes whether they are fixed or not - then I would tend to care less about what it tells me.

If there is a mountain of busy work associated with the current report - then I propose that we work on bringing master into the quiet zone of this tool before adding it to CI.

inducer:
> If there is a mountain of busy work associated with the current report - then I propose that we work on bringing master into the quiet zone of this tool before adding it to CI.

I agree. The main question I'm meaning to raise here is whether the signal/noise ratio of this tool looks sufficiently high to justify that effort. I'm leaning slightly yes, which is why I'm proposing it.

MTCam:
> > If there is a mountain of busy work associated with the current report - then I propose that we work on bringing master into the quiet zone of this tool before adding it to CI.
> 
> I agree. The main question I'm meaning to raise here is whether the signal/noise ratio of this tool looks sufficiently high to justify that effort. I'm leaning slightly yes, which is why I'm proposing it.

Looks like master could be quieted pretty easily. Regardless, this seems pretty useful to me.

inducer:
There's https://pypi.org/project/flake8-docstrings/ which integrates this with Flake8.

MTCam:
> There's https://pypi.org/project/flake8-docstrings/ which integrates this with Flake8.

That seems mean, but I'm game :)

inducer:
Decision: crowdsource, shepherded by @MTCam

inducer:
FYI Resolved some merge conflicts

MTCam:
We need to do something with this to make it accept our files.   Currently, it reports we don't have any module-level docstrings because it appears to hate __doc__ for constructing the docstring - as well as the fact that __copyright__ comes before __doc__ (i.e. assuming the __doc__ issue can be fixed).

Any ideas @inducer?

inducer:
My take is that `__doc__` goes away and turns into an undecorated string at the start of the module. Fine by me, just a different convention.

MTCam:
Silencing some of these complaints required my adding some pretty obvious single line documentation.

inducer:
Thanks for working on this! LGTM, though I can't approve because I initially created this PR.

MTCam:
Thanks for turning this on!  Adding this linter to flake8 has helped (me) with the docs.  I'm sure you noticed some of the docs weren't especially useful or informative, as we have discussed, but it seems to be worth it.

.github/workflows/ci.yaml
```diff
@@ -47,6 +47,22 @@ jobs:
...
+        -
+          uses: actions/setup-python@v1
+          with:
+            python-version: '3.x'
+        - name: Lint
+          run: |
+            python3 -m venv myenv
+            source myenv/bin/activate
+            python -m pip install wheel
+            python -m pip install pydocstyle
```
lukeolson, line 63: we could consider removing this ci section and absorb it above with `flake8-docstrings` (and possibly other flake8 extensions.

mirgecom/boundary.py
```diff
@@ -1,3 +1,12 @@
+"""Boundary condition implementations.
+
+Boundary Conditions
+^^^^^^^^^^^^^^^^^^^
```
lukeolson, line 4: Why do we use an H5 header here? (there may be a good reason)

mirgecom/boundary.py
```diff
@@ -28,28 +37,28 @@
...
     .. automethod:: __init__
     .. automethod:: boundary_pair
     """
+
     def __init__(self, userfunc):
+        """Set the boundary function.
+
+        Parameters
+        ----------
+        userfunc
```
lukeolson, line 52: no type, but we're not checking for this yet.  skip.

lukeolson (approved):

PR #91, opened 2020-09-07
Silence flake8 for conf files

matthiasdiener:


inducer:
Shouldn't we also make sure flake8 runs on those conf files during CI?

matthiasdiener:
> Shouldn't we also make sure flake8 runs on those conf files during CI?

Good point, done.

.github/workflows/ci.yaml
```diff
@@ -18,8 +18,8 @@ jobs:
             python-version: '3.x'
         - name: Flake8 test
           run: |
-            python3 -m venv myenv
-            source myenv/bin/activate
+            python3 -m venv ../myenv
+            source ../myenv/bin/activate
```
inducer, line 22: I disagree.

- We don't know if we're allowed to write there.
- We should just provide flake8 a list of places to test, rather than be grumpy if it picks up all sorts of stuff.

.github/workflows/ci.yaml
```diff
@@ -18,8 +18,8 @@ jobs:
...
             python -m pip install wheel
             python -m pip install flake8 pep8-naming flake8-quotes
             python -m flake8 --show-source --statistics && echo "Flake8 found no errors."
```
inducer, line 25:
```suggestion
            python -m flake8 --show-source --statistics "$(basename $GITHUB_REPOSITORY)" test examples setup.py doc/conf.py && echo "Flake8 found no errors."
```

.github/workflows/ci.yaml
matthiasdiener: Fair enough. I accepted your suggestion, however:
- This is only run for CI, and we know we are allowed to write there.
- I'm not sure how people usually run flake8. I just run `flake8` in `mirgecom/` which is how these errors were shown. If someone adds a new top-level file or directory, we might see this kind of error again.

inducer: >  If someone adds a new top-level file or directory 

That is true. And I don't have a good way around it. The problem is that various bits of the ecosystem (*cough* setuptools *cough*) like to dump random crap that may or may not include python source files in your package directory. So the blacklist approach is equally unsustainable.

inducer (approved):

PR #27, opened 2020-07-15
Add spectral filter to Euler

MTCam:
Just had to make a PR on this.  I think I've spent a bit too much time trying to figure out the error that comes up from the test:
```
( ... )
>       filtered_soln = spectral_filter(vol_discr, uniform_soln)

test_filter.py:118: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../mirgecom/filter.py:87: in __call__
    self._knl()(queue=queue, mat=filter_operator,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = LoopKernel(state=0, temporary_variables={}, symbol_manglers=[], target=<loopy.target.pyopencl.PyOpenCLTarget object at...))))))], overridden_get_grid_sizes_for_insn_ids=None, name='diff', index_dtype=np:dtype('int32'), silenced_warnings=[])
args = (), kwargs = {}

    def __call__(self, *args, **kwargs):
>       key = self.target.get_kernel_executor_cache_key(*args, **kwargs)
E       TypeError: get_kernel_executor_cache_key() missing 1 required positional argument: 'queue'

../../loopy/loopy/kernel/__init__.py:1442: TypeError
```

PR #92, opened 2020-09-08
Add filter funcs, tests

MTCam:
This PR will take over from the old one (#27).  Filter funcs ported to main.

### DIFF
--- a/mirgecom/filter.py
+++ b/mirgecom/filter.py
@@ -161,8 +161,10 @@ def apply_linear_operator(discr, operator, fields):
     return result
 
 
-def create_group_filter_operator(group, filter_mat):
+def create_group_filter_operator(group, cutoff, response_func):
     """Create spectral filter operator for *group*."""
+    filter_mat = make_spectral_filter(group, cutoff,
+                                      mode_response_function=response_func)
     from modepy import vandermonde
     vander = vandermonde(group.basis(), group.unit_nodes)
     vanderm1 = np.linalg.inv(vander)
@@ -171,7 +173,7 @@ def create_group_filter_operator(group, filter_mat):
 
 
 @obj_array_vectorize
-def filter_modally(discrwb, dd, field):
+def filter_modally(discrwb, dd, field, cutoff, mode_resp_func):
     """Stand-alone procedural interface to spectral filtering."""
     dd = sym.as_dofdesc(dd)
     discr = discrwb.discr_from_dd(dd)
@@ -184,7 +186,7 @@ def filter_modally(discrwb, dd, field):
 
     @memoize_in(discrwb, (filter_modally, "get_matrix"))
     def get_matrix(group):
-        return create_group_filter_operator(group)
+        return create_group_filter_operator(group, cutoff, mode_resp_func)
 
     actx = field.array_context
     result = discr.empty(actx, dtype=field.entry_dtype)
--- a/mirgecom/filter.py
+++ b/mirgecom/filter.py
@@ -6,7 +6,6 @@ JSH/TW Nodal DG Methods (DOI: 10.1007/978-0-387-72067-8), Section 5.3
 .. automethod: exponential_mode_response_function
 .. automethod: make_spectral_filter
 .. automethod: linear_operator_kernel
-.. automethod: apply_linear_operator
 .. automethod: create_group_filter_operator
 .. automethod: filter_modally
 """
@@ -40,10 +39,7 @@ import loopy as lp
 from grudge import sym
 from meshmode.dof_array import DOFArray
 from pytools import memoize_in
-from pytools.obj_array import (
-    obj_array_vectorized_n_args,
-    #    obj_array_vectorize
-)
+from pytools.obj_array import obj_array_vectorized_n_args
 
 
 def exponential_mode_response_function(mode, alpha, cutoff, nfilt, filter_order):
@@ -52,7 +48,7 @@ def exponential_mode_response_function(mode, alpha, cutoff, nfilt, filter_order)
                   ** (2*filter_order))
 
 
-def make_spectral_filter(element_group, cutoff, mode_response_function):
+def make_spectral_filter(group, cutoff, mode_response_function):
     r"""
     Create a spectral filter with the provided *mode_response_function*.
 
@@ -76,9 +72,9 @@ def make_spectral_filter(element_group, cutoff, mode_response_function):
     filter: np.ndarray
         filter operator in the modal basis
     """
-    mode_ids = element_group.mode_ids()
-    order = element_group.order
-    dim = element_group.dim
+    mode_ids = group.mode_ids()
+    order = group.order
+    dim = group.dim
 
     nmodes = len(mode_ids)
     filter = np.identity(nmodes)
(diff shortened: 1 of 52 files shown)


mirgecom/filter.py
```diff
@@ -0,0 +1,194 @@
...
+    nmodes = len(mode_ids)
+    filter = np.identity(nmodes)
+    nfilt = order - cutoff
+    if nfilt <= 0:
+        return filter
+
+    for mode_index, mode_id in enumerate(mode_ids):
+        mode = mode_id
+        if dim > 1:
+            mode = sum(mode_id)
```
inducer, line 92: Why not pass on the mode ID unchanged?

mirgecom/filter.py
```diff
@@ -0,0 +1,194 @@
...
+        0<=idof<ndiscr_nodes_out and
+        0<=j<ndiscr_nodes_in}""",
+        "result[iel,idof] = sum(j, mat[idof, j] * vec[iel, j])",
+        name="spectral_filter")
+    knl = lp.tag_array_axes(knl, "mat", "stride:auto,stride:auto")
+    return knl
+
+
+@obj_array_vectorized_n_args
+def apply_linear_operator(discr, operator, fields):
```
inducer, line 115: A (meshmode) discretization can consist of multiple element groups, which may have different element types. The same matrix is unlikely to cut it for all of them, since they're not even guaranteed to have the same DOF count.

Since it's conceptually wrong, I kind of don't want it in the public API. I see that it's only used in the tests. I'd be happier with it if it were moved to the tests and asserted that `discr` only has one group.

test/test_euler.py
```diff
@@ -750,6 +773,9 @@ def rhs(t, q):
                 write_soln()
 
         fields = rk4_step(fields, t, dt, rhs)
+        fields = filter_modally(discr, "vol", cutoff=cutoff,
+                                       mode_resp_func=frfunc, field=fields)
```
inducer, line 777: It would be interesting to test this someplace that otherwise blows up due to aliasing.

test/test_euler.py
```diff
@@ -738,6 +746,21 @@ def write_soln(write_status=True):
     def rhs(t, q):
         return inviscid_operator(discr, eos=eos, boundaries=boundaries, q=q, t=t)
 
+    filter_order = 8
+    eta = .5
+    alpha = -1.0*np.log(np.finfo(float).eps)
+    nummodes = int(1)
+    for i in range(dim):
+        nummodes *= int(order + dim + 1)
+    nummodes /= math.factorial(int(dim))
```
inducer, line 755: Reference? Would it be helpful to stick this in a utility function?

test/test_euler.py
inducer, line 777: The filter should drop the convergence order. That might be useful to test.

mirgecom/filter.py
```diff
@@ -0,0 +1,194 @@
...
+        actx.call_loopy(
+            linear_operator_kernel(),
+            mat=actx.from_numpy(operator),
+            result=result[group.index],
+            vec=fields[group.index])
+    return result
+
+
+def create_group_filter_operator(group, cutoff, response_func):
+    """Create spectral filter operator for *group*."""
```
inducer, line 129: Document type of *group*.

mirgecom/filter.py
```diff
@@ -0,0 +1,194 @@
...
+    for group in discr.groups:
+        actx.call_loopy(
+            linear_operator_kernel(),
+            mat=actx.from_numpy(operator),
+            result=result[group.index],
+            vec=fields[group.index])
+    return result
+
+
+def create_group_filter_operator(group, cutoff, response_func):
```
inducer, line 128: Argument name consistency: `group` here, `element_group` elsewhere.

mirgecom/filter.py
```diff
@@ -0,0 +1,194 @@
...
+        - *mode_resp_func* function returns a filter coefficient
+        for a given mode
+        - memoized into the discretization
+    * a kernel to apply the operator
+        - memoized into the array context
+    * a filtered solution wherein the filter is applied to *field*.
+
+    Parameters
+    ----------
+    discrwb: :class:`grudge.discrwb`
```
inducer, line 156: I'm surprised this passed the reference checker---that identifier doesn't exist. Do you have the latest master merged?

mirgecom/filter.py
```diff
@@ -0,0 +1,194 @@
...
+        Grudge discretization with boundaries object
+    dd:
+        Discretization restriction
```
inducer, line 159:
```suggestion
    dd: :class:`grudge.sym.DOFDesc` or similar, as accepted by :func:`grudge.sym.as_dof_desc`
        Describe the type of DOF vector on which to operate.
```

mirgecom/filter.py
```diff
@@ -0,0 +1,194 @@
...
+        Mode below which *field* will not be filtered
+    mode_resp_func:
+        Modal response function returns a filter coefficient for input mode id
+    field: numpy.ndarray
+        DOFArray or object array of DOFArrays
+
+    Returns
+    -------
+    result: numpy.ndarray
+        Filtered *field* like *field*.
```
inducer, line 170:
```suggestion
        Filtered version of *field*.
```

test/test_filter.py
```diff
@@ -0,0 +1,267 @@
...
+from pytools.obj_array import (
+    make_obj_array
+)
+from meshmode.dof_array import thaw  # noqa
+from mirgecom.filter import (
+    make_spectral_filter,
+    apply_linear_operator,
+)
+# Uncomment if you want to inspect results in VTK
+# from grudge.shortcuts import make_visualizer
```
inducer, line 49: Just add a flag for this. (maybe a default argument to the test? `visualize=False` maybe?)

test/test_filter.py
```diff
@@ -0,0 +1,267 @@
...
+import pytest
+import numpy as np
+import pyopencl as cl
+from functools import partial
+
+from meshmode.array_context import PyOpenCLArrayContext
+from meshmode.dof_array import thaw
+from grudge.eager import EagerDGDiscretization
+from pyopencl.tools import (  # noqa
+    pytest_generate_tests_for_pyopencl as pytest_generate_tests,
```
inducer, line 38: There's a better version of this that gets you the array context directly, see `test_euler` on master.

test/test_filter.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+@pytest.mark.parametrize("dim", [1, 2, 3])
+@pytest.mark.parametrize("order", [2, 3, 4])
+@pytest.mark.parametrize("filter_order", [1, 2, 3])
+def test_filter_coeff(ctx_factory, filter_order, order, dim):
+    """
+    Test the construction of filter coefficients.
+
+    Tests that the filter coefficients have the right shape
```
inducer, line 59: > right shape

Not sure what that means

inducer (commented):
I enjoyed reviewing this. Nicely done! I noted just a couple wrinkles.

test/test_euler.py
MTCam: Yes!

mirgecom/filter.py
MTCam: I had an error when doing 1D.  I think when calling `sum()` on an integer.

test/test_euler.py
MTCam: Utility?  Nah I don't think, but I defer to your judgement.

mirgecom/filter.py
MTCam: Reminding myself now... the way I had it before, I would get an error when trying to run the filtering on 1D discretizations.  The issue was that `sum()` was called on a non-iterable. I guess that `group.mode_ids()` returns an `integer` for 1D, and an iterable  object only on numdim > 1.

mirgecom/filter.py
MTCam: Yeah, it's merged.  I was sloppy here because it auto-filled it.  Is it better now?

MTCam:
> I enjoyed reviewing this. Nicely done! I noted just a couple wrinkles.

Thanks :)

test/test_euler.py
MTCam: Running order-2 elements on the Vortex2D case, the filter appears to improve the order of convergence.

thomasgibson:
Revived in https://github.com/illinois-ceesd/mirgecom/pull/324

PR #93, opened 2020-09-09
Add y0 driver

MTCam:
This adds the driver than runs an inviscid ~sonic flow through the scaled pseudo-Y0 geometry.   Let's discuss how to add it?

### DIFF
--- a/examples/hello-world-mpi.py
+++ b/examples/hello-world-mpi.py
@@ -1,3 +1,5 @@
+"""Demonstrate simple MPI use."""
+
 from mpi4py import MPI
 
 comm = MPI.COMM_WORLD
--- a/examples/lump-mpi.py
+++ b/examples/lump-mpi.py
@@ -1,3 +1,5 @@
+"""Demonstrate simple mass lump advection."""
+
 __copyright__ = """
 Copyright (C) 2020 University of Illinois Board of Trustees
 """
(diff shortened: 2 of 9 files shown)


majosm:
I guess I would lean towards leaving these kinds of one-off simulations out of master. I do think we should keep them around somewhere, though... maybe a set of archive branches in the main repo, or a separate archive fork?

MTCam:
I lean towards keeping the production (and other experiments) out as well.   Besides examples, I would be happy if we kept separate the constructs that "add to _mirgecom_" vs. the constructs that "use _mirgecom_".   That is, as long as it is possible to reproduce the code and experiment used to generate results - even after _mirgecom_ itself has evolved since the results were produced.

inducer:
> That is, as long as it is possible to reproduce the code and experiment used to generate results - even after mirgecom itself has evolved since the results were produced.

I agree, and in that sense, a separate repo (with runnability guarded by periodic CI, though likely not full-scale runnability) might be a reasonable spot for these. The README of those could include the output of https://github.com/illinois-ceesd/emirge/blob/master/version.sh (if that included precise git revisions, which it doesn't seem to? @matthiasdiener).

MTCam:
> I agree, and in that sense, a separate repo (with runnability guarded by periodic CI, though likely not full-scale runnability) might be a reasonable spot for these. The README of those could include the output of https://github.com/illinois-ceesd/emirge/blob/master/version.sh (if that included precise git revisions, which it doesn't seem to? @matthiasdiener).

How about `illinois-ceesd/experiments`?

inducer:
> How about `illinois-ceesd/experiments`?

Sure.

MTCam:
Superseded by [experiments PR 2](https://github.com/illinois-ceesd/experiments/pull/2)

PR #95, opened 2020-09-10
Add support for scalar advection

MTCam:
This PR is intended to add:

- [X] Multispecies/mixture components to conserved quantities
- [X] Scalar component advection to flux routines
- [x] Tests for scalar advection
- [x] Example / integrated test

(Also fixes some documentation issues from main)


MTCam:
Example runs are fixed by [this meshmode PR](https://github.com/inducer/meshmode/pull/51).

mirgecom/initializers.py
```diff
@@ -462,6 +490,228 @@ def exact_rhs(self, discr, q, t=0.0):
         return flat_obj_array(massrhs, energyrhs, momrhs)
 
 
+class MultiLump:
+    r"""Solution initializer for multi-species N-dimensional Gaussian lump of mass.
```
inducer, line 494: `Lump` itself also allows multiple species. How is this different? Why do we need both? The docs should say.

mirgecom/initializers.py
```diff
@@ -339,8 +340,11 @@ class Lump:
     """
 
     def __init__(
```
inducer, line 342: This might be a good use-case for keyword-only args.

mirgecom/initializers.py
```diff
@@ -588,9 +856,24 @@ def __call__(self, t, x_vec, eos=IdealSingleGas()):
...
+        from mirgecom.euler import join_conserved
+        return join_conserved(dim=self._dim, mass=mass, energy=energy,
+                              momentum=mom, massfractions=massfrac)
+    # Save this - some things in Uniform flow generator are broken
+    # and one or the other of the above or below fixes the issue.
+    # =======
+    #         return _make_uniform_flow(x_vec=x_vec, mass=self._rho,
+    #                                  pressure=self._p, energy=self._e,
+    #                                  velocity=self._velocity, eos=eos)
+    # >>>>>>> master
```
inducer, line 876: Explain. What's broken? Why can't we just fix it?

test/test_euler.py
```diff
@@ -57,8 +57,9 @@
 logger = logging.getLogger(__name__)
 
 
+@pytest.mark.parametrize("nspecies", [1, 2, 3])
 @pytest.mark.parametrize("dim", [1, 2, 3])
```
inducer, line 61: Do we really need to test all nine combinations?

test/test_euler.py
```diff
@@ -97,23 +98,42 @@ def test_inviscid_flux(actx_factory, dim):
...
 
     # {{{ create the expected result
 
     p = eos.pressure(cv)
     escale = (energy + p) / mass
 
-    expected_flux = np.zeros((dim + 2, dim), dtype=object)
+    numeq = dim + 2
+    if nspecies > 1:
+        numeq = dim + 2 + nspecies
```
inducer, line 123: Redundant logic.

test/test_euler.py
```diff
@@ -377,6 +406,8 @@ def test_facial_flux(actx_factory, order, dim):
         iff_split = split_conserved(dim, interior_face_flux)
         assert fnorm(iff_split.mass) < tolerance
         assert fnorm(iff_split.energy) < tolerance
+        if nspecies > 1:
+            assert fnorm(iff_split.massfractions) < tolerance
```
inducer, line 410: We should make it so that this succeeds (i.e. the norm returns 0) for an empty object array, to avoid the branch.

test/test_euler.py
```diff
@@ -410,6 +446,8 @@ def test_facial_flux(actx_factory, order, dim):
         bf_split = split_conserved(dim, boundary_flux)
         assert fnorm(bf_split.mass) < tolerance
         assert fnorm(bf_split.energy) < tolerance
+        if nspecies > 1:
+            assert fnorm(bf_split.massfractions) < tolerance
```
inducer, line 450: We should make it so that this succeeds (i.e. the norm returns 0) for an empty object array, to avoid the branch.

test/test_euler.py
```diff
@@ -431,9 +469,10 @@ def test_facial_flux(actx_factory, order, dim):
     )
 
 
+@pytest.mark.parametrize("nspecies", [1, 2, 3])
 @pytest.mark.parametrize("dim", [1, 2, 3])
 @pytest.mark.parametrize("order", [1, 2, 3])
```
inducer, line 474: These are 27 tests. That's way too many. Use tuple parametrization to avoid the full Cartesian product.

test/test_euler.py
```diff
@@ -660,6 +708,76 @@ def test_lump_rhs(actx_factory, dim, order):
     )
 
 
+@pytest.mark.parametrize("dim", [1, 2, 3])
+@pytest.mark.parametrize("order", [1, 2, 4])
+@pytest.mark.parametrize("nspecies", [1, 2, 10])
+@pytest.mark.parametrize("v0", [0.0, 1.0])
```
inducer, line 714: Way too many.

test/test_euler.py
```diff
@@ -362,9 +383,17 @@ def test_facial_flux(actx_factory, order, dim):
         mom_input = flat_obj_array(
             [discr.zeros(actx) for i in range(discr.dim)]
         )
+        massfrac_input = None
+        if nspecies > 1:
+            massfrac_input = flat_obj_array(
+                [discr.zeros(actx) for i in range(nspecies)]
+            )
+            # massfrac_input = massfrac_input * make_obj_array([1.0 / nspecies])
```
inducer, line 391: ?

test/test_euler.py
```diff
@@ -362,9 +383,17 @@ def test_facial_flux(actx_factory, order, dim):
...
+            massfrac_input = massfrac_input + 1.0 / nspecies
```
inducer, line 392: Hard to read. Just add that to the expression above.

test/test_euler.py
```diff
@@ -466,8 +505,17 @@ def test_uniform_rhs(actx_factory, dim, order):
         mom_input = make_obj_array(
             [discr.zeros(actx) for i in range(discr.dim)]
         )
+
+        massfrac_input = None
+        if nspecies > 1:
+            massfrac_input = flat_obj_array(
+                [discr.zeros(actx) for i in range(nspecies)]
+            )
+            massfrac_input = massfrac_input + 1.0 / nspecies
```
inducer, line 514: Hard to read. Merge into the expression above.

test/test_euler.py
```diff
@@ -660,6 +708,76 @@ def test_lump_rhs(actx_factory, dim, order):
...
+@pytest.mark.parametrize("dim", [1, 2, 3])
+@pytest.mark.parametrize("order", [1, 2, 4])
+@pytest.mark.parametrize("nspecies", [1, 2, 10])
+@pytest.mark.parametrize("v0", [0.0, 1.0])
+def test_multilump_rhs(actx_factory, dim, order, nspecies, v0):
+    """Tests the inviscid rhs using the non-trivial
+    1, 2, and 3D mass lump case against the analytic
+    expressions of the RHS. Checks several different
+    orders and refinement levels to check error behavior.
+    """
```
inducer, line 720: Use the full available line length.

test/test_init.py
```diff
@@ -48,6 +53,63 @@
 logger = logging.getLogger(__name__)
 
 
+@pytest.mark.parametrize("dim", [1, 2, 3])
+@pytest.mark.parametrize("nspecies", [0, 1, 2, 10])
```
inducer, line 57: Can probably also be cut down.

test/test_init.py
```diff
@@ -48,6 +53,63 @@
...
+    exp_mass = 1.0
+    merrmax = discr.norm(cv.mass - exp_mass, np.inf)
+
+    exp_energy = 2.5 + .5 * dim
+    eerrmax = discr.norm(cv.energy - exp_energy, np.inf)
+
+    if nspecies > 1:
+        exp_massfracs = massfracs
+        mferrmax = discr.norm(cv.massfractions - exp_massfracs, np.inf)
+        assert mferrmax < 1e-15
```
inducer, line 106: Avoid branch.

mirgecom/euler.py
```diff
@@ -139,29 +146,55 @@ def _aux_shape(ary, leading_shape):
...
     """
-    assert len(q) == 2+dim
-    return ConservedVars(mass=q[0], energy=q[1], momentum=q[2:2+dim])
+    #    assert len(q) == dim + 2 + get_num_species(dim, q)
+    nspec = get_num_species(dim, q)
+    if nspec > 0:
+        return ConservedVars(mass=q[0], energy=q[1], momentum=q[2:2+dim],
+                             massfractions=q[2+dim:2+dim+nspec])
+    else:
+        return ConservedVars(mass=q[0], energy=q[1], momentum=q[2:2+dim])
```
inducer, line 172: I hadn't really thought about this, but it should be spelled out in the documentation (perhaps of `mirgecom.euler`): I'm guessing the mass fractions add up to one. If so, do we need all n or would n-1 suffice? (and which would should be the "1"?) Can we make it so that there are no branches and the same logic applies always? (I.e. loops over mass fractions transparently become no-ops.) One thing that might help might be to make the mass fractions an empty object array.

mirgecom/euler.py
```diff
@@ -139,29 +146,55 @@ def _aux_shape(ary, leading_shape):
...
+            _aux_shape(momentum, (dim,))])
 
-    result = np.zeros((2+dim,) + aux_shape, dtype=object)
+    result = np.zeros((2+dim+nspec,) + aux_shape, dtype=object)
     result[0] = mass
     result[1] = energy
-    result[2:] = momentum
+    result[2:dim+2] = momentum
+    if nspec > 0:
+        result[dim+2:] = massfractions
```
inducer, line 197: Avoid this branch.

test/test_euler.py
```diff
@@ -660,6 +708,76 @@ def test_lump_rhs(actx_factory, dim, order):
...
+            maxxerr = err_max
+
+        eoc_rec.add_data_point(1.0 / nel_1d, err_max)
+    logger.info(f"Max error: {maxxerr}")
+
+    message = (
+        f"Error for (dim,order) = ({dim},{order}):\n"
+        f"{eoc_rec}"
+    )
+    logger.info(message)
```
inducer, line 774: Why a separate variable?

mirgecom/initializers.py
```diff
@@ -404,19 +412,39 @@ def __call__(self, t, x_vec, eos=IdealSingleGas()):
...
+        #         assert len(spec_center) == self._dim
+        #         self._spec_center[ispec] = spec_center
+        #     else:
+        #         self._spec_center[ispec] = self._center
+        #     lump_loc = self._spec_center[ispec] + t * self._velocity
+        #     if spec_amp > 0:
+        #         amplitude = spec_amp
+        #     if y0 > 0:
+        #         mass0 = y0
+        # else:
```
inducer, line 430: ?

mirgecom/initializers.py
```diff
@@ -360,33 +364,37 @@ def __init__(
...
+            if spec_centers is None:
+                spec_centers = make_obj_array([np.zeros(shape=numdim,)
+                                               for i in range(nspecies)])
+            if spec_amplitudes is None:
+                spec_amplitudes = np.ones(shape=(nspecies,))
+            assert len(spec_y0s) == nspecies
+            assert len(spec_amplitudes) == nspecies
+            assert len(spec_centers) == nspecies
+            for i in range(nspecies):
+                assert len(spec_centers[i]) == numdim
```
inducer, line 386: `assert` is for internal invariants, not for user-facing input checks.

mirgecom/euler.py
MTCam: For just plain-ole scalar advection - which is what we are adding here - there is no such restriction on the Ys.  In fact, they are completely independent.  When we add mixture EOS and mixture Initializers, the restriction of SUM(Y) = 1 will be introduced, and we are currently discussing whether this will be a restriction we need to enforce or whether it will be one we simply monitor.

inducer: > we are currently discussing whether this will be a restriction we need to enforce or whether it will be one we simply monitor.

Where is that discussion taking place?

> When we add mixture EOS and mixture Initializers, the restriction of SUM(Y) = 1 will be introduced

Does it make sense to write code here without knowing the answer to the question above (given what this code should look like depends on the answer to that question)?

MTCam: There are no such requirements on these general transported scalars. They do not need to sum to 1, or have any other properties.  Any such properties need to be imposed by the EOS.  It does not make sense to introduce this arbitrary restriction on the scalars at this point.

inducer: > It does not make sense to introduce this arbitrary restriction on the scalars at this point.

Fair enough. In that case, `nspec` should be renamed `nscalars`.

mirgecom/initializers.py
MTCam: The attempt to extend Lump to support multi-component flows was abandoned in favor of MultiLump, I just failed to rip out the remnants.

mirgecom/initializers.py
MTCam: This was a temporary note to myself, used during a merge.  Should have been removed.

test/test_euler.py
MTCam: Perhaps not, but I perceive that I should test at least nspec x ndim = [1, 10] x [1, 2, 3].

mirgecom/euler.py
```diff
@@ -139,29 +146,55 @@ def _aux_shape(ary, leading_shape):
...
+        aux_shape = single_valued([
+            _aux_shape(mass, ()),
+            _aux_shape(energy, ()),
+            _aux_shape(momentum, (dim,)),
+            _aux_shape(massfractions, (nspec,))])
+    else:
+        aux_shape = single_valued([
+            _aux_shape(mass, ()),
+            _aux_shape(energy, ()),
+            _aux_shape(momentum, (dim,))])
```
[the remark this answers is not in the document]
inducer: I see now what you're getting at. There were a few circular dependencies in my suggestion. I've updated it to hopefully remove those.

test/test_euler.py
MTCam: Because I (and others) cannot figure out how to see our output from tests when it goes into this `logger`.  If I stick it in a message string, then I can also `print` the message.  I don't want to set the logger level so that I see miles of extraneous output before I get to the thing I really want to see.

mirgecom/euler.py
MTCam: I agree completely.  It might be advantageous to allow this slight of expression, at least temporarily.  Can we talk about this again after the full stack merge with Pyrometheus EOS and everything?

test/test_euler.py
MTCam: Unneeded at least.    I think it's better now?

test/test_euler.py
MTCam: I cut down on the number of tests some, but I at least perceive that the remainder of them are needed.  I have the feeling this is not what you wanted.

mirgecom/euler.py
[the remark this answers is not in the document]
MTCam: I tried to make this work - but there is some scoping or bracketing issue that is really messing with the editor/linter.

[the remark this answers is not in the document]
MTCam: I changed it slightly to rearrange the bracketing and set nspec which is needed in the next expression.  It gets the following error:

```py
    def join_conserved(dim, mass, energy, momentum, massfractions=None):
        """Create an agglomerated solution array from the conserved quantities."""
        from pytools import single_valued
        nspec = 0
        if massfractions is not None:
            nspec = len(massfractions)
>       aux_shape = single_valued(*([
            _aux_shape(mass, ()),
            _aux_shape(energy, ()),
            _aux_shape(momentum, (dim,)), ]
            + ([_aux_shape(massfractions, (nspec,))])
            if nspec > 0 else []))
E       TypeError: single_valued() missing 1 required positional argument: 'iterable'
../mirgecom/euler.py:181: TypeError

```

mirgecom/initializers.py
MTCam: Ripping done.

MTCam:
This could use another peek at your convenience, @inducer.

inducer:
Wow. I missed that this was ready for review for a whole week. Sorry!

MTCam:
> Wow. I missed that this was ready for review for a whole week. Sorry!

No worries!  Thanks for your attention on it.  I've continued with prometheus-eos and prometheus-chemistry PRs that should fold in easily after this.  The chemistry one is very WIP.

mirgecom/euler.py
inducer: It doesn't make me happy, but sure: #165

Also, please get out of the habit of writing things like `nspec` when `nspecies` takes the same time to type. Why sacrifice clarity for new users?

mirgecom/initializers.py
inducer: Could you explain why extending the `Lump` for multi-species was difficult?

mirgecom/euler.py
```diff
@@ -95,13 +96,18 @@ class ConservedVars:  # FIXME: Name?
 
         Momentum vector per unit volume
 
+    .. attribute:: massfractions
+
+        Species mass fraction per unit volume
```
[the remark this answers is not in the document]
inducer: That's not how this is marked up (check the Sphinx rendering!). Just write the data type in the text.

mirgecom/euler.py
```diff
@@ -76,31 +77,36 @@
...
         Energy per unit volume
 
-    .. attribute:: momentum
+    .. attribute:: momentum: numpy.ndarray
 
         Momentum vector per unit volume
 
+    .. attribute:: massfractions: numpy.ndarray
+
+        Species mass fraction per unit volume
```
inducer, line 100: Arrays should have their expected shape documented (not just here, in general)

test/test_init.py
inducer: The `if nspecies > 1` is still there though.

test/test_init.py
inducer: Why not just 0 and 10? Do we gain any additional information from the in-between values?

mirgecom/euler.py
```diff
@@ -138,47 +145,88 @@ def _aux_shape(ary, leading_shape):
...
+    nspec = get_num_species(dim, q)
+    if nspec > 0:
+        return ConservedVars(mass=q[0], energy=q[1], momentum=q[2:2+dim],
+                             massfractions=q[2+dim:2+dim+nspec])
+    else:
+        return ConservedVars(mass=q[0], energy=q[1], momentum=q[2:2+dim])
 
 
-def join_conserved(dim, mass, energy, momentum):
+def join_conserved(dim, mass, energy, momentum, massfractions=None):
```
inducer, line 174:
```suggestion
def join_conserved(dim, mass, energy, momentum, mass_fractions=None):
```
as per PEP8.

mirgecom/euler.py
```diff
@@ -76,31 +77,36 @@
...
+    .. attribute:: energy: numpy.ndarray
 
         Energy per unit volume
 
-    .. attribute:: momentum
+    .. attribute:: momentum: numpy.ndarray
 
         Momentum vector per unit volume
 
+    .. attribute:: massfractions: numpy.ndarray
```
inducer, line 98:
```suggestion
    .. attribute:: mass_fractions
```

mirgecom/euler.py
```diff
@@ -76,31 +77,36 @@
...
-    object array.
+    Get the canonical conserved quantities (mass, energy, momentum,
+    mass_fraction) per unit volume = $(\rho,\rho{E},\rho\vec{V},
+    \rho{Y_s})$ from an agglomerated object array.
 
-    .. attribute:: dim
+    .. attribute:: dim: int
 
-    .. attribute:: mass
+    .. attribute:: mass: numpy.ndarray
```
inducer, line 86: Not true. This is a `DOFArray`.

mirgecom/euler.py
```diff
@@ -76,31 +77,36 @@
...
 
         Mass per unit volume
 
-    .. attribute:: energy
+    .. attribute:: energy: numpy.ndarray
```
inducer, line 90: Not true. This is a `DOFArray`.

mirgecom/euler.py
```diff
@@ -76,31 +77,36 @@
...
 
         Energy per unit volume
 
-    .. attribute:: momentum
+    .. attribute:: momentum: numpy.ndarray
 
         Momentum vector per unit volume
```
inducer, line 96: Doc shape (and fix how the type is documented)

mirgecom/euler.py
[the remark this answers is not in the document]
inducer: 71b3d8a54ed53b945ee3ab9e2a8828be35c7530a

mirgecom/initializers.py
```diff
@@ -462,6 +453,228 @@ def exact_rhs(self, discr, q, t=0.0):
...
 
 
+class MultiLump:
+    r"""Solution initializer for multi-species N-dimensional Gaussian lump of mass.
+
+    The Gaussian lump is defined by:
+
+    .. math::
+
+         {\rho}(r) = {\rho}_{0} + {\rho}_{a}{e}^{(1-r^{2})}\\
```
inducer, line 463: Contradiction with below:

>     For multi-species or mixutres, :math:`\rho=1`,

mirgecom/initializers.py
```diff
@@ -462,6 +453,223 @@ def exact_rhs(self, discr, q, t=0.0):
...
+                 for i in range(self._nspecies)]
+            )
+            mass_fracs = make_obj_array([self._spec_y0s[i] + expterms[i]
+                                        for i in range(self._nspecies)])
+
+        return flat_obj_array(mass, energy, mom, mass_fracs)
+
+    def exact_rhs(self, discr, q, t=0.0):
+        """
+        Create the RHS for multlump-of-mass solution at time *t*, locations *x_vec*.
```
inducer, line 610: > multlump

mirgecom/initializers.py
```diff
@@ -462,6 +453,223 @@ def exact_rhs(self, discr, q, t=0.0):
...
+
+        Note that this routine is only useful for testing under the condition of
+        uniform, and constant velocity field.
```
inducer, line 613: Is it possible to create anything else using this IC?

inducer (commented):
Some more thoughts on this below.

mirgecom/euler.py
```diff
@@ -138,47 +147,83 @@ def _aux_shape(ary, leading_shape):
...
+    return dim + 2 + get_num_species(dim, q)
+
+
 def split_conserved(dim, q):
     """Get the canonical conserved quantities.
 
     Return a :class:`ConservedVars` that is the canonical conserved quantities,
     mass, energy, and momentum from the agglomerated object array extracted
-    from the state vector *q*.
+    from the state vector *q*. If present, species mass fractions are included.
```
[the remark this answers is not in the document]
MTCam: So changed.

mirgecom/initializers.py
```diff
@@ -316,31 +315,29 @@ class Lump:
          {\rho}\vec{V} = {\rho}(r)\vec{V_0}\\
          {\rho}E = (\frac{p_0}{(\gamma - 1)} + \frac{1}{2}\rho{|V_0|}^2
```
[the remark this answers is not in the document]
MTCam: Fixed.

mirgecom/euler.py
```diff
@@ -9,13 +9,12 @@
...
 where:
 
--   state $\mathbf{Q} = [\rho, \rho{E}, \rho\vec{V} ]$
--   flux $\mathbf{F} = [\rho\vec{V},(\rho{E} + p)\vec{V},
-    (\rho(\vec{V}\otimes\vec{V}) + p*\mathbf{I})]$,
--   domain boundary $\partial\Omega$,
--   sources $\mathbf{S} = [{(\partial_t{\rho})}_s,
-    {(\partial_t{\rho{E}})}_s, {(\partial_t{\rho\vec{V}})}_s]$
-
+-  state $\mathbf{Q} = [\rho, \rho{E}, \rho\vec{V}, \rho{Y}_\alpha]$
```
inducer, line 12:
```suggestion
-  state $\mathbf{Q} = [\rho, \rho{E}, \rho\vec{V}, \rho{Y}_\beta]$
```

mirgecom/euler.py
```diff
@@ -9,13 +9,12 @@
...
+-  flux $\mathbf{F} = [\rho\vec{V},(\rho{E} + p)\vec{V},
+   (\rho(\vec{V}\otimes\vec{V}) + p*\mathbf{I}), \rho{Y}_\alpha\vec{V}]$,
+-  unit normal $\hat{n}$ to the domain boundary $\partial\Omega$,
```
inducer, line 15:
```suggestion
-  unit normal $\hat{n}$ to the domain boundary $\partial\omega$,
```

inducer (commented):
Nonsense, disregard

mirgecom/initializers.py
MTCam: IMO it is wrong to refer to these as IC.  They can be *used* as IC, but in general they are just solution initializers that can be used at any time (e.g. for boundary solution at arbitrary time, or introducing an acoustic pulse). 

Regardless, the note is still valid because it reminds the user (me, in this case) - that the routine cannot be used to get the expected RHS for unsteady configurations.

mirgecom/initializers.py
```diff
@@ -462,6 +453,223 @@ def exact_rhs(self, discr, q, t=0.0):
...
+    for the mass fraction of each species, *spec_y0s*, and :math:`a_\alpha` is the
+    user-specified vector of amplitudes for each species, *spec_amplitudes*, and
+    :math:`c_\alpha` is the user-specified origin for each species, *spec_centers*.
+
+    A call to this object after creation/init creates the lump solution at a given
+    time (*t*) relative to the configured origin (*center*) and background flow
+    velocity (*velocity*).
+
+    This object also functions as a boundary condition by providing the
+    "get_boundary_flux" method to prescribe exact field values on the given boundary.
```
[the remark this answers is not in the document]
MTCam: It is true for none of them.  The `get_boundary_flux` routine was factored out long ago in the `euler` PR but the docs never updated to reflect the change.

mirgecom/initializers.py
```diff
@@ -462,6 +453,223 @@ def exact_rhs(self, discr, q, t=0.0):
...
+class MultiLump:
+    r"""Solution initializer for multi-species N-dimensional Gaussian lump of mass.
+
+    The Gaussian lump is defined by:
+
+    .. math::
+
+         \rho = 1.0\\
+         {\rho}\vec{V} = {\rho}(r)\vec{V_0}\\
+         {\rho}E = (\frac{p_0}{(\gamma - 1)} + \frac{1}{2}\rho{|V_0|}^
```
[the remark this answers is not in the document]
MTCam: Fixed

mirgecom/initializers.py
```diff
@@ -462,6 +453,223 @@ def exact_rhs(self, discr, q, t=0.0):
...
+         {\rho~Y_\alpha}(r) = {\rho~Y_\alpha}_{0}
+         + {\rho~Y_\alpha}_{a_\alpha}{e}^{(1-{r_\alpha}^{2})}\\
```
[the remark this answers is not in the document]
MTCam: fixed

mirgecom/initializers.py
```diff
@@ -462,6 +453,223 @@ def exact_rhs(self, discr, q, t=0.0):
...
+
+    Where :math:`V_0` is the fixed velocity specified by the user at init time,
```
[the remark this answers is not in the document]
MTCam: fixed

mirgecom/initializers.py
```diff
@@ -462,6 +453,223 @@ def exact_rhs(self, discr, q, t=0.0):
...
+    This object also supplies the exact expected RHS terms from the analytic
+    expression in the "expected_rhs" method.
+
+    .. automethod:: __init__
+    .. automethod:: __call__
+    .. automethod:: exact_rhs
+    """
+
+    def __init__(
+            self, numdim, nspecies=0,
```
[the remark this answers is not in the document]
MTCam: Just a personal preference thing.  I despise the use of plain-ole "dim", and find "numdim" much better to read and understand.  Changed to your preferred term, however.

mirgecom/initializers.py
```diff
@@ -462,6 +453,223 @@ def exact_rhs(self, discr, q, t=0.0):
...
+        r"""Initialize Lump parameters.
+
+        Parameters
+        ----------
+        numdim: int
+            specify the number of dimensions for the lump
+        rho0: float
+            specifies the value of :math:`\rho_0`
+        rhoamp: float
+            specifies the value of :math:`\rho_a`
```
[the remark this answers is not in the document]
MTCam: fixed

mirgecom/initializers.py
MTCam: fixed

mirgecom/initializers.py
MTCam: The two initializers are not related, except that they both use a Gaussian.  I am unsure why I would want to create a relationship through documentation?

mirgecom/initializers.py
```diff
@@ -8,6 +8,7 @@
 .. autoclass:: Lump
 .. autoclass:: Uniform
 .. autoclass:: AcousticPulse
+.. autoclass:: MultiLump
```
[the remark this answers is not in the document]
MTCam: I don't understand the reasoning for this, but done.

mirgecom/euler.py
```diff
@@ -138,47 +147,83 @@ def _aux_shape(ary, leading_shape):
...
     cv = split_conserved(dim, q)
     p = eos.pressure(cv)
 
     mom = cv.momentum
+
+    mass_frac = None
+    if cv.mass_fractions is not None:
+        # mass fractions require a reshape here to get the right numpy
+        # broadcast behavior. Reshaped to [numspecies x 1] object.
+        mass_frac = mom * cv.mass_fractions.reshape(-1, 1) / cv.mass
```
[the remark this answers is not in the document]
MTCam: I have been unable to make this work on my own.   I'll need help with it.  Any chance we can pursue this in a dedicated effort?

mirgecom/initializers.py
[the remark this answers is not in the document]
MTCam: non-existent function

mirgecom/euler.py
```diff
@@ -101,6 +108,7 @@ class ConservedVars:  # FIXME: Name?
     mass: np.ndarray
     energy: np.ndarray
```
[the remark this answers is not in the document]
MTCam: I think I have this fixed up.

mirgecom/euler.py
```diff
@@ -138,47 +147,83 @@ def _aux_shape(ary, leading_shape):
...
     """
-    assert len(q) == 2+dim
-    return ConservedVars(mass=q[0], energy=q[1], momentum=q[2:2+dim])
+    #    assert len(q) == dim + 2 + get_num_species(dim, q)
+    nspec = get_num_species(dim, q)
+    if nspec > 0:
+        return ConservedVars(mass=q[0], energy=q[1], momentum=q[2:2+dim],
+                             mass_fractions=q[2+dim:2+dim+nspec])
+    else:
+        return ConservedVars(mass=q[0], energy=q[1], momentum=q[2:2+dim])
```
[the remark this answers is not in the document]
MTCam: This is still open.  I was unable to make this work using an empty array on my own.  I'll need help with it if we really want to make this change.

mirgecom/initializers.py
```diff
@@ -316,31 +315,29 @@ class Lump:
          {\rho}\vec{V} = {\rho}(r)\vec{V_0}\\
          {\rho}E = (\frac{p_0}{(\gamma - 1)} + \frac{1}{2}\rho{|V_0|}^2
 
-    Where $V_0$ is the fixed velocity specified
-    by the user at init time, and $\gamma$ is taken
-    from the equation-of-state object (eos).
+    Where $V_0$ is the fixed velocity specified by the user at init
+    time, and $\gamma$ is taken from the equation-of-state object (eos).
```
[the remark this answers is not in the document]
MTCam: ?  Hoping I have done what you wanted here, but I don't have much of an idea.

MTCam:
For some reason, ${Y_stuff}$ generated a documentation warning in CI when at the beginning of a sentence.  Moving that expression to a word inside the sentence cleared it up.

mirgecom/initializers.py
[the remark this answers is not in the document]
MTCam: It might be helpful to rename  Lump-->MassLump, and MultiLump-->MulticomponentLump. The "Lump" designation is only there because they both initialize solution components using a Gaussian.

mirgecom/initializers.py
MTCam: I am curious about this, but unsure what advantage it offers here.   Can we discuss it further?

mirgecom/initializers.py
MTCam: > Is it possible to create anything else using this IC?

Specifically, no.  If the user initializes Q using *only* this initializer, which will initialize Q with a uniform flow with disparate, uncoupled bumps in the scalar components, then this routine can be used to pick up the expected RHS at any time t >= t_initial, because it just does Q(r,t) = Q(r - Vt), where V is the uniform and constant velocity that was specified by the user at init time.

So if anything goes into the soln fields Q introducing or inducing an unsteadiness, then the assumption of constant, uniform V is wrong. This note is only here as a reminder to anyone who would use this function in a test (its only real use) to be cautious not to run afoul of the assumptions.

Clearly, if it is has raised a question for you, then the comment isn't clear enough?

mirgecom/initializers.py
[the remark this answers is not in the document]
MTCam: https://github.com/illinois-ceesd/mirgecom/pull/95/commits/939cb3ff0f71ef3730b9fdeeaceb3bb8f2618b82

mirgecom/initializers.py
MTCam: https://github.com/illinois-ceesd/mirgecom/pull/95/commits/939cb3ff0f71ef3730b9fdeeaceb3bb8f2618b82

MTCam: I like this linking in the commit for addressing each comment.  Is this what you had in mind?

mirgecom/initializers.py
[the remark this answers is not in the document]
MTCam: I'll eventually recover from this, and live on :)

mirgecom/initializers.py
```diff
@@ -312,41 +312,35 @@ class Lump:
 
     .. math::
 
-         {\rho}(r) = {\rho}_{0} + {\rho}_{a}\exp^{(1-r^{2})}\\
-         {\rho}\vec{V} = {\rho}(r)\vec{V_0}\\
-         {\rho}E = (\frac{p_0}{(\gamma - 1)} + \frac{1}{2}\rho{|V_0|}^2
+         {\rho} = {\rho}_{0} + {\rho}_{a}\exp^{(1-r^{2})}\\
+         {\rho}\vec{V} = {\rho}\vec{V_0}\\
+         {\rho}E = (\frac{p_0}{(\gamma - 1)} + \frac{1}{2}\rho{|V_0|}^2,
```
majosm, line 317:
```suggestion
         {\rho}E = \frac{p_0}{(\gamma - 1)} + \frac{1}{2}\rho{|V_0|}^2,
```

mirgecom/initializers.py
```diff
@@ -462,6 +450,218 @@ def exact_rhs(self, discr, q, t=0.0):
...
+class MulticomponentLump:
+    r"""Solution initializer for multi-component N-dimensional Gaussian lump of mass.
+
+    The Gaussian lump is defined by:
+
+    .. math::
+
+         \rho = 1.0\\
+         {\rho}\vec{V} = {\rho}\vec{V_0}\\
+         {\rho}E = (\frac{p_0}{(\gamma - 1)} + \frac{1}{2}\rho{|V_0|}^{2}
```
majosm, line 462:
```suggestion
         {\rho}E = \frac{p_0}{(\gamma - 1)} + \frac{1}{2}\rho{|V_0|}^{2}\\
```

mirgecom/initializers.py
```diff
@@ -462,6 +450,218 @@ def exact_rhs(self, discr, q, t=0.0):
...
+         {\rho~Y_\alpha} = {\rho~Y_\alpha}_{0}
+         + {\rho~Y_\alpha}_{a_\alpha}{e}^{(1-{r_\alpha}^{2})},
```
majosm, line 464: `{\rho~Y_\alpha}_{a_\alpha}` is kind of awkward. How about just `{\rho~a_\alpha}`? (And then `a` instead of `{\rho}_{a}` in `Lump` for symmetry?)

Also maybe try using `&` before all the `=`'s to vertically align the equations? Looks a little nicer IMO.

majosm (commented):
Added a few thoughts.

mirgecom/euler.py
```diff
@@ -76,31 +75,40 @@
...
     .. automethod:: join
     .. automethod:: replace
     """
 
-    mass: np.ndarray
-    energy: np.ndarray
-    momentum: np.ndarray
+    mass: DOFArray
+    energy: DOFArray
+    momentum: DOFArray
```
majosm, line 110:
```suggestion
    momentum: np.ndarray
```

(`mass_fractions` should be an `np.ndarray` too, but I'm not sure if that's a one line fix.)

mirgecom/euler.py
```diff
@@ -76,31 +75,40 @@
...
     .. attribute:: momentum
 
-        Momentum vector per unit volume
+        Object array of :class:`~meshmode.dof_array.DOFArray` for momentum density
+        of shape ``(ndim,)``
+
+    .. attribute:: mass_fractions
+
+        Object array of :class:`~meshmode.dof_array.DOFArray` for species mass
+        fractions, $\rho~Y_\alpha$ shape=(nspecies,).
```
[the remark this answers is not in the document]
MTCam: You aren't the only one who picked up on this.   Technically, it should be just "scalars," as you mention, and should include any transported scalar (including non-mixture ones, e.g. like marker particles or ionic species for plasmas).  

The main issue I have with changing it here is that it means we'll need to "do something" to then interpret part of the "scalar" array as mixture species, and some not (at least in name).   Since our current use of this part of the Q array is as mixture species *only* - I sought to allow this slight misnomer until we have non-mixture scalars that would warrant such a distinction.

mirgecom/initializers.py
```diff
@@ -462,6 +450,218 @@ def exact_rhs(self, discr, q, t=0.0):
...
+            rel_poss = make_obj_array([x_vec - lump_locs[i]
+                                       for i in range(self._nspecies)])
+            r2s = make_obj_array([np.dot(rel_poss[i], rel_poss[i])
+                                  for i in range(self._nspecies)])
+            expterms = make_obj_array(
+                [self._spec_amplitudes[i] * actx.np.exp(- r2s[i])
+                 for i in range(self._nspecies)]
+            )
+            mass_fracs = make_obj_array([self._spec_y0s[i] + expterms[i]
+                                        for i in range(self._nspecies)])
```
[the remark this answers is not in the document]
MTCam: Great question.   The use case is testing that each scalar component does the right thing separately.   The purpose of the Multicomponent lump initializer is to add multiple scalar lumps in disparate locations for testing purposes.  Having them in separate locations avoids any potential for getting the final solution (i.e. after advecting the lumps around) right by fluke.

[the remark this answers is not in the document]
majosm: Different amplitudes should be sufficient for that, I would think?

mirgecom/euler.py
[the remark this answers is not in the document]
majosm: My main complaint is just that it's confusing to name it `mass_fractions` when it's really `rho * mass_fractions`. `energy` has this problem too to an extent, but you can kind of hand-wave it away by interpreting it as "energy component of CV" (`rho*E`) instead of literally "energy" (`E`). It's harder (for me at least) to do that with `mass_fractions`. Naming it something like `scalar` would make it more symmetric, IMO:

mass [component of CV]: `rho`
momentum [component of CV]: `rho * V`
energy [component of CV]: `rho * E`
scalar [component of CV]: `rho * Y`

🤷

mirgecom/euler.py
MTCam: I think it is fixed up now.

mirgecom/initializers.py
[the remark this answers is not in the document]
MTCam: :thinking:
Since the RHS is a function of r_alpha, I wanted to makes sure that I had a distinct r_alpha for each component.  I do see that using different amplitudes for each component would be sufficient for making the RHS vary across components.

mirgecom/euler.py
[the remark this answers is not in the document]
MTCam: Alright, I'm convinced and will change it.

mirgecom/initializers.py
```diff
@@ -462,6 +450,218 @@ def exact_rhs(self, discr, q, t=0.0):
...
+            rdotvs = make_obj_array([np.dot(rel_poss[i], v)
+                                     for i in range(self._nspecies)])
+            r2s = make_obj_array([np.dot(rel_poss[i], rel_poss[i])
+                                  for i in range(self._nspecies)])
+            expterms = make_obj_array(
+                [self._spec_amplitudes[i] * actx.np.exp(- r2s[i])
+                 for i in range(self._nspecies)]
+            )
+            specrhs = make_obj_array([2 * expterms[i] * rdotvs[i]
+                                      for i in range(self._nspecies)])
```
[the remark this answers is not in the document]
MTCam: That'd be terrific.  I haven't looked at the symbolic stuff at all and was intrigued by it in our discussions leading up to review.

mirgecom/euler.py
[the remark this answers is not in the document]
MTCam: Changing here #191

mirgecom/euler.py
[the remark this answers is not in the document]
majosm: Trying it out in #196.

mirgecom/initializers.py
inducer: Yep, exactly. It helps me a lot.

mirgecom/initializers.py
inducer: Mostly because people will read the names (which are similar) in the docs and will wonder what the difference is.

MTCam: Indeed. I think the current docs help, but I can add more if needed.  I don't think anyone would use these outside the tests that use them.

mirgecom/initializers.py
MTCam: de488a8

mirgecom/initializers.py
```diff
@@ -462,6 +450,218 @@ def exact_rhs(self, discr, q, t=0.0):
...
+        # mass0 = self._rho0
+        amplitude = 0.0
+        loc_update = t * self._velocity
+        lump_loc = self._center + loc_update
+        # coordinates relative to lump center
+        rel_center = make_obj_array(
+            [x_vec[i] - lump_loc[i] for i in range(self._dim)]
+        )
+        r = actx.np.sqrt(np.dot(rel_center, rel_center))
+        expterm = amplitude * actx.np.exp(- r ** 2)
```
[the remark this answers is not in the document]
MTCam: de488a8

mirgecom/initializers.py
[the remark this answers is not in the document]
MTCam: 1ba6198
84f3a40

mirgecom/initializers.py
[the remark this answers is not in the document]
MTCam: c7022f8
ec6e544

mirgecom/initializers.py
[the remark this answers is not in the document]
MTCam: There may be more associated changes for this scattered about, but the above commit was the brunt of it.

mirgecom/euler.py
MTCam: This changed over several commits.  Can you have another look here, please?
https://github.com/illinois-ceesd/mirgecom/blob/55b37f304d597f27e139d032b9a3ed53006f23e9/mirgecom/euler.py#L109

mirgecom/euler.py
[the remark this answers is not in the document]
MTCam: #196

mirgecom/initializers.py
```diff
@@ -412,8 +412,8 @@ def exact_rhs(self, discr, q, t=0.0):
         """
         Create the RHS for the lump-of-mass solution at time *t*, locations *x_vec*.
 
-        Note that this routine is only useful for testing under the condition of
-        uniform, and constant velocity field.
+        Note that *t* is used to advect the mass lumps under the assumption of
+        constant, and uniform velocity.
```
MTCam, line 417: Changed the comment to this.  Hope it helps.

mirgecom/initializers.py
MTCam: https://github.com/illinois-ceesd/mirgecom/pull/95/commits/ec6e5443ba60de55351a4360f1671e64390c4257#r562656000

test/test_init.py
MTCam: I changed this - but I do perceive that I needed at least 0, 1, 10 when developing this.  0 is just the original test - no scalar components.  1, and 10 species to help me make sure that everything is working when there are (a) any scalar components at all, (b) many scalar components.

test/test_euler.py
```diff
@@ -91,22 +92,32 @@ def rand():
...
     expected_flux[0] = mom
     expected_flux[1] = mom * escale
 
     for i in range(dim):
         for j in range(dim):
             expected_flux[2+i, j] = (mom[i] * mom[j] / mass + (p if i == j else 0))
 
+            #    if nspecies > 1:
+    for i in range(nspecies):
+        expected_flux[dim+2+i] = mom * make_obj_array([mass_fractions[i] / mass])
```
majosm, line 119:
```suggestion
        expected_flux[dim+2+i] = mom * make_obj_array([species_mass[i] / mass])
```
Right? (I don't know what it's actually supposed to be, just comparing with euler.py.)

The `make_obj_array([...])` can be omitted too.

MTCam: Yes, good catch.

mirgecom/initializers.py
```diff
@@ -610,6 +792,7 @@ def exact_rhs(self, discr, q, t=0.0):
         mass[:] = 1.0
         massrhs = 0.0 * mass
         energyrhs = 0.0 * mass
-        momrhs = make_obj_array([0 * mass])
+        momrhs = make_obj_array([0 * mass for i in range(self._dim)])
+        yrhs = make_obj_array([0 * mass for i in range(self._nspecies)])
 
-        return flat_obj_array(massrhs, energyrhs, momrhs)
+        return flat_obj_array(massrhs, energyrhs, momrhs, yrhs)
```
majosm, line 798: (Question: When should `join_conserved` be used vs. `flat_obj_array`?)

test/test_init.py
```diff
@@ -270,3 +333,76 @@ def test_pulse(ctx_factory, dim):
...
+    errmax = discr.norm(p - exp_p, np.inf)
+    assert len(cv.species_mass) == nspecies
+    species_mass = cv.species_mass
+
+    spec_r = make_obj_array([nodes - centers[i] for i in range(nspecies)])
+    r2 = make_obj_array([np.dot(spec_r[i], spec_r[i]) for i in range(nspecies)])
+    expfactor = make_obj_array([spec_amplitudes[i] * actx.np.exp(- r2[i])
+                                for i in range(nspecies)])
+    exp_mass = make_obj_array([rho0 * (spec_y0s[i] + expfactor[i])
+                               for i in range(nspecies)])
```
majosm, line 397: This bothers me a little (the code seems too similar to the code that it's trying to test). Not exactly sure how to fix it though. Maybe if we end up symbolic-izing the initializer this will be fine.

majosm (approved):
Looks good to me! Suggested a couple of minor cosmetic things.

mirgecom/initializers.py
MTCam: I can't actually think of a good reason to use `flat_obj_array` inside the Euler module outside of tests. This particular instance (I believe) is fixed up in the PyrometheusEOS PR sitting just above this one.

test/test_init.py
MTCam: This is an identity test. It is duplicative code, but this bothers me less because the code/test here defines exactly what is expected of the initializer.  If the initializer is broken, then this test will fail.  

If it seems better we *could* create a datafile where the expected state is stored and compare against that - but I don't see how that's effectively different.

mirgecom/initializers.py
MTCam: [Just wanted to remind myself that these were indeed changed.](https://github.com/illinois-ceesd/mirgecom/pull/147/files#diff-98e0920213ef8557c3ac6ded9d1dbf50a9a6c7f6de1b7b998fe0891c3de54ee1)

MTCam:
Did you want to follow-up on this @inducer?

mirgecom/initializers.py
```diff
@@ -312,41 +312,35 @@ class Lump:
...
 
-    This object also functions as a boundary condition
-    by providing the "get_boundary_flux" method to
-    prescribe exact field values on the given boundary.
-
-    This object also supplies the exact expected RHS
-    terms from the analytic expression in the
-    "expected_rhs" method.
+    This object also supplies the exact expected RHS terms from the
+    analytic expression in the :func:`~Lump.exact_rhs` method.
```
inducer, line 327:
```suggestion
    analytic expression through :meth:`exact_rhs`.
```

mirgecom/initializers.py
```diff
@@ -360,35 +354,24 @@ def __init__(
...
-            self._velocity = np.zeros(shape=(numdim,))
-
-        assert len(self._velocity) == numdim
-        assert len(self._center) == numdim
+        if center is None:
+            center = np.zeros(shape=(dim,))
+        if velocity is None:
+            velocity = np.zeros(shape=(dim,))
+        dimmsg = f"is expected to be {dim}-dimensional"
+        if len(center) != dim:
```
inducer, line 362:
```suggestion
        if center.shape != (dim,):
```

mirgecom/initializers.py
```diff
@@ -360,35 +354,24 @@ def __init__(
...
+            raise ValueError(f"Lump center {dimmsg}.")
+        if len(velocity) != dim:
+            raise ValueError(f"Lump velocity {dimmsg}.")
```
inducer, line 365:
```suggestion
        if velocity != (dim,):
            raise ValueError(f"Lump velocity {dimmsg}.")
```

mirgecom/initializers.py
inducer: What's this responding to?

mirgecom/initializers.py
```diff
@@ -425,8 +413,8 @@ def exact_rhs(self, discr, q, t=0.0):
         """
         Create the RHS for the lump-of-mass solution at time *t*, locations *x_vec*.
 
-        Note that this routine is only useful for testing under the condition of
-        uniform, and constant velocity field.
+        Note that *t* is used to advect the mass lumps under the assumption of
+        constant, and uniform velocity.
```
inducer, line 417: (FYI) ReST also has a
```
.. note::

    Some text
```
environment that's useful for this type of thing.

mirgecom/initializers.py
```diff
@@ -462,6 +450,176 @@ def exact_rhs(self, discr, q, t=0.0):
...
+
+    .. math::
+
+         \rho &= 1.0\\
+         {\rho}\vec{V} &= {\rho}\vec{V_0}\\
+         {\rho}E &= \frac{p_0}{(\gamma - 1)} + \frac{1}{2}\rho{|V_0|}^{2}\\
+         {\rho~Y_\alpha} &= {\rho~Y_\alpha}_{0}
+         + {a_\alpha}{e}^{({c_\alpha}-{r_\alpha})^2},
+
+    where $V_0$ is the fixed velocity specified by the user at init time,
```
inducer, line 466: Some `V`s are missing the `\vec`.

inducer, line 466: `\vec{V_0}` doesn't render well: https://notes.tiker.net/tB7jWx_-SGSaB9BdO7IK9

mirgecom/initializers.py
```diff
@@ -462,6 +450,176 @@ def exact_rhs(self, discr, q, t=0.0):
...
+        constant, and uniform velocity.
+
+        Parameters
+        ----------
+        t: float
+            Current time at which the solution is desired
+        x_vec: numpy.ndarray
+            Nodal coordinates
+        eos: :class:`mirgecom.eos.GasEOS`
+            Equation of state class to be used in construction of soln (if needed)
```
inducer, line 557: > if needed

It looks like it's always needed? What does this mean? And: do all EOSs have a `.gamma()`?

IOW: This interface doesn't seem well-specified.

mirgecom/initializers.py
```diff
@@ -462,6 +450,176 @@ def exact_rhs(self, discr, q, t=0.0):
...
+    for the mass fraction of each species, *spec_y0s*, and $a_\alpha$ is the
+    user-specified vector of amplitudes for each species, *spec_amplitudes*, and
+    $c_\alpha$ is the user-specified origin for each species, *spec_centers*.
+
+    A call to this object after creation/init creates the lump solution at a given
+    time (*t*) relative to the configured origin (*center*) and background flow
+    velocity (*velocity*).
+
+    This object also supplies the exact expected RHS terms from the analytic
+    expression in the :func:`~MulticomponentLump.exact_rhs` method.
```
inducer, line 479:
```suggestion
    expression via :meth:`exact_rhs`.
```

test/test_init.py
inducer: I'm kind of with @majosm here: I'm not sure I see the value. But I'm also not going to fight it.

mirgecom/euler.py
```diff
@@ -138,47 +150,69 @@ def _aux_shape(ary, leading_shape):
...
     mom = cv.momentum
+
     return join_conserved(dim,
             mass=mom,
             energy=mom * (cv.energy + p) / cv.mass,
-            momentum=np.outer(mom, mom)/cv.mass + np.eye(dim)*p)
+            momentum=np.outer(mom, mom) / cv.mass + np.eye(dim)*p,
+            # mixture species require a reshape here to get the right numpy
+            # broadcast behavior. Reshaped to [numspecies x 1] object.
+            species_mass=mom * cv.species_mass.reshape(-1, 1) / cv.mass)
```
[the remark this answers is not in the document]
MTCam: 05d1538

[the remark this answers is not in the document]
MTCam: 05d1538

mirgecom/initializers.py
MTCam: velocity.shape?

MTCam: ab1f9fc

mirgecom/initializers.py
MTCam: This really shouldn't be a *note* to self, but part of the documentation for the routine.

ee56b62

mirgecom/euler.py
```diff
@@ -76,31 +76,42 @@
...
     .. automethod:: join
     .. automethod:: replace
     """
 
-    mass: np.ndarray
-    energy: np.ndarray
+    mass: DOFArray
+    energy: DOFArray
     momentum: np.ndarray
+    species_mass: np.ndarray = np.empty((0,), dtype=object)
```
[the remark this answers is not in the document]
MTCam: 32ff631

mirgecom/initializers.py
MTCam: 54230c7

MTCam: 54230c7

mirgecom/initializers.py
MTCam: 54230c7

mirgecom/boundary.py
```diff
@@ -66,7 +66,7 @@ def boundary_pair(
 
         boundary_discr = discr.discr_from_dd(btag)
         nodes = thaw(actx, boundary_discr.nodes())
-        ext_soln = self._userfunc(t, nodes)
+        ext_soln = self._userfunc(nodes, t=t, eos=eos)
```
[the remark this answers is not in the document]
MTCam: #204

MTCam:
I think everything open has been addressed.  Just req follow-up for a quick peek if you'd like.

mirgecom/boundary.py
```diff
@@ -53,20 +53,20 @@ def __init__(self, userfunc):
...
+            of the boundary. The given user function (*userfunc*) must take at
+            least one parameter that specifies the coordinates at which to prescribe
+            the solution.
         """
         self._userfunc = userfunc
 
-    def boundary_pair(
-            self, discr, q, btag, eos, t=0.0
-    ):
+    def boundary_pair(self, discr, q, btag, *args, **kwargs):
```
inducer, line 63: AFAICT `*args get swallowed silently. That shouldn't happen IMO.

mirgecom/euler.py
```diff
@@ -138,47 +150,67 @@ def _aux_shape(ary, leading_shape):
...
-    assert len(q) == 2+dim
-    return ConservedVars(mass=q[0], energy=q[1], momentum=q[2:2+dim])
+    nspec = get_num_species(dim, q)
+    return ConservedVars(mass=q[0], energy=q[1], momentum=q[2:2+dim],
+                         species_mass=q[2+dim:2+dim+nspec])
 
 
-def join_conserved(dim, mass, energy, momentum):
+def join_conserved(dim, mass, energy, momentum, species_mass=np.empty((0,),
+        dtype=object)):
```
inducer, line 174:
```suggestion
def join_conserved(dim, mass, energy, momentum,
        # empty: immutable
        species_mass=np.empty((0,), dtype=object)):
```

inducer:
Did another read-through. Those two things, then IMO this thing is good to go. :tada:

mirgecom/boundary.py
MTCam: I thought I had fixed these.  Must have missed one.

inducer:
Came out looking nice FWIW. I found its final state easy to digest and think about.

mirgecom/boundary.py
inducer, line 63:
```suggestion
    def boundary_pair(self, discr, q, btag, **kwargs):
```

inducer (approved):

MTCam:
> Did another read-through. Those two things, then IMO this thing is good to go. 🎉

Yay indeed!  Thanks so much, AK!

PR #106, opened 2020-09-24
profile: fix footprint mean warning

matthiasdiener:
Fixes
`
/usr/WS1/diener3/Work/emirge/mirgecom/mirgecom/profiling.py:156: FutureWarning: Format strings passed to MaskedConstant are ignored, but in future may error or produce different behavior
  fprint_min, f"{fprint_mean:{g}}", fprint_max,
`

### DIFF
--- a/mirgecom/profiling.py
+++ b/mirgecom/profiling.py
@@ -59,6 +59,7 @@ class ProfileEvent:
 class PyOpenCLProfilingArrayContext(PyOpenCLArrayContext):
     """An array context that profiles kernel executions.
 
+    .. automethod:: clear_profiling_data
     .. automethod:: tabulate_profiling_data
     .. automethod:: call_loopy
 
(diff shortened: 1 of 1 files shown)


mirgecom/profiling.py
```diff
@@ -94,6 +95,11 @@ def _finish_profile_events(self) -> None:
 
         self.profile_events = []
 
+    def clear_profiling_data(self) -> None:
```
inducer, line 98: This is an inherently stateful API, which I don't like. Maybe add the notion of a `ProfilingDataContainer` and allow the user to set that in the array context.

matthiasdiener: Ok, modified the PR to just fix the footprint warning and will that the data container in a separate PR.

inducer (approved):

Issue #80, opened 2020-08-28
advance_state should tell the other ranks if an exception occurs


inducer:

MTCam:
Notably, the current issue in our examples appears to be addressed by this grudge issue https://github.com/inducer/grudge/issues/16

inducer:
>  the current issue in our examples

Which one? Could you add a link to a failing run (/or similar)?

MTCam:
> > the current issue in our examples
> 
> Which one? Could you add a link to a failing run (/or similar)?

The (as-of-yet-unencountered) issue were an exception occurs on one rank, but possibly not others.  Currently, if the grudge norm is MPI/mesh world global, then all ranks will have the exception and it avoids (i.e. fails to reproduce) the potential issue that #80 is intended to report.

PR #107, opened 2020-09-25
Add mpi_entry_point decorator

### LINKED ISSUE #105: Importing from mirgecom forces call to MPI_Init (via mpi4py)
Currently, `mpi4py.MPI` gets imported in mirgecom's [\_\_init\_\_.py](https://github.com/illinois-ceesd/mirgecom/blob/dc69fec5b7fabac2022ff4ddfa223a28d2ff947e/mirgecom/__init__.py#L30). As a result, importing anything from mirgecom will trigger an `MPI_Init`. For most uses this isn't really an issue, but it seems to be causing me some trouble in the context of parallel testing.

As is, when I try to launch MPI from inside a test in mirgecom, it fails (I get a non-zero exit code, and the command doesn't get executed). If I comment out the MPI stuff in \_\_init\_\_.py, it works fine. (This is with OpenMPI on my laptop.) I'm guessing it doesn't like running `mpiexec` from inside a process that has already initialized MPI?

Is there maybe a way we could tweak the code for that warning message (or put it somewhere else) so it doesn't always force an `MPI_Init` on import?

majosm:
Makes it a little more explicit where `MPI_Init()` should occur.

Fixes #105

### DIFF
--- a/doc/support.rst
+++ b/doc/support.rst
@@ -11,6 +11,11 @@ Time Integration Loops
 
 .. automodule:: mirgecom.steppers
 
+MPI
+---
+
+.. automodule:: mirgecom.mpi
+
 Input/Output
 ------------
 
--- a/mirgecom/mpi.py
+++ b/mirgecom/mpi.py
@@ -1,0 +1,65 @@
+""":mod:`mirgecom.mpi` contains MPI helper functionality."""
+
+__copyright__ = """
+Copyright (C) 2020 University of Illinois Board of Trustees
+"""
+
+__license__ = """
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+"""
+
+from functools import wraps
+import os
+import sys
+
+
+def mpi_entry_point(func):
+    """
+    Designate a function as the "main" function for MPI.
+
+    In other words, declares that all MPI code that will be executed on the current
+    process is contained within *func*. Calls `MPI_Init()`/`MPI_Init_thread()` and
+    sets up a hook to call `MPI_Finalize()` on exit.
+    """
+    @wraps(func)
+    def wrapped_func(*args, **kwargs):
+        if "mpi4py.MPI" in sys.modules:
+            raise RuntimeError("mpi4py.MPI imported before designated MPI entry "
+                        "point.")
+
+        # Runs MPI_Init()/MPI_Init_thread() and sets up a hook for MPI_Finalize() on
+        # exit
+        from mpi4py import MPI
+
+        # This code warns the user of potentially slow startups due to file system
+        # locking when running with large numbers of ranks. See
+        # https://mirgecom.readthedocs.io/en/latest/running.html#running-with-large-numbers-of-ranks-and-nodes
+        # for more details
+        size = MPI.COMM_WORLD.Get_size()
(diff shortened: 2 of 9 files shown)


mirgecom/mpi.py
```diff
@@ -0,0 +1,65 @@
...
+        if "mpi4py.MPI" in sys.modules:
+            raise RuntimeError("mpi4py.MPI imported before designated MPI entry "
+                        "point.")
+
+        # Runs MPI_Init()/MPI_Init_thread() and sets up a hook for MPI_Finalize() on
+        # exit
+        from mpi4py import MPI
+
+        # This code avoids slow startups due to file system locking when running with
+        # large numbers of ranks. See
```
matthiasdiener, line 51:
```suggestion
        # This code warns the user of potentially slow startups due to file system locking 
        # when running with large numbers of ranks. See
```

matthiasdiener (approved):
A bit overkill perhaps, but looks ok.

mirgecom/mpi.py
```diff
@@ -0,0 +1,65 @@
...
+
+    In other words, declares that all MPI code that will be executed on the current
+    process is contained within *func*. Calls `MPI_Init()`/`MPI_Init_thread()` and
+    sets up a hook to call `MPI_Finalize()` on exit.
+    """
+    @wraps(func)
+    def wrapped_func(*args, **kwargs):
+        if "mpi4py.MPI" in sys.modules:
+            raise RuntimeError("mpi4py.MPI imported before designated MPI entry "
+                        "point.")
```
inducer, line 44: Provide advice on what to do to fix this.

mirgecom/mpi.py
```diff
@@ -0,0 +1,65 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+"""
+
+from functools import wraps
+import os
+import sys
+
+
+def mpi_entry_point(func):
```
inducer, line 32: Link into the documentation.

mirgecom/mpi.py
inducer, line 44: I wonder if this is a bit harsh. How come you decided to not go with the "initialized" flags you found? I figured we would make sure in `__init__.py` that

- MPI isn't yet initialized, and that
- The flags are set so as to not initialize upon import of `mpi4py.MPI`.

That would make it OK for everyone to import mpi4py as they please, at least downstream.

mirgecom/mpi.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    @wraps(func)
+    def wrapped_func(*args, **kwargs):
+        if "mpi4py.MPI" in sys.modules:
+            raise RuntimeError("mpi4py.MPI imported before designated MPI entry "
+                        "point.")
+
+        # Runs MPI_Init()/MPI_Init_thread() and sets up a hook for MPI_Finalize() on
+        # exit
+        from mpi4py import MPI
```
inducer, line 48: Does this call `MPI_Abort` on exception?

x-ref: #80

mirgecom/mpi.py
majosm: Seems like setting mpi4py to not initialize from inside `__init__.py` would lead to some surprising behavior downstream. For example:

```python
# Without mirgecom import

def print_rank():
    from mpi4py import MPI
    print(f"Rank = {MPI.Get_rank()}")

print_rank()
```

```python
# With mirgecom import

import mirgecom

def print_rank():
    from mpi4py import MPI
    print(f"Rank = {MPI.Get_rank()}")

print_rank()  # Error: Didn't call MPI_Init()
```

The check inside `mpi_entry_point` seemed less intrusive. It's harsh, but using that decorator is entirely optional to begin with, so 🤷.

mirgecom/mpi.py
majosm: I just checked. It does, but only if you use `python -m mpi4py ...` (I think that's the situation this run mode was created to handle, actually).

mirgecom/mpi.py
majosm: I added a little more, but now I wonder if it basically amounts to saying "Don't do that.". Were you thinking it should say something more specific? Is there a way to check where a module was imported, maybe?

inducer: OK, that's a fair point.

mirgecom/mpi.py
inducer: Can we detect this run mode? (And complain if it's not used?)

doc/support.rst
```diff
@@ -11,6 +11,11 @@ Time Integration Loops
 
 .. automodule:: mirgecom.steppers
 
+MPI
```
inducer, line 14: In `running.rst`, document that `-m mpi4py` is the preferred way to run.

mirgecom/mpi.py
majosm: I added a check for this. It looks at whether the `mpi4py.run` module was imported. That module only has mpi4py run mode stuff in it, so I don't _think_ it would get imported otherwise (but I haven't found any guarantees). In my testing it was only there when running with `-m mpi4py`.

inducer (approved):
Thanks for working on this!

matthiasdiener (approved):

doc/support.rst
matthiasdiener: Is this already done?

majosm: I don't think it's explicitly stated anywhere, but it's shown that way in the run script examples, so I figured that was good enough. Also, with this change`-m mpi4py` will become mandatory for running MPI examples (see other comment).

matthiasdiener: Ok, in any case we can add this in another PR.

PR #125, opened 2020-10-27
add example run scripts

matthiasdiener:


doc/running.rst
```diff
@@ -88,6 +91,14 @@ An example batch script for the slurm batch system is given below:
...
+   # Run this script with 'sbatch <script.sh>'
+
+   # Put any environment activation here, e.g.:
+   # source ../../config/activate_env.sh
+
+   # Put any device selection here, e.g.:
+   # export PYOPENCL_CTX=":"
+
    nnodes=$SLURM_JOB_NUM_NODES
    nproc=$nnodes # 1 rank per node
```
inducer, line 104: This bit looks pretty redundant with the script in examples.

How about
```rst
.. literalinclude:: ../examples/lassen.bsub.sh
    :language: bash
```
instead? (Also below)

doc/running.rst
```diff
@@ -65,6 +65,9 @@ Note that profiling has a performance impact (~20% at the time of this writing).
 Running on specific systems
 ---------------------------
 
+This section discusses how to run mirgecom on various clusters.
+There are also several example run scripts in mirgecom's ``examples/`` folder.
```
inducer, line 69: Link to it on Github?

doc/running.rst
matthiasdiener: good idea, changed.

doc/running.rst
matthiasdiener: done

majosm:
Do we have anything in the documentation that describes how to set `PYOPENCL_CTX` to select the GPUs on Lassen? If not, I think that would be good to add (or maybe we could just set it that way by default in the example script?).

matthiasdiener:
> Do we have anything in the documentation that describes how to set `PYOPENCL_CTX` to select the GPUs on Lassen? If not, I think that would be good to add (or maybe we could just set it that way by default in the example script?).

I don't think we have that documentation, but it would be nice to add. Imho we shouldn't really set anything here, since we don't know what the user wants or has installed.

doc/running.rst
```diff
@@ -65,6 +65,10 @@ Note that profiling has a performance impact (~20% at the time of this writing).
 Running on specific systems
 ---------------------------
 
+This section discusses how to run mirgecom on various clusters.
+There are also several example run scripts in mirgecom's ``examples/``
+`folder <https://github.com/illinois-ceesd/mirgecom/tree/master/examples>`_.
```
inducer, line 70:
```suggestion
`folder <https://github.com/illinois-ceesd/mirgecom/tree/master/examples>`__.
```

inducer (approved):

PR #112, opened 2020-10-05
Add heat equation solver

majosm:

PR #126, opened 2020-10-27
Add heat equation solver (continued)

majosm:
Continued from #112 (closed due to CI issues).


majosm:
Wow, even this one is broken. What did I do? 😞

PR #128, opened 2020-10-27
Add heat equation solver (continued)

majosm:
Continued from #112 (closed due to CI issues).

### DIFF
--- a/examples/heat-source-mpi.py
+++ b/examples/heat-source-mpi.py
@@ -1,0 +1,130 @@
+__copyright__ = "Copyright (C) 2020 University of Illinois Board of Trustees"
+
+__license__ = """
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+"""
+
+
+import numpy as np
+import numpy.linalg as la  # noqa
+import pyopencl as cl
+
+from pytools.obj_array import make_obj_array
+
+from meshmode.array_context import PyOpenCLArrayContext
+from meshmode.dof_array import thaw
+
+from meshmode.mesh import BTAG_ALL, BTAG_NONE  # noqa
+
+from grudge.eager import EagerDGDiscretization
+from grudge.shortcuts import make_visualizer
+from mirgecom.integrators import rk4_step
+from mirgecom.heat import heat_operator
+from mirgecom.mpi import mpi_entry_point
+import pyopencl.tools as cl_tools
+
+
+@mpi_entry_point
+def main():
+    cl_ctx = cl.create_some_context()
+    queue = cl.CommandQueue(cl_ctx)
+    actx = PyOpenCLArrayContext(queue,
+        allocator=cl_tools.MemoryPool(cl_tools.ImmediateAllocator(queue)))
+
+    from mpi4py import MPI
+    comm = MPI.COMM_WORLD
+    num_parts = comm.Get_size()
+
+    from meshmode.distributed import MPIMeshDistributor, get_partition_by_pymetis
+    mesh_dist = MPIMeshDistributor(comm)
+
+    dim = 2
+    nel_1d = 16
+
+    if mesh_dist.is_mananger_rank():
+        from meshmode.mesh.generation import generate_regular_rect_mesh
+        mesh = generate_regular_rect_mesh(
+            a=(-0.5,)*dim,
+            b=(0.5,)*dim,
+            n=(nel_1d,)*dim)
+
+        print("%d elements" % mesh.nelements)
+
+        part_per_element = get_partition_by_pymetis(mesh, num_parts)
+
+        local_mesh = mesh_dist.send_mesh_parts(mesh, part_per_element, num_parts)
+
+        del mesh
+
+    else:
+        local_mesh = mesh_dist.receive_mesh_part()
+
+    order = 3
+
+    discr = EagerDGDiscretization(actx, local_mesh, order=order,
+                    mpi_communicator=comm)
+
+    if dim == 2:
+        # no deep meaning here, just a fudge factor
+        dt = 0.0025/(nel_1d*order**2)
+    else:
+        raise ValueError("don't have a stable time step guesstimate")
+
+    source_width = 0.2
+
+    nodes = thaw(actx, discr.nodes())
+
+    fields = make_obj_array([discr.zeros(actx)])
+
+    vis = make_visualizer(discr, order+3 if dim == 2 else order)
+
+    def rhs(t, w):
+        return heat_operator(discr, alpha=1, w=w) + make_obj_array([actx.np.exp(
+            -np.dot(nodes, nodes)/source_width**2)])
+
+    rank = comm.Get_rank()
+
+    vis.write_vtk_file("fld-heat-source-mpi-%03d-%04d.vtu" % (rank, 0),
+            [
+                ("u", fields[0])
+                ])
+
+    t = 0
+    t_final = 0.01
+    istep = 1
+
+    while t < t_final:
+        fields = rk4_step(fields, t, dt, rhs)
+
+        if istep % 10 == 0:
+            print(istep, t, discr.norm(fields[0]))
+
+        vis.write_vtk_file("fld-heat-source-mpi-%03d-%04d.vtu" % (rank, istep),
+                [
+                    ("u", fields[0])
+                    ])
+
+        t += dt
+        istep += 1
+
+
+if __name__ == "__main__":
+    main()
+
+# vim: foldmethod=marker
--- a/examples/heat-source-mpi.py
+++ b/examples/heat-source-mpi.py
@@ -115,9 +115,9 @@ def main():
         if t >= t_final:
             break
 
-        istep += 1
         fields = rk4_step(fields, t, dt, rhs)
         t += dt
+        istep += 1
 
 
 if __name__ == "__main__":
--- a/examples/heat-source-mpi.py
+++ b/examples/heat-source-mpi.py
@@ -35,7 +35,7 @@ from meshmode.mesh import BTAG_ALL, BTAG_NONE  # noqa
 from grudge.eager import EagerDGDiscretization
 from grudge.shortcuts import make_visualizer
 from mirgecom.integrators import rk4_step
-from mirgecom.heat import diffusion_operator
+from mirgecom.diffusion import diffusion_operator
 from mirgecom.mpi import mpi_entry_point
 import pyopencl.tools as cl_tools
 
--- a/examples/heat-source-mpi.py
+++ b/examples/heat-source-mpi.py
@@ -35,7 +35,7 @@ from meshmode.mesh import BTAG_ALL, BTAG_NONE  # noqa
 from grudge.eager import EagerDGDiscretization
 from grudge.shortcuts import make_visualizer
 from mirgecom.integrators import rk4_step
-from mirgecom.heat import heat_operator
+from mirgecom.heat import diffusion_operator
 from mirgecom.mpi import mpi_entry_point
 import pyopencl.tools as cl_tools
 
@@ -95,7 +95,7 @@ def main():
     vis = make_visualizer(discr, order+3 if dim == 2 else order)
 
     def rhs(t, w):
-        return heat_operator(discr, alpha=1, w=w) + make_obj_array([actx.np.exp(
+        return diffusion_operator(discr, alpha=1, w=w) + make_obj_array([actx.np.exp(
             -np.dot(nodes, nodes)/source_width**2)])
 
     rank = comm.Get_rank()
--- a/examples/heat-source-mpi.py
+++ b/examples/heat-source-mpi.py
@@ -25,8 +25,6 @@ import numpy as np
 import numpy.linalg as la  # noqa
 import pyopencl as cl
 
-from pytools.obj_array import make_obj_array
-
 from meshmode.array_context import PyOpenCLArrayContext
 from meshmode.dof_array import thaw
 
@@ -90,13 +88,13 @@ def main():
 
     nodes = thaw(actx, discr.nodes())
 
-    fields = make_obj_array([discr.zeros(actx)])
+    u = discr.zeros(actx)
 
     vis = make_visualizer(discr, order+3 if dim == 2 else order)
 
-    def rhs(t, w):
-        return diffusion_operator(discr, alpha=1, w=w) + make_obj_array([actx.np.exp(
-            -np.dot(nodes, nodes)/source_width**2)])
+    def rhs(t, u):
+        return (diffusion_operator(discr, alpha=1, u=u)
+            + actx.np.exp(-np.dot(nodes, nodes)/source_width**2))
 
     rank = comm.Get_rank()
 
@@ -106,16 +104,16 @@ def main():
 
     while True:
         if istep % 10 == 0:
-            print(istep, t, discr.norm(fields[0]))
+            print(istep, t, discr.norm(u))
             vis.write_vtk_file("fld-heat-source-mpi-%03d-%04d.vtu" % (rank, istep),
                     [
-                        ("u", fields[0])
+                        ("u", u)
                         ])
 
         if t >= t_final:
             break
 
-        fields = rk4_step(fields, t, dt, rhs)
+        u = rk4_step(u, t, dt, rhs)
         t += dt
         istep += 1
 
--- a/examples/heat-source-mpi.py
+++ b/examples/heat-source-mpi.py
@@ -100,28 +100,24 @@ def main():
 
     rank = comm.Get_rank()
 
-    vis.write_vtk_file("fld-heat-source-mpi-%03d-%04d.vtu" % (rank, 0),
-            [
-                ("u", fields[0])
-                ])
-
     t = 0
     t_final = 0.01
-    istep = 1
+    istep = 0
 
-    while t < t_final:
-        fields = rk4_step(fields, t, dt, rhs)
-
+    while True:
         if istep % 10 == 0:
             print(istep, t, discr.norm(fields[0]))
+            vis.write_vtk_file("fld-heat-source-mpi-%03d-%04d.vtu" % (rank, istep),
+                    [
+                        ("u", fields[0])
+                        ])
 
-        vis.write_vtk_file("fld-heat-source-mpi-%03d-%04d.vtu" % (rank, istep),
-                [
-                    ("u", fields[0])
-                    ])
+        if t >= t_final:
+            break
 
-        t += dt
         istep += 1
+        fields = rk4_step(fields, t, dt, rhs)
+        t += dt
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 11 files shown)


majosm:
Yeah, it works if I create the branch here.

majosm:
I think this is ready for review now, finally. There is one lingering oddity, which is that when comparing mirgecom's heat operator directly to nodal-dg's, the error between them grows slightly as the mesh is refined:

N == 1:
-------
```
3.0198326307524304e-15 0.9118906527810351
1.2042989150232512e-14 0.4376217957149026
9.445224670486095e-14 0.20829385445059984
2.783906735706316e-13 0.10121167167644349
1.2134916584028513e-12 0.04985105191980737
```

N == 2:
-------
```
2.5340469209094632e-14 0.5737297961775063
1.836755567855069e-13 0.22971334354626274
9.313678514063255e-13 0.1052864157566982
4.339504590408563e-12 0.050735698584393106
2.0211426541157586e-11 0.024941019365565267
```

N == 3:
-------
```
5.1969650506740275e-14 0.07081506522261419
4.0689302665408125e-13 0.006178915800968799
1.692017144022513e-12 0.0006397549129422477
9.994558983400895e-12 7.276676245892586e-05
3.54880979301179e-11 8.677613611996699e-06
```

N == 4:
-------
```
2.0854998593464584e-13 0.013564377819863853
1.018264530465991e-12 0.001077996970593952
5.63220112264572e-12 0.00010942466185966015
2.9654808220611815e-11 1.2391939114330744e-05
1.2305671015516178e-10 1.476247667758688e-06
```

(Left is error vs. nodal-dg, right is error vs. exact.) Not sure if this a strong vs. weak thing, or maybe just floating point error? In any case, it's still converging to the correct answer, so it's probably not worth dwelling on right now.

majosm:
Also, unfortunately the tests do add some time. Looks like about 5 additional minutes for Ubuntu, 16 (yuck) for Mac. I'm already omitting the order 4 case, and I don't think I can really take any fewer steps... so I'm not sure if I can shorten them further.

matthiasdiener:
```
============================= slowest 10 durations =============================
267.87s call     test/test_heat.py::test_heat_accuracy_and_stability[<array context factory for <pyopencl.Device 'Intel(R) Xeon(R) CPU E5-1650 v2 @ 3.50GHz' on 'Apple'>-problem2-50-5e-05-3]
200.33s call     test/test_heat.py::test_heat_accuracy_and_stability[<array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-problem2-50-5e-05-3]
135.94s call     test/test_heat.py::test_heat_accuracy_and_stability[<array context factory for <pyopencl.Device 'Intel(R) Xeon(R) CPU E5-1650 v2 @ 3.50GHz' on 'Apple'>-problem2-50-5e-05-2]
118.01s call     test/test_heat.py::test_heat_accuracy_and_stability[<array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-problem2-50-5e-05-2]
39.29s call     test/test_wave.py::test_wave_accuracy[<array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-4-problem3]
38.45s call     test/test_heat.py::test_heat_accuracy_and_stability[<array context factory for <pyopencl.Device 'Intel(R) Xeon(R) CPU E5-1650 v2 @ 3.50GHz' on 'Apple'>-problem1-50-5e-05-3]
38.30s call     test/test_heat.py::test_heat_accuracy_and_stability[<array context factory for <pyopencl.Device 'Intel(R) Xeon(R) CPU E5-1650 v2 @ 3.50GHz' on 'Apple'>-problem1-50-5e-05-2]
35.57s call     test/test_heat.py::test_heat_accuracy_and_stability[<array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-problem1-50-5e-05-2]
34.62s call     test/test_wave.py::test_wave_stability[<array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-4-problem3-0.025]
34.21s call     test/test_heat.py::test_heat_accuracy_and_stability[<array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-problem1-50-5e-05-3]
```

Most oft hese tests are duplicated, and we could only keep the ones with POCL, not with Apple OpenCL.

majosm:
Down to 19 minutes from 30 before (but up from current master's 10). Thoughts?

mirgecom/heat.py
```diff
@@ -0,0 +1,115 @@
+r""":mod:`mirgecom.heat` computes the rhs of the heat equation.
```
inducer, line 1: Replace `heat` with `diffusion` (everywhere) to more closely match the role that this will play in Navier-Stokes.

mirgecom/heat.py
```diff
@@ -0,0 +1,115 @@
...
+
+Heat equation:
```
inducer, line 3: Maybe make this part of the `heat_operator` docstring.

inducer, line 3: The docs should contain full sentences.

mirgecom/heat.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def _q_flux(discr, alpha, u_tpair):
+    actx = u_tpair.int[0].array_context
+
+    normal = thaw(actx, discr.normal(u_tpair.dd))
+
+    flux_weak = math.sqrt(alpha)*u_tpair.avg*normal
+
+    return discr.project(u_tpair.dd, "all_faces", flux_weak)
```
inducer, line 53: A lot of empty lines here (also below). Why?

mirgecom/heat.py
```diff
@@ -0,0 +1,115 @@
...
+        the (constant) diffusivity
+    w: numpy.ndarray
+        an object array containing the DOF array state vector
+
+    Returns
+    -------
+    numpy.ndarray
+        an object array containing the DOF array RHS vector
+    """
+    u = w
```
inducer, line 83: ? :shrug:

mirgecom/heat.py
```diff
@@ -0,0 +1,115 @@
...
+    """Compute the RHS of the heat equation.
+
+    Parameters
+    ----------
+    discr: grudge.eager.EagerDGDiscretization
+        the discretization to use
+    alpha: float
+        the (constant) diffusivity
+    w: numpy.ndarray
+        an object array containing the DOF array state vector
```
inducer, line 76: Make this work on a scalar field, and then use the `@obj_array_vectorize` decorator to make it work identically on arrays. Also, document this usage.

test/test_heat.py
```diff
@@ -0,0 +1,241 @@
...
+    [
+        get_decaying_cosine(1, 1.)
+    ])
+def test_heat_compare_to_nodaldg(actx_factory, problem, order, print_err=False):
+    """Compares heat operator to Hesthaven/Warburton Nodal-DG code."""
+    if "NODALDG" in os.environ:
+        nodaldg_path = os.environ["NODALDG"]
+    else:
+        pytest.skip("Nodal-DG code not detected. Set NODALDG environment variable "
+                    "to Nodal-DG code's install path to enable.")
```
inducer, line 176: Maybe look at how meshmode handles nodal-DG comparisons:

- https://github.com/inducer/meshmode/blob/master/test/test_interop.py#L38-L71
- https://github.com/inducer/meshmode/blob/2694be0015f19113d907c49ababc7a5ef62f311b/.github/workflows/ci.yml#L32-L33

test/test_heat.py
```diff
@@ -0,0 +1,241 @@
...
+    sym_rhs = sym_alpha * sym.div(sym.grad(dim, sym_u))
+
+    # f = u_t - rhs
+    sym_f = sym.diff(sym_t)(sym_u) - sym_rhs
+
+    return sym_f, sym_rhs
+
+
+# Note: Must integrate in time for a while in order to acheive expected spatial
+# accuracy. Checking the RHS alone will give lower numbers.
```
inducer, line 94: Also give a (working hypothesis of) an explanation for this.

test/test_heat.py
```diff
@@ -0,0 +1,241 @@
...
+
+import pytest
+import os
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+# Tests below take a problem description as input, which is a tuple
+#   (dim, alpha, mesh_factory, sym_u)
```
inducer, line 45: This might be worth a dataclass.

inducer (commented):
This looks great. Nice job! Ready to go in once the few wrinkles are addressed that I mention below.

mirgecom/heat.py
majosm: Me: _Should it be `w`, or should it be `u`?_ Also me: _Yes._

I dunno, I guess I was thinking about `w` as being the state vector and `u` as being the solution. I think I originally had `u = w[0]`, but then I needed to do `make_obj_array([u])` (basically `w` again) to pass it to `interior_trace_pair`. What do you think, should I maybe just call it `w` everywhere and forget about `u`?

test/test_heat.py
majosm: I don't think I understand quite well enough to write up an explanation, can you suggest one and I'll add it?

inducer, line 94:
```suggestion
# Note: Must integrate in time for a while in order to acheive expected spatial
# accuracy. Checking the RHS alone will give lower numbers.
#
# Working hypothesis: RHS lives in lower order polynomial space and thus doesn't
# attain full-order convergence.
```
(My main point is that we preserve our thought process in the code, to justify why it's there. It may well turn out to be wrong, that's OK.)

test/test_heat.py
```diff
@@ -0,0 +1,241 @@
...
+    # rhs = alpha * div(grad(u))
+    sym_rhs = sym_alpha * sym.div(sym.grad(dim, sym_u))
+
+    # f = u_t - rhs
+    sym_f = sym.diff(sym_t)(sym_u) - sym_rhs
+
+    return sym_f, sym_rhs
+
+
+# Note: Must integrate in time for a while in order to acheive expected spatial
```
matthiasdiener, line 93:
```suggestion
# Note: Must integrate in time for a while in order to achieve expected spatial
```

mirgecom/heat.py
majosm: I ran into a little snag with this... I modified the function to work with `@obj_array_vectorized_n_args` (this is the version I should be using, right?), and it seems like it's working in the array case, but it fails in the scalar case. Taking a closer look, I noticed that when `obj_array_vectorize_n_args` inspects the arguments, it runs this check:
```python
        if isinstance(arg, np.ndarray) and arg.dtype.char == "O":
```
In the scalar case, even though `w` is a bare `DOFArray`, the above condition still evaluates to `True`. So it tries to vectorize, and `heat_operator` ends up getting passed an OpenCL array rather than the `DOFArray` that it expects, leading to errors. Is there a way to work around this, or will I need to vectorize it manually?

inducer: Known problem without a known solution. I suggest you follow [this pattern](https://github.com/inducer/grudge/blob/3d9039d035be81c952775b5d7156e34892be8afb/grudge/eager.py#L86-L90) for now.

majosm:
This should be ready for another look once #138 is in.

majosm:
Ready for review @inducer. I have boundary conditions implemented too, but I think I will put those changes into a new PR since we will probably need to iterate on them a little.

examples/heat-source-mpi.py
```diff
@@ -0,0 +1,130 @@
...
+    def rhs(t, w):
+        return diffusion_operator(discr, alpha=1, w=w) + make_obj_array([actx.np.exp(
+            -np.dot(nodes, nodes)/source_width**2)])
+
+    rank = comm.Get_rank()
+
+    vis.write_vtk_file("fld-heat-source-mpi-%03d-%04d.vtu" % (rank, 0),
+            [
+                ("u", fields[0])
+                ])
```
inducer, line 106: Not liking this repetition much. Maybe use a `while` loop with a `break` in the middle to avoid it?

mirgecom/diffusion.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def _u_flux(discr, alpha, q_tpair):
+    normal = thaw(q_tpair.int[0].array_context, discr.normal(q_tpair.dd))
+    flux_weak = math.sqrt(alpha)*make_obj_array([np.dot(q_tpair.avg, normal)])
+    return discr.project(q_tpair.dd, "all_faces", flux_weak)
+
+
+def _operator(discr, alpha, w):
+    u = make_obj_array([w])
```
inducer, line 53: This nomenclature is exactly backwards to me. `u` evokes "scalar", `w` evokes vector (at least in my head).

inducer, line 53: Why is it necessary to wrap this in an object array? That seems weird and unnecessary.

mirgecom/diffusion.py
```diff
@@ -0,0 +1,116 @@
...
+                    _u_flux(discr, alpha=alpha, q_tpair=tpair)
+                    for tpair in cross_rank_trace_pairs(discr, q))
+                )
+            )[0]
+        )
+
+
+def diffusion_operator(discr, alpha, w):
+    r"""
+    Compute the diffusion operator.
```
inducer, line 90: Document which boundary condition is implemented by this.

mirgecom/diffusion.py
```diff
@@ -0,0 +1,116 @@
...
+
+    dir_u = discr.project("vol", BTAG_ALL, u)
+
+    q = discr.inverse_mass(
+        -math.sqrt(alpha)*discr.weak_grad(u[0])
+        +  # noqa: W504
+        discr.face_mass(
+            _q_flux(discr, alpha=alpha, u_tpair=interior_trace_pair(discr, u))
+            + _q_flux(discr, alpha=alpha,
+                u_tpair=TracePair(BTAG_ALL, interior=dir_u, exterior=-dir_u))
```
inducer, line 63: It might be useful to allow more flexibility in boundary conditions, e.g. Dirichlet on this tag, Neumann on another.

mirgecom/diffusion.py
```diff
@@ -0,0 +1,116 @@
...
+    -------
+    meshmode.dof_array.DOFArray or numpy.ndarray
+        the diffusion operator applied to *w*
+    """
+    if (isinstance(w, np.ndarray)
+            and w.dtype.char == "O"
+            and not isinstance(w, DOFArray)):
+        return obj_array_vectorize(lambda u: _operator(discr, alpha, u), w)
+    else:
+        return _operator(discr, alpha, w)
```
inducer, line 116: Could you make a helper function `obj_array_vectorize_dof` in grudge that packages up this pattern, make `grudge.eager` use it, and use it here (and elsewhere in mirgecom where the pattern occurs)?

test/test_diffusion.py
```diff
@@ -0,0 +1,319 @@
...
+    mesh_factory: Callable
+    sym_u: prim.Expression
+
+
+def get_decaying_cosine(dim, alpha):
+    # 1D: u(x,t) = exp(-alpha*t)*cos(x)
+    # 2D: u(x,y,t) = exp(-2*alpha*t)*cos(x)*cos(y)
+    # 3D: u(x,y,z,t) = exp(-3*alpha*t)*cos(x)*cos(y)*cos(z)
+    # on [-pi/2, pi/2]^{#dims}
+    def mesh_factory(n):
```
inducer, line 78: Method names should start with a verb. (e.g. `get_mesh` or `make_mesh`)

test/test_diffusion.py
```diff
@@ -0,0 +1,319 @@
...
+        if visualize:
+            from grudge.shortcuts import make_visualizer
+            vis = make_visualizer(discr, discr.order+3)
+            vis.write_vtk_file("diffusion_accuracy_{order}_{n}.vtu".format(
+                order=order, n=n), [
+                    ("u", fields[0]),
+                    ("expected_u", expected_fields[0]),
+                    ])
+
+    print("Approximation error:")
```
inducer, line 175:
```suggestion
    print("L^inf error:")
```

test/test_diffusion.py
```diff
@@ -0,0 +1,319 @@
...
+                        / flat_norm(diffusion_u_ndg[0], np.inf))
+
+            if print_err:
+                diffusion_u_exact = make_obj_array([sym_eval_mirgecom(
+                    sym_diffusion_u)])
+                err_exact = (flat_norm(diffusion_u_mirgecom[0]-diffusion_u_exact[0],
+                            np.inf) / flat_norm(diffusion_u_exact[0], np.inf))
+                print(err, err_exact)
+
+            assert err < 1.e-9
```
inducer, line 251:
```suggestion
            assert err < 1e-9
```

inducer (commented):
Thanks! I spotted a few more things, but after that it should be actually good to go in. :)

mirgecom/diffusion.py
majosm: A while back (when working on the wave equation stuff I think) I tried using a scalar and ran into trouble, so I kind of internalized that it needed to be an object array. And after trying it again just now, it looks like it still does. But, judging by the error I think it might just come down to a bug:

```python
    def interior_trace_pair(discrwb, vec):
        i = discrwb.project("vol", "int_faces", vec)
    
        if (isinstance(vec, np.ndarray)
                and vec.dtype.char == "O"
                and not isinstance(vec, DOFArray)):
            e = obj_array_vectorize(
                    lambda el: discrwb.opposite_face_connection()(el),
                    i)
    
>       return TracePair("int_faces", interior=i, exterior=e)
E       UnboundLocalError: local variable 'e' referenced before assignment

../../grudge/grudge/eager.py:360: UnboundLocalError
```

Is there a missing `else` case here?

MTCam: Hi Matt,  Can you please post the error?  This may be exactly what I'm running into with scalar advection.

inducer: > Is there a missing else case here?

Yes, there totally is. Thanks for spotting that.

mirgecom/diffusion.py
majosm: I have that implemented, but I was thinking I'd put that in a new PR to avoid leaving this one outstanding too long. I'm guessing there will be some back and forth over whether/how to connect this up with the Euler BCs, how to deal with BCs for `q`, etc.

inducer: Fair.

examples/heat-source-mpi.py
majosm: How's it look now?

mirgecom/diffusion.py
majosm: Would it maybe make sense to add an optional condition to `obj_array_vectorize` to further limit vectorizing? Then the above could become something like

```python
    obj_array_vectorize(lambda u: _operator(discr, alpha, u), w),
                condition=lambda u: return not isinstance(u, DOFArray))
```
or
```python
    obj_array_vectorize_if(lambda u: _operator(discr, alpha, u), w),
                lambda u: return not isinstance(u, DOFArray))
```
?

inducer: Done in https://github.com/inducer/meshmode/pull/77. I'm hoping to merge that today-ish.

mirgecom/diffusion.py
majosm: Fixed by inducer/grudge#30 (needs merge 🙂).

examples/heat-source-mpi.py
inducer: bc4d122ad70e6f332b2ef57b62801c7f37d77a81 LGTM

mirgecom/diffusion.py
```diff
@@ -0,0 +1,111 @@
...
+from pytools.obj_array import make_obj_array, obj_array_vectorize
+from meshmode.mesh import BTAG_ALL, BTAG_NONE  # noqa
+from meshmode.dof_array import thaw
+from grudge.symbolic.primitives import TracePair
+from grudge.eager import interior_trace_pair, cross_rank_trace_pairs
+
+
+def _q_flux(discr, alpha, u_tpair):
+    normal = thaw(u_tpair.int.array_context, discr.normal(u_tpair.dd))
+    flux_weak = make_obj_array([math.sqrt(alpha)*u_tpair.avg])*normal
```
inducer, line 42:
```suggestion
    flux_weak = scalar([math.sqrt(alpha)*u_tpair.avg])*normal
```
looks less weird to me. How do you feel about that? (`scalar` lives [here](https://github.com/illinois-ceesd/mirgecom/blob/master/mirgecom/euler.py#L168-L170), but it might be better to dump it in some shared utility module)

inducer (approved):
Looks great. One small suggestion, good to go otherwise.

mirgecom/diffusion.py
majosm: Yep, I can do that.

I was wondering though: now that `DOFArray` is no longer an object array, would it make sense to define multiplication (and other operations) between scalar `DOFArray`s and arrays of `DOFArray`s? Maybe something like this:

```diff
     def _bop(self, f, op1, op2):
         """Broadcasting logic for a generic binary operator."""
+        if isinstance(op1, np.ndarray):
+            return obj_array_vectorize(lambda op: self._bop(f, op, op2), op1)
+        if isinstance(op2, np.ndarray):
+            return obj_array_vectorize(lambda op: self._bop(f, op1, op), op2)
         ...
```
?

It seems to work, at least for this case. I'm currently running through all the meshmode/grudge/mirgecom tests to see if it breaks anything.

majosm: Hmm. Runs into problems with expressions that use regular non-object numpy arrays, like `np.eye(dim)*scalar(p)`.

majosm:
I think I'm just going to merge this as is for the moment (don't want it to get out of date again), but I'll keep looking at the broadcasting stuff and submit a new PR for it if I come up with anything (and a PR for the shared `scalar()` otherwise).

inducer:
> I think I'm just going to merge this as is for the moment

:+1:

PR #134, opened 2020-10-28
Make final checkpoint run on all ranks instead of just rank 0

### LINKED ISSUE #133: Final time step output not generated for all ranks
Running on the Illinois campus cluster using the sod-mpi.py example problem with multiple mpi ranks I have an issue with file output at the last time step. The ".pvtu" contains references to individual rank file outputs, however the rank outputs for all ranks except 0 are missing.  This is only an issue at the final time step output.

For testing the final time was reduced in sod-mpi.py
```
t_final = 0.0002
```

Tested with 2 and 3 ranks and observed similar results.

majosm:
Fixes #133

### DIFF
--- a/examples/lump-mpi.py
+++ b/examples/lump-mpi.py
@@ -144,9 +144,9 @@ def main(ctx_factory=cl.create_some_context):
     #    if current_t != checkpoint_t:
     if rank == 0:
         logger.info("Checkpointing final state ...")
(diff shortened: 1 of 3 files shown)


MTCam (approved):
I am unsure how this keeps getting reverted.  I've fixed this a dozen times.

inducer (approved):

inducer:
Thx!

PR #135, opened 2020-10-29
Mark ConservedVars frozen

inducer:


MTCam (approved):
Cool.  Will this help the linters thump a skull if one runs afoul of the constraint?

inducer:
It *could*, but I'm not sure any linters currently have support for it. Things like `cv.energy = (stuff)` will fail at runtime though.

MTCam:
> It _could_, but I'm not sure any linters currently have support for it. Things like `cv.energy = (stuff)` will fail at runtime though.

I see.  Runtime error is better than data corruption, or unknown consequences of using them as I previously had done.

PR #137, opened 2020-11-02
doc: record/replay python package versions

matthiasdiener:
Depends on https://github.com/illinois-ceesd/emirge/pull/83

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -72,7 +72,7 @@ In some cases, it can be helpful to install certain packages from source, for de
 a git version. Most packages are straightforward to install from source. For pocl, you can follow this
 `installation script <https://gist.github.com/matthiasdiener/838ccbdb5d8f4e4917b58fe3da811777>`__.
 
-.. _Pyopencl source:
+.. _Pyopencl source installation:
 
 How can I build pyopencl from source?
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -111,7 +111,8 @@ exact versions of the packages at a later time::
 .. note::
 
    This will build pyopencl by source, which can be challenging on some systems. Please
-   see :ref:`Pyopencl source` for details.
+   see :ref:`Pyopencl source installation` for information on the prerequisites of Pyopencl,
+   or remove/comment the pyopencl line from the generated requirements file.
 
 .. note::
 
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -116,8 +116,6 @@ exact versions of the packages at a later time::
 
 .. note::
 
-   You can also install the packages from the created requirements.txt into a new conda environment::
+   You can also install the packages from the created requirements.txt with a new emirge installation::
    
-      $ conda create -n new_env
-      $ conda activate new_env
-      $ pip install --src . -r myreq.txt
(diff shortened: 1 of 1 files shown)


doc/faq.rst
```diff
@@ -87,3 +87,20 @@ You can build pyopencl against conda's OpenCL driver in the following way::
...
+exact versions of the packages at a later time::
+
+   $ cd emirge/
+   $ ./version.sh
+   [...]
+   *** Creating requirements file with current emirge module versions
+   [...]
+   *** Created file 'requirements.txt.XXXXXX'. Install it with 'pip install --src . -r requirements.txt.XXXXXX'.
+
+   $ pip install --src . -r requirements.txt.XXXXXX
```
inducer, line 106: See https://github.com/illinois-ceesd/emirge/pull/83. I think "fresh env" will be safer, more reproducible, and more extensible to when the procedure includes conda packages.

inducer (commented):
Thanks. As I said in https://github.com/illinois-ceesd/emirge/pull/83#pullrequestreview-521777909, we should make it so that the command line procedure doesn't change too drastically when we start incorporating conda env files.

doc/faq.rst
```diff
@@ -87,3 +89,27 @@ You can build pyopencl against conda's OpenCL driver in the following way::
...
+exact versions of the packages at a later time::
+
+   $ cd emirge/
+   $ ./version.sh --output-requirements=myreq.txt
+   [...]
+   *** Creating requirements file with current emirge module versions
+   [...]
+   *** Created file 'myreq.txt'. Install it with 'pip install --src . -r myreq.txt'.
+
+   $ pip install --src . -r myreq.txt
```
inducer, line 108: Also mention that the same `requirements.txt` can be used to create a new environment. Looks good otherwise.

matthiasdiener: added.

doc/faq.rst
```diff
@@ -87,3 +89,35 @@ You can build pyopencl against conda's OpenCL driver in the following way::
...
+   see :ref:`Pyopencl source installation` for information on the prerequisites of Pyopencl,
+   or remove/comment the pyopencl line from the generated requirements file.
+
+.. note::
+
+   You can also install the packages from the created requirements.txt into a new conda environment::
+   
+   $ conda create -n new_env
+   $ conda activate new_env
+   $ pip install --src . -r myreq.txt
```
inducer, line 123: This block needs indenting in order to render properly.

inducer, line 123: I figured that the `requirements.txt` could be provided to `install.sh` via a flag?

matthiasdiener: Fixed.

matthiasdiener: No, that won't work for now. Our parser doesn't handle the `--editable` flag inside the requirements.txt.

matthiasdiener:
Rendered html is here: https://mirgecom.readthedocs.io/en/doc-req-versions.sh/faq.html

doc/faq.rst
inducer: It should. Otherwise why are we building systems that don't make sense and where nothing fits together?

matthiasdiener: Done in https://github.com/illinois-ceesd/emirge/pull/88

inducer (approved):
Thanks!

PR #138, opened 2020-11-04
Misc. fixes from heat equation PR

majosm:
Pulled a few things out of the heat PR so I can keep it more heat focused (and so I can squash it later).

### DIFF
--- a/test/test_wave.py
+++ b/test/test_wave.py
@@ -120,6 +120,7 @@ def sym_wave(dim, sym_phi):
     return sym_u, sym_v, sym_f, sym_rhs
 
 
+@pytest.mark.parametrize("order", [2, 3, 4])
 @pytest.mark.parametrize("problem",
     [
         get_standing_wave(2),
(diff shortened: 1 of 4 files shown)


test/test_wave.py
```diff
@@ -120,20 +142,20 @@ def sym_wave(dim, sym_phi):
...
         get_manufactured_cubic(3)
     ])
-@pytest.mark.parametrize("order", [2, 3, 4])
 def test_wave_accuracy(actx_factory, problem, order, visualize=False):
     """Checks accuracy of the wave operator for a given problem setup.
     """
     actx = actx_factory()
 
-    dim, c, mesh_factory, sym_phi = problem
+    dim, c, mesh_factory, sym_phi = dataclasses.astuple(problem)
```
inducer, line 158: This feels potentially error-prone, say if somebody shuffles the fields. Why not go with `p = problem` and then reference the variables as `p.dim` and `p.mesh_factory`? Not so much longer.

inducer (approved):
Thanks! LGTM, one optional suggestion.

inducer (approved):

PR #146, opened 2020-11-10
Add variable diffusivity to diffusion operator

majosm:


majosm:
@inducer, @lukeolson I took a stab at implementing the DG version of `q = sqrt(alpha(x))*grad(u)` the way we talked about on Friday. Relevant code is [here](https://github.com/illinois-ceesd/mirgecom/pull/146/files#diff-8c80087b48245eecc882a14fa78e26b0891982bff838e184fcc80bbb396b7465R155-R158). The good news is that it's running and not blowing up. The bad news is that it's converging to the wrong answer. 🙁 I'm using a steady-state manufactured case `u(x,t) = cos(x)` on `[-pi/2,pi/2]` with `alpha(x) = 1 + 0.05*x^2`.

The error looks like:
error
i.e., it's converging to something, but not the expected result.

Does anything look obviously off with my implementation? Here's a reminder of roughly what I'm trying to implement:
Screen Shot 2020-11-10 at 1 03 45 PM
where `g = sqrt(alpha)` and `h = u`.

Some things I'm not sure about that may be contributing:
* How do I use overintegration for the first term?
* Is it allowed to just take a DG derivative and then use it without adding extra "flux" terms to couple the result together (looking at that `grad(sqrt(alpha))` in the first term)? Wondering if maybe this is another instance of the same pitfall that led to `q` being introduced...

majosm:
Hang on, I found a bug in my test. Let's see if that fixes it.

majosm:
Well, that was anticlimactic. 🙂 Seems to be working now. Still kind of wondering about these two things from the other comment though:

> * How do I use overintegration for the first term?
> * Is it allowed to just take a DG derivative and then use it without adding extra "flux" terms to couple the result together (looking at that `grad(sqrt(alpha))` in the first term)? Wondering if maybe this is another instance of the same pitfall that led to `q` being introduced...

inducer:
> How do I use overintegration for the first term?

Let's discuss tomorrow. Could you add that to our agenda?

> s it allowed to just take a DG derivative

I assume that by "DG derivative" you mean "just the volume derivative"? Generally, that's not consistent with anything. Volume derivative + central fluxes is generically consistent with the distributional derivative, i.e. a version of the derivative that tolerates discontinuities (at the cost of introducing delta functions). Any deviations from that are only legitimized by consistency/stability proofs.

PR #152, opened 2020-11-19
fix rtd

matthiasdiener:


.rtd-env-py3.yml
```diff
@@ -6,7 +6,11 @@ dependencies:
 - python=3.8
 - mpi4py
 - islpy
+- pip
 - pip:
+  - "pybind11"
+  - "mako"
+  - "pyopencl"
```
inducer, line 13: Why get pyopencl from pip (where it might need compilation)? (rather than conda, where it doesn't)

matthiasdiener: Ok, that fixed it.

.rtd-env-py3.yml
```diff
@@ -6,6 +6,8 @@ dependencies:
 - python=3.8
 - mpi4py
 - islpy
+- pip
```
inducer, line 9:
```suggestion
```

inducer (approved):
Good to merge after that suggestion

.rtd-env-py3.yml
matthiasdiener: I added this because of this warning message:
```
Warning: you have pip-installed dependencies in your environment file, but you do not list pip itself as one of your conda dependencies.  Conda may not use the correct pip to install your packages, and they may end up in the wrong place.  Please add an explicit pip dependency.  I'm adding one for you, but still nagging you.
```

inducer: Huh, OK. TIL. Never mind then!

PR #153, opened 2020-11-19
Update prometheus object to take numpy-like namespace for numpy-like ops

MTCam:
This changes the prometheus uiuc mechanism port to use a passed-in numpy context.   Just testing and discussing here.  Once we have this worked out, we'll update the pyrometheus code generator to reflect our current expression for the themochemistry object.

cc: @inducer , @ecisneros8

### DIFF
--- a/mirgecom/prometheus.py
+++ b/mirgecom/prometheus.py
@@ -1,9 +1,9 @@
-import numpy as realnp
+import numpy as np
 from pytools.obj_array import make_obj_array
 
 
 class UIUCMechanism:
-    def __init__(self, np):
+    def __init__(self, npctx):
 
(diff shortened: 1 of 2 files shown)


inducer:
It's also possible to assign `np` to a local vairable in each function:
```py
np = self.np
```
That way, the rest of the code after that looks less foreign.

mirgecom/prometheus.py
```diff
@@ -1,9 +1,9 @@
-import numpy as np
+import numpy as realnp
```
lukeolson, line 1: For clarity I think `import numpy as np` should be used.  And then a different name for the pseudo numpy.

PR #156, opened 2020-11-20
Add boundary conditions to diffusion operator

majosm:


mirgecom/diffusion.py
```diff
@@ -49,22 +48,26 @@ def _u_flux(discr, alpha, q_tpair):
...
     discr: grudge.eager.EagerDGDiscretization
         the discretization to use
     alpha: float
         the (constant) diffusivity
+    u_boundaries:
+        dictionary (or object array of dictionaries) of boundary functions for *u*,
+        one for each valid btag
+    q_boundaries:
+        dictionary (or object array of dictionaries) of boundary functions for
+        :math:`q = \sqrt{\alpha}\nabla u`, one for each valid btag
```
inducer, line 70: Document the signatures of these boundary functions.

inducer, line 70: Would it make sense to expose Dirichlet and Neumann (and perhaps Robin) as the classical second-order BCs, rather than expecting users to piece together their own out of boundary states for `u` and `q`? (Honest question: I'm not presuming the answer to this is yes.)

mirgecom/diffusion.py
```diff
@@ -75,34 +78,39 @@ def diffusion_operator(discr, alpha, u):
...
     """
     if isinstance(u, np.ndarray):
-        return obj_array_vectorize(lambda u_i: diffusion_operator(discr, alpha, u_i),
-                    u)
-
-    dir_u = discr.project("vol", BTAG_ALL, u)
+        if not isinstance(u_boundaries, np.ndarray) or len(u_boundaries) != len(u):
+            raise RuntimeError("u_boundaries must be the same length as u")
+        if not isinstance(q_boundaries, np.ndarray) or len(q_boundaries) != len(u):
+            raise RuntimeError("q_boundaries must be the same length as u")
```
inducer, line 84: Separate out checks for type and length?

test/test_diffusion.py
```diff
@@ -70,24 +72,35 @@ class HeatProblem:
     sym_u: prim.Expression
 
 
-def get_decaying_cosine(dim, alpha):
+def get_decaying_trig(dim, alpha):
     # 1D: u(x,t) = exp(-alpha*t)*cos(x)
-    # 2D: u(x,y,t) = exp(-2*alpha*t)*cos(x)*cos(y)
-    # 3D: u(x,y,z,t) = exp(-3*alpha*t)*cos(x)*cos(y)*cos(z)
+    # 2D: u(x,y,t) = exp(-2*alpha*t)*sin(x)*cos(y)
+    # 3D: u(x,y,z,t) = exp(-3*alpha*t)*sin(x)*sin(y)*cos(z)
```
inducer, line 78: You have one test (per dimension) with mixed BCs. Do you feel that that adequately supports that all relevant BCs work?

inducer, line 78: Would it make sense to also test inhomogeneous Dirichlet/Neumann (not just homogeneous)?

inducer (commented):
Thanks for working on this! I overlooked the "Draft" marker and decided to take a look anyway...

test/test_diffusion.py
majosm: Added.

majosm: Ideally it would be nice to have separate tests for each BC type, etc., but given how long they take to run I'm a little hesitant to add more. As far as I can tell the existing tests should cover things adequately enough.

mirgecom/diffusion.py
majosm: I made an attempt... is this kind of what you had in mind API-wise?

mirgecom/diffusion.py
```diff
@@ -49,24 +53,102 @@ def _u_flux(discr, alpha, q_tpair):
...
+    Diffusion boundary base class.
+
+    .. automethod:: get_q_flux
+    .. automethod:: get_u_flux
+    """
+
+    @abc.abstractmethod
+    def get_q_flux(self, discr, alpha, dd, u):
+        """Compute the flux for *q* on the boundary corresponding to *dd*."""
+        pass
```
inducer, line 67:
```suggestion
        raise NotImplementedError
```

mirgecom/diffusion.py
```diff
@@ -49,24 +53,102 @@ def _u_flux(discr, alpha, q_tpair):
...
+
+    def get_u_flux(self, discr, alpha, dd, q):
+        """Compute the flux for *u* on the boundary corresponding to *dd*."""
+        pass
```
inducer, line 71:
```suggestion
        raise NotImplementedError
```

mirgecom/diffusion.py
```diff
@@ -49,24 +53,102 @@ def _u_flux(discr, alpha, q_tpair):
...
+
+
+class DirichletDiffusionBoundary(DiffusionBoundary):
+    """
+    Dirichlet boundary condition for the diffusion operator.
+
+    .. automethod:: __init__
+    """
+
+    def __init__(self, value=0.):
```
inducer, line 81:
```suggestion
    def __init__(self, value):
```
[PEP20](https://www.python.org/dev/peps/pep-0020/): Explicit is better than implicit.

mirgecom/diffusion.py
```diff
@@ -49,24 +53,102 @@ def _u_flux(discr, alpha, q_tpair):
...
+
+
+class NeumannDiffusionBoundary(DiffusionBoundary):
+    """
+    Neumann boundary condition for the diffusion operator.
+
+    .. automethod:: __init__
+    """
+
+    def __init__(self, value=0.):
```
inducer, line 110:
```suggestion
    def __init__(self, value):
```

mirgecom/diffusion.py
```diff
@@ -49,24 +53,102 @@ def _u_flux(discr, alpha, q_tpair):
...
-    Currently sets $u = 0$ on the boundaries.
-
     Parameters
     ----------
     discr: grudge.eager.EagerDGDiscretization
         the discretization to use
     alpha: float
         the (constant) diffusivity
+    boundaries:
+        dictionary (or object array of dictionaries) mapping boundary tags to
```
inducer, line 148: Object array of dictionaries doesn't make much sense, since dictionaries don't meaningfully support arithmetic. Tuple or list?

mirgecom/diffusion.py
```diff
@@ -49,24 +53,102 @@ def _u_flux(discr, alpha, q_tpair):
...
+
+        Parameters
+        ----------
+        value: float or meshmode.dof_array.DOFArray
+            the value(s) along the boundary
+        """
+        self.value = value
+
+    def get_q_flux(self, discr, alpha, dd, u):  # noqa: D102
+        dir_u = discr.project("vol", dd, u)
```
inducer, line 122: `dir` sseems like a misnomer... (everywhere in Neumann)

mirgecom/diffusion.py
```diff
@@ -49,24 +53,102 @@ def _u_flux(discr, alpha, q_tpair):
...
+        u_tpair = TracePair(dd, interior=dir_u, exterior=2.*self.value-dir_u)
```
inducer, line 94: Mention whether you have a source for this expression.

mirgecom/diffusion.py
```diff
@@ -49,24 +53,102 @@ def _u_flux(discr, alpha, q_tpair):
...
+    def get_q_flux(self, discr, alpha, dd, u):  # noqa: D102
+        dir_u = discr.project("vol", dd, u)
+        u_tpair = TracePair(dd, interior=dir_u, exterior=dir_u)
+        return _q_flux(discr, alpha, u_tpair)
+
+    def get_u_flux(self, discr, alpha, dd, q):  # noqa: D102
+        ones = discr.zeros(q[0].array_context) + 1.
+        dir_ones = discr.project("vol", dd, ones)
+        flux_weak = alpha*self.value*dir_ones
+        return discr.project(dd, "all_faces", flux_weak)
```
inducer, line 130: Mention whether you have a source for this expression.

inducer, line 130: Dirichlet uses penalization, but Neumann doesn't. How come?

inducer (commented):

mirgecom/diffusion.py
majosm: Ohhhhh... that's what `dir_` means. 🙂 I had been wondering. Fixed.

mirgecom/diffusion.py
majosm: It's only an object array when `u` is an object array. I think I need that for `obj_array_vectorize_n_args` to work, right?

majosm:
@inducer I think I'm going to hold off on the `u`/`q` reordering we talked about and apply that to the variable diffusivity branch instead of this one (to avoid potential merging nightmares).

mirgecom/diffusion.py
```diff
@@ -49,24 +53,135 @@ def _u_flux(discr, alpha, q_tpair):
...
+    def get_q_flux(self, discr, alpha, dd, u):  # noqa: D102
+        u_int = discr.project("vol", dd, u)
+        u_tpair = TracePair(dd, interior=u_int, exterior=u_int)
+        return _q_flux(discr, alpha, u_tpair)
+
+    def get_u_flux(self, discr, alpha, dd, q):  # noqa: D102
+        ones = discr.zeros(q[0].array_context) + 1.
+        bdry_ones = discr.project("vol", dd, ones)
+        # Computing the flux directly instead of constructing an external q value
+        # avoids some potential nastiness with sqrt(alpha) and overintegration in
```
inducer, line 160: > nastiness

Be more precise

mirgecom/diffusion.py
```diff
@@ -49,24 +53,135 @@ def _u_flux(discr, alpha, q_tpair):
...
+        # the variable alpha case
```
inducer, line 161:
```suggestion
        # the case of spatially varying alpha.
```

mirgecom/diffusion.py
```diff
@@ -49,24 +53,135 @@ def _u_flux(discr, alpha, q_tpair):
...
+
+    def get_q_flux(self, discr, alpha, dd, u):  # noqa: D102
+        u_int = discr.project("vol", dd, u)
+        u_tpair = TracePair(dd, interior=u_int, exterior=u_int)
+        return _q_flux(discr, alpha, u_tpair)
+
+    def get_u_flux(self, discr, alpha, dd, q):  # noqa: D102
+        ones = discr.zeros(q[0].array_context) + 1.
+        bdry_ones = discr.project("vol", dd, ones)
+        # Computing the flux directly instead of constructing an external q value
```
inducer, line 159:
```suggestion
        # Computing the flux directly instead of constructing an external q value (and the associated TracePair)
```

mirgecom/diffusion.py
inducer: If you'd like to use them internally to make use of `obj_array_vectorize_n_args`, that's OK, but IMO it's not a sane interface to expose to the user.

inducer: (I.e. you could accept a list and internally convert it.)

mirgecom/diffusion.py
majosm: How's it look now?

inducer: LGTM!

inducer (approved):
LGTM now. Thanks for working on this, and happy new year!

PR #160, opened 2020-11-23
add pytato dependency

matthiasdiener:


inducer:
Since it's not used yet (and still stablizing), I'd say we leave this for now. It'll come once we start using it in earnest.

inducer (approved):

matthiasdiener:
When we actually start using it, we'll also need to add it to `setup.py`.

PR #168, opened 2020-12-04
Refactor checkpoint code

### LINKED ISSUE #167: Final solution dump after exception can run into "file exists" error
This should not happen IMO.

majosm:
I took a stab at refactoring the checkpoint code to: (1) make it a little easier to follow, and (2) fix #167.

One potential downside of the approach I took here is that there is now a little more code on the driver side. But, IMO that code is fairly easy to follow and less likely to need frequent refactoring vs. a big monolithic "do everything" function.

Thoughts?

### DIFF
--- a/examples/vortex-mpi.py
+++ b/examples/vortex-mpi.py
@@ -36,14 +36,14 @@ from grudge.eager import EagerDGDiscretization
 from grudge.shortcuts import make_visualizer
 
 
-from mirgecom.euler import inviscid_operator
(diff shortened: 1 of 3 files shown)


inducer:
Thanks for taking a stab at this! My preference would be to aim at reducing the volume of code in the driver and making it so that a lot of this logic is common and reused, which is why this doesn't totally reflect the goal state I had in mind.

majosm:
Superseded by #257.

PR #171, opened 2020-12-08
vortex-mpi: Don't run in 3D, fix warning

matthiasdiener:
Background: Running this with `dim=3` just hangs.

### DIFF
--- a/examples/vortex-mpi.py
+++ b/examples/vortex-mpi.py
@@ -94,7 +94,7 @@ def main(ctx_factory=cl.create_some_context):
     rank = comm.Get_rank()
 
     if dim != 2:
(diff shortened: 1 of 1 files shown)


examples/vortex-mpi.py
```diff
@@ -93,6 +93,9 @@ def main(ctx_factory=cl.create_some_context):
     comm = MPI.COMM_WORLD
     rank = comm.Get_rank()
 
+    if dim != 2:
+        raise RuntimeError("This example must be run with dim = 2.")
```
inducer, line 97:
```suggestion
        raise ValueError("This example must be run with dim = 2.")
```

inducer (approved):
Thx!

PR #174, opened 2020-12-14
doc: split into separate files

matthiasdiener:
- Allows hierarchical content menu
- Reduces 'wall of text'

Rendered: https://mirgecom.readthedocs.io/en/doc-split/


inducer (approved):
I like (in the sense of personal preference) the individual RSTs a bit longer than you've made them, but I'm OK with this.

matthiasdiener:
> I like (in the sense of personal preference) the individual RSTs a bit longer than you've made them, but I'm OK with this.

I agree, and we should address this in the future. For now, this PR just aims to replicate the structure of the current manual, but with the nesting.

PR #180, opened 2020-12-18
Add variable diffusivity to diffusion_operator

majosm:
Depends on #156 and inducer/grudge#37.

### DIFF
--- a/examples/heat-source-mpi.py
+++ b/examples/heat-source-mpi.py
@@ -33,6 +33,7 @@ from meshmode.mesh import BTAG_ALL, BTAG_NONE  # noqa
 from grudge.eager import EagerDGDiscretization
 from grudge import sym as grudge_sym
 from grudge.shortcuts import make_visualizer
+from grudge.symbolic.primitives import QTAG_NONE
 from mirgecom.integrators import rk4_step
 from mirgecom.diffusion import (
     diffusion_operator,
@@ -109,7 +110,8 @@ def main():
     vis = make_visualizer(discr, order+3 if dim == 2 else order)
 
     def rhs(t, u):
-        return (diffusion_operator(discr, alpha=alpha, boundaries=boundaries, u=u)
+        return (diffusion_operator(
+            discr, quad_tag=QTAG_NONE, alpha=alpha, boundaries=boundaries, u=u)
             + actx.np.exp(-np.dot(nodes, nodes)/source_width**2))
 
     rank = comm.Get_rank()
--- a/examples/heat-source-mpi.py
+++ b/examples/heat-source-mpi.py
@@ -97,17 +97,19 @@ def main():
 
     nodes = thaw(actx, discr.nodes())
 
-    u = discr.zeros(actx)
-
-    vis = make_visualizer(discr, order+3 if dim == 2 else order)
+    alpha = discr.zeros(actx) + 1.
 
     boundaries = {
         grudge_sym.DTAG_BOUNDARY("dirichlet"): DirichletDiffusionBoundary(0.),
         grudge_sym.DTAG_BOUNDARY("neumann"): NeumannDiffusionBoundary(0.)
     }
 
+    u = discr.zeros(actx)
+
+    vis = make_visualizer(discr, order+3 if dim == 2 else order)
+
     def rhs(t, u):
-        return (diffusion_operator(discr, alpha=1, boundaries=boundaries, u=u)
+        return (diffusion_operator(discr, alpha=alpha, boundaries=boundaries, u=u)
             + actx.np.exp(-np.dot(nodes, nodes)/source_width**2))
 
     rank = comm.Get_rank()
--- a/mirgecom/diffusion.py
+++ b/mirgecom/diffusion.py
@@ -31,6 +31,7 @@ THE SOFTWARE.
 """
 
 import abc
+import math
 import numpy as np
 import numpy.linalg as la  # noqa
 from pytools.obj_array import make_obj_array, obj_array_vectorize_n_args
@@ -39,6 +40,20 @@ from meshmode.dof_array import thaw, DOFArray
 from grudge.symbolic.primitives import DOFDesc
 from grudge.eager import interior_trace_pair, cross_rank_trace_pairs
 from grudge.symbolic.primitives import TracePair, as_dofdesc
+
+
+def _sqrt(actx, x):
+    if isinstance(x, DOFArray):
+        return actx.np.sqrt(x)
+    else:
+        return math.sqrt(x)
+
+
+def _grad(discr, x):
+    if isinstance(x, DOFArray):
+        return discr.grad(x)
+    else:
+        return 0.
 
 
 def _q_flux(discr, quad_tag, alpha, u_tpair):
@@ -51,7 +66,7 @@ def _q_flux(discr, quad_tag, alpha, u_tpair):
     normal_quad = thaw(actx, discr.normal(dd_quad))
 
     alpha_quad = discr.project("vol", dd_quad, alpha)
-    sqrt_alpha_quad = actx.np.sqrt(alpha_quad)
+    sqrt_alpha_quad = _sqrt(actx, alpha_quad)
 
     u_avg_quad = discr.project(dd, dd_quad, u_tpair.avg)
 
@@ -69,7 +84,7 @@ def _u_flux(discr, quad_tag, alpha, q_tpair):
     normal_quad = thaw(actx, discr.normal(dd_quad))
 
     alpha_quad = discr.project("vol", dd_quad, alpha)
-    sqrt_alpha_quad = actx.np.sqrt(alpha_quad)
+    sqrt_alpha_quad = _sqrt(actx, alpha_quad)
 
     q_avg_quad = discr.project(dd, dd_quad, q_tpair.avg)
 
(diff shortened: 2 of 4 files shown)


majosm:
I think this is ready for a look.

Currently the tests are taking a fairly long time to run. What do you think? Can we live with it like this for now (I assume it will become faster once we have lazy evaluation, etc.), or should I try to condense the tests more?

mirgecom/diffusion.py
```diff
@@ -207,33 +233,52 @@ def diffusion_operator(discr, alpha, boundaries, u):
...
+    dd_allfaces_quad = DOFDesc("all_faces", var_diff_quad_tag)
+
+    alpha_quad = discr.project("vol", dd_quad, alpha)
+    sqrt_alpha_quad = actx.np.sqrt(alpha_quad)
+    grad_alpha_quad = discr.project("vol", dd_quad, discr.grad(alpha))
+    u_quad = discr.project("vol", dd_quad, u)
+
     q = discr.inverse_mass(
-        -math.sqrt(alpha)*discr.weak_grad(u)
+        discr.mass(dd_quad, -0.5/sqrt_alpha_quad * grad_alpha_quad * u_quad)
```
inducer, line 247: Add comments to explain the chain-rule derivation you're using here.

mirgecom/diffusion.py
```diff
@@ -149,25 +171,26 @@ def __init__(self, value):
...
-        flux_weak = alpha*self.value*bdry_ones
-        return discr.project(dd, "all_faces", flux_weak)
+        flux_quad = discr.project(dd, dd_quad, -alpha_int*self.value)
+        return discr.project(dd_quad, dd_allfaces_quad, flux_quad)
 
 
-def diffusion_operator(discr, alpha, boundaries, u):
+def diffusion_operator(discr, alpha, boundaries, u, var_diff_quad_tag=QTAG_NONE):
     r"""
     Compute the diffusion operator.
```
inducer, line 195: Give the operator (including the diffusivity) as TeX.

test/test_diffusion.py
```diff
@@ -307,10 +361,13 @@ def test_diffusion_compare_to_nodal_dg(actx_factory, problem, order,
             def sym_eval_mirgecom(expr):
                 return sym.EvaluationMapper({"x": nodes_mirgecom, "t": t})(expr)
 
+            alpha_mirgecom = discr_mirgecom.zeros(actx) + 1.
```
inducer, line 364: Maybe use `_num` instead of `_mirgecom` as a suffix here?

inducer (commented):

> Currently the tests are taking a fairly long time to run. What do you think? Can we live with it like this for now (I assume it will become faster once we have lazy evaluation, etc.), or should I try to condense the tests more?

IIUC, this goes from 17m to 28m for CI. I think that's OK, and I agree that lazy should help. IMO, we can live with this for now.

majosm:
> IIUC, this goes from 17m to 28m for CI.

Yeah, and also for MacOS it goes from 32m to 53m.

mirgecom/diffusion.py
```diff
@@ -31,26 +31,48 @@
...
-def _u_flux(discr, alpha, q_tpair):
-    normal = thaw(q_tpair.int[0].array_context, discr.normal(q_tpair.dd))
-    flux_weak = math.sqrt(alpha)*np.dot(q_tpair.avg, normal)
-    return discr.project(q_tpair.dd, "all_faces", flux_weak)
+    normal = thaw(actx, discr.normal(dd))
+
+    alpha_quad = discr.project("vol", dd_quad, alpha)
+    sqrt_alpha_quad = actx.np.sqrt(alpha_quad)
+
+    flux_quad = discr.project(dd, dd_quad, -u_tpair.avg*normal)
```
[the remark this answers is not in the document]
majosm: Interesting. How come we don't do that for Euler, etc.? Is that on the to-do list for later?

mirgecom/diffusion.py
```diff
@@ -31,26 +31,48 @@
...
 from grudge.eager import interior_trace_pair, cross_rank_trace_pairs
-from grudge.symbolic.primitives import TracePair
+from grudge.symbolic.primitives import TracePair, as_dofdesc
 
 
-def _q_flux(discr, alpha, u_tpair):
-    normal = thaw(u_tpair.int.array_context, discr.normal(u_tpair.dd))
-    flux_weak = math.sqrt(alpha)*u_tpair.avg*normal
-    return discr.project(u_tpair.dd, "all_faces", flux_weak)
+def _q_flux(discr, var_diff_quad_tag, alpha, u_tpair):
```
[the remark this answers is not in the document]
inducer: It might be a tad less stiff, but I don't know. Maybe a that's a good question for CS meeting tomorrow?

mirgecom/diffusion.py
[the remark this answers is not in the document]
majosm: I think I misunderstood this comment when I read it yesterday. The multiplication by the normal doesn't demand overintegration in general, it's just that since we're already doing it we want to use the more accurate normal from the quadrature discretization. Something like that?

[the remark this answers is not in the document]
inducer: *If* we use curvilinear geometry or quad/hex elements, *then* the normal (as a function on the element) won't be a constant. As a result, the multiplication has the potential to drive an aliasing instability, and overintegration is generically necessary.

Does that explain it better?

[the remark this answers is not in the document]
majosm: I think so... Does that mean that in the curvilinear case the other operators (Euler/wave/whatever) will likely need overintegration for the normal as well?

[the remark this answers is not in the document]
inducer: Yep. *If* we end up doing curvi geometry.

mirgecom/diffusion.py
majosm: It's there in the lines below that (pydocstyle complains when I make the first sentence of the docstring span more than one line).

mirgecom/diffusion.py
```diff
@@ -149,25 +171,26 @@ def __init__(self, value):
...
         # computing sqrt(alpha); _u_flux would need to be modified to accept such
         # values).
-        flux_weak = alpha*self.value*bdry_ones
-        return discr.project(dd, "all_faces", flux_weak)
+        flux_quad = discr.project(dd, dd_quad, -alpha_int*self.value)
+        return discr.project(dd_quad, dd_allfaces_quad, flux_quad)
 
 
-def diffusion_operator(discr, alpha, boundaries, u):
+def diffusion_operator(discr, alpha, boundaries, u, var_diff_quad_tag=QTAG_NONE):
```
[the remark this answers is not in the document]
majosm: What do you think about making the quad tag a required argument (as in b106785)? Probably less error-prone.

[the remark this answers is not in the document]
inducer: I agree that it's probably better to require it.

majosm:
Ready for review when you have a chance @inducer.

mirgecom/diffusion.py
inducer: Ah, missed it!

mirgecom/diffusion.py
```diff
@@ -199,41 +233,67 @@ def diffusion_operator(discr, alpha, boundaries, u):
...
 
+    actx = u.array_context
+
+    dd_quad = DOFDesc("vol", quad_tag)
+    dd_allfaces_quad = DOFDesc("all_faces", quad_tag)
+
+    if isinstance(alpha, DOFArray):
+        alpha_array = alpha
+    else:
+        alpha_array = discr.zeros(actx) + alpha
```
inducer, line 252: How come that's necessary?

test/test_diffusion.py
```diff
@@ -309,7 +363,8 @@ def sym_eval_mirgecom(expr):
 
             u_mirgecom = sym_eval_mirgecom(p.sym_u)
 
-            diffusion_u_mirgecom = diffusion_operator(discr_mirgecom, alpha=p.alpha,
+            diffusion_u_mirgecom = diffusion_operator(discr_mirgecom,
```
inducer, line 366: Still a bit unhappy with those `_mirgecom` suffixes. `_num`? Something else?

inducer (approved):
Thanks! The suffixes thing is the only thing that's left, and that's nonessential from my view. LGTM otherwise.

test/test_diffusion.py
majosm: The `_mirgecom` suffix is to distinguish from the nodal-dg equivalents (`_ndg` suffix). Both are numerical, so I don't know if I like `_num`... I'm having trouble thinking of something else, apart from maybe just omitting the `_mirgecom` suffix?

mirgecom/diffusion.py
majosm: I added that to allow specifying e.g. `alpha=2.` for the uniform case on the user side. But in retrospect I'm less sure that this is a good idea. Maybe we can talk about this today.

test/test_diffusion.py
inducer: Ah, duh. Makes sense. I take back what I said.

mirgecom/diffusion.py
inducer: As discussed, maybe rewire `project` to be able to deal with constants.

majosm: I was able to add the constant case to `project` easily enough, but to get things working I needed to add special handling for a couple of other operations too (see code [here](https://github.com/majosm/mirgecom/commit/a493803d8e48b8074980c71d48e682e9c99a2ee0#diff-8c80087b48245eecc882a14fa78e26b0891982bff838e184fcc80bbb396b7465R45-R56)). I'm on the fence about this... What do you think?

inducer: I actually like these modifications just fine.

majosm:
@inducer Good to go?

inducer (commented):
Argh, sorry, wasn't done. Here are two pending comments I head.

examples/heat-source-mpi.py
```diff
@@ -97,17 +98,20 @@ def main():
 
     nodes = thaw(actx, discr.nodes())
 
-    u = discr.zeros(actx)
-
-    vis = make_visualizer(discr, order+3 if dim == 2 else order)
+    alpha = discr.zeros(actx) + 1.
```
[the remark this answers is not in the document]
inducer: #209

mirgecom/diffusion.py
```diff
@@ -149,42 +188,52 @@ def __init__(self, value):
...
         # computing sqrt(alpha); _u_flux would need to be modified to accept such
         # values).
-        flux_weak = alpha*self.value*bdry_ones
-        return discr.project(dd, "all_faces", flux_weak)
+        alpha_int_quad = discr.project("vol", dd_quad, alpha)
+        if isinstance(self.value, DOFArray):
+            value_quad = discr.project(dd, dd_quad, self.value)
+            flux_quad = -alpha_int_quad*value_quad
+        else:
+            flux_quad = -alpha_int_quad*self.value
```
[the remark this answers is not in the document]
inducer: #209

PR #181, opened 2020-12-22
add example git commit hook to run flake8/pydocstyle

matthiasdiener:
This doesn't run pylint cause that takes a long time.


inducer (approved):
Thx!

PR #182, opened 2021-01-04
Set autodoc_typehints = "description" in doc/conf

inducer:


inducer:
Personally, I like type hints in the signature. But with napoleon, I think the consistent spot for the type annotations is in the description. And, again IMO, to avoid redundancy, we should standardize towards type annotations and away from specifying types via the numpy docstring convention.

lukeolson:
Possible issues with the returns: https://github.com/sphinx-doc/sphinx/issues/7077

lukeolson:
`autodoc_typehints` does pick up the types even with numpy docstrings (and napoleon).  So, let try this.

PR #183, opened 2021-01-05
add advection example

matthiasdiener:
based on wave-eager/grudge advection example.

### DIFF
--- a/examples/advection.py
+++ b/examples/advection.py
@@ -96,7 +96,11 @@ def main(use_profiling=False):
         flux_type=flux_type)
 
     bound_op = bind(discr, op.sym_operator())
-    u = bind(discr, u_analytic(sym.nodes(dim)))(actx, t=0)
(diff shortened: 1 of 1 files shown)


examples/advection.py
```diff
@@ -0,0 +1,126 @@
...
+        t = sym.var("t", sym.DD_SCALAR)
+        return f(-np.dot(c, x) / norm_c + t * norm_c)
+
+    from grudge.models.advection import WeakAdvectionOperator
+    op = WeakAdvectionOperator(c,
+        inflow_u=u_analytic(sym.nodes(dim, sym.BTAG_ALL)),
+        flux_type=flux_type)
+
+    bound_op = bind(discr, op.sym_operator())
+    u = bind(discr, u_analytic(sym.nodes(dim)))(actx, t=0)
```
inducer, line 99: This interface to grudge is going away. Advection should be expressed using eager operators, like everything else in mirgecom.

PR #140, opened 2020-11-06
Prototype a driver with restart

inducer:
Towards #139 

cc @MTCam

PR #188, opened 2021-01-08
Add restart module

inducer:
This is a version of #140 that uses https://github.com/inducer/meshmode/pull/99 to remove  a little bit of array context gymnastics from restart.

The diff vs that branch is here: https://github.com/illinois-ceesd/mirgecom/compare/restart-prototype...restart-prototype-v2

cc @MTCam @anderson2981


MTCam:
We need to port this to a more relevant example in *MIRGE-Com*, and move it away from the old crusty EagerDiscretization. I think we can just roll in an example that looks like @anderson2981's nozzle or so and kill multiple birds with a single projectile.

MTCam:
This is ready for merge as soon as you're a comfy @inducer.

inducer:
LGTM. Can't approve, since it's (nominally) my PR, apparently.

MTCam:
Sorry, just wanted to update the wave-eager-mpi restart example since it demonstrates `restart_order != saved_order`.  I'll make sure it works and submit a review on it.

MTCam (approved):
Yah, itworks.

MTCam:
Okie for merge @inducer?

mirgecom/restart.py
```diff
@@ -0,0 +1,59 @@
...
+    """Pickle the simulation data into a file for use in restarting."""
+    rank = 0
+    if comm:
+        rank = comm.Get_rank()
+    if rank == 0:
+        import os
+        rst_dir = os.path.dirname(filename)
+        if rst_dir:
+            if not os.path.exists(rst_dir):
+                os.makedirs(rst_dir)
```
inducer, line 54:
```suggestion
            os.makedirs(rst_dir)
```
(The `if` is redundant.)

inducer (commented):
LGTM. One minor suggestion.

mirgecom/restart.py
matthiasdiener: Should it be `os.makedirs(rst_dir, exist_ok=True)` ?

PR #96, opened 2020-09-11
Time series logging

matthiasdiener:
Implements https://github.com/illinois-ceesd/planning/issues/57

PR #192, opened 2021-01-18
profiling: add pyopencl.Array elementwise kernels

matthiasdiener:
Also adds totals for num_calls and time.

Mostly extracted from #96.


matthiasdiener:
Ready for review. I think this is pretty important, because people that run profiling from the master branch will get an incomplete profile without this PR.

inducer (approved):
LGTM. Thanks for working on this!

PR #194, opened 2021-01-19
batch script fixes

### LINKED ISSUE #195: Document that performance runs should be done with python -O
x-ref: https://github.com/inducer/loopy/pull/196

cc @matthiasdiener

matthiasdiener:
Lassen+Quartz:
- Print nnodes/nproc to aid in reproducing runs
- add -O option to Python command line (Fixes #195)

Lassen:
- Use jsrun instead of lrun
- Use GPU by default
- Print task allocation to aid in checking that it is doing what we expect

### DIFF
--- a/examples/lassen.bsub.sh
+++ b/examples/lassen.bsub.sh
@@ -33,4 +33,6 @@ $jsrun_cmd js_task_info
 
 echo "----------------------------"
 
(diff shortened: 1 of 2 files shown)


inducer (commented):
Thanks!

inducer (approved):

PR #196, opened 2021-01-19
Use empty array instead of None when scalars aren't present

majosm:


mirgecom/euler.py
```diff
@@ -164,43 +164,33 @@ def split_conserved(dim, q):
     """Get the canonical conserved quantities.
 
     Return a :class:`ConservedVars` that is the canonical conserved quantities,
-    mass, energy, and momentum from the agglomerated object array extracted
-    from the state vector *q*. For single component gases, i.e. for those state
-    vectors *q* that do not contain multispecies mixtures, the returned
-    dataclass :attr:`ConservedVars.species_mass` will be set to *None*..
+    mass, energy, momentum, and any scalars, from the agglomerated object array
+    extracted from the state vector *q*.
```
MTCam: mass, energy, momentum, and any scalars, from the agglomerated object array
    extracted from the state vector *q*. For single component gases, i.e. for 
    those state vectors *q* that do not contain multi-species mixtures, the 
    returned dataclass :attr:`ConservedVars.species_mass` will be set to an empty
    array.

MTCam: We should retain the explanation of the handling of non-mixtures. AK asked for this explicitly.

mirgecom/initializers.py
```diff
@@ -616,22 +609,14 @@ def exact_rhs(self, discr, q, t=0.0):
...
-                                     for i in range(self._nspecies)])
-            r2s = make_obj_array([np.dot(rel_poss[i], rel_poss[i])
-                                  for i in range(self._nspecies)])
-            expterms = make_obj_array(
-                [self._spec_amplitudes[i] * actx.np.exp(- r2s[i])
-                 for i in range(self._nspecies)]
-            )
-            specrhs = make_obj_array([2 * expterms[i] * rdotvs[i]
-                                      for i in range(self._nspecies)])
+        # @MikeC: Guess I need to insert rho0 here somewhere after adding it above?
```
MTCam, line 612: Yes, it should just be rho0 * (original).  How does the test pass without that?

MTCam (changes requested):
Just had a couple of comments.   All lgtm, and thanks very much for tackling this.

mirgecom/euler.py
majosm: Added back.

mirgecom/initializers.py
majosm: The test only looks at `__call__` I think, not `exact_rhs`. (And the original test used `rho0 = 1`; I changed that in this PR.)

MTCam: There is a test down in `test_euler.py` (at least there should be) that should be checking the RHS. The test (as written before your latest change to put rho0 in) should have cause that one to fail if rho0 was not present (i.e. and not = 1.0).  rho0 should be specified by user, not hard-coded.

MTCam: I see now, the test is just letting rho0 = 1.0.  Should have let it be != 1.0 so that it would test it better.

MTCam:
looks great to me @majosm, merge at will :)

MTCam (approved):

PR #200, opened 2021-01-21
CI fix

matthiasdiener:


matthiasdiener:
Well, that seems to have fixed it, but it's not clear to me how that ever worked at all... The only relevant change that I could find is that GitHub updated its Ubuntu 18.04 CI image 24 hours ago (to https://github.com/actions/virtual-environments/blob/ubuntu18/20201210.0/images/linux/Ubuntu1804-README.md), seemingly coinciding with the start of the failures.

From what I could see, `pip install git+...` seems to check against the version already installed, and will skip building the package when the exact same version is already install (through conda in our case). So for the current situation with pyopencl v2020.3.1 installed, 

- `pip install "git+https://github.com/inducer/pyopencl.git#egg=pyopencl"` will try to build pyopencl (as the master branch is newer than v2020.3.1)
- `pip install "git+https://github.com/inducer/pyopencl.git@v2020.3.1#egg=pyopencl"` will **not** try to build pyopencl
- `pip install "git+https://github.com/inducer/pyopencl.git@v2020.3#egg=pyopencl"` will try to build pyopencl

🤷‍♂️

inducer:
Huh, OK. I guess that sort of makes sense. Thanks for figuring this out!

inducer (approved):

PR #203, opened 2021-01-26
check that we are not running multiple ranks on the same GPU

### LINKED ISSUE #190: Abort if multiple ranks are using the same GPU
cc @matthiasdiener 

Context: Lrun bug found by @matthiasdiener

matthiasdiener:
Fixes #190.


matthiasdiener:
WIP - looking for feedback.

mirgecom/mpi.py
```diff
@@ -78,6 +78,42 @@ def wrapped_func(*args, **kwargs):
...
+            dev = cl_ctx.get_info(cl.context_info.DEVICES)[0]
+            platform = dev.get_info(cl.device_info.PLATFORM)
+            platform_name = platform.get_info(cl.platform_info.NAME)
+
+            if platform_name == "Portable Computing Language":
+                try:
+                    domain_id = dev.get_info(0x4010)
+                except cl._cl.LogicError:
+                    from warnings import warn
+                    warn("Can not detect whether multiple ranks are running on the"
```
inducer, line 96:
```suggestion
                    warn("Cannot detect whether multiple ranks are running on the"
```

mirgecom/mpi.py
```diff
@@ -78,6 +78,42 @@ def wrapped_func(*args, **kwargs):
...
+                         " same GPU because you need at least pocl version 1.7.")
```
inducer, line 97: As version advice goes, this is probably misleading: We'll likely patch this into our version of 1.6 on conda forge as soon as it's merged upstream. Maybe just link to the PR?

mirgecom/mpi.py
```diff
@@ -78,6 +78,42 @@ def wrapped_func(*args, **kwargs):
...
+        if size > 1:
+            node_comm = MPI.COMM_WORLD.Split_type(MPI.COMM_TYPE_SHARED)
+            node_rank = node_comm.Get_rank()
+            cl_ctx = cl.create_some_context()
+            dev = cl_ctx.get_info(cl.context_info.DEVICES)[0]
+            platform = dev.get_info(cl.device_info.PLATFORM)
+            platform_name = platform.get_info(cl.platform_info.NAME)
+
+            if platform_name == "Portable Computing Language":
+                try:
```
inducer, line 92: Probably no need to try this on devices that don't satisfy `bool(dev.type & cl.device_type.GPU)`.

mirgecom/mpi.py
```diff
@@ -78,6 +78,42 @@ def wrapped_func(*args, **kwargs):
                  "ranks. See https://mirgecom.readthedocs.io/en/latest/running.html#running-with-large-numbers-of-ranks-and-nodes"  # noqa: E501
                  " for more information.")
 
+        # Check whether multiple ranks are running on the same GPU on each node.
+        # Only works with pocl-cuda devices currently.
+        if size > 1:
+            node_comm = MPI.COMM_WORLD.Split_type(MPI.COMM_TYPE_SHARED)
+            node_rank = node_comm.Get_rank()
+            cl_ctx = cl.create_some_context()
+            dev = cl_ctx.get_info(cl.context_info.DEVICES)[0]
```
inducer, line 87: Should better handle multi-device contexts. (Maybe by just raising an error?)

mirgecom/mpi.py
```diff
@@ -78,6 +78,42 @@ def wrapped_func(*args, **kwargs):
...
+                    domain_id = dev.get_info(0x4010)
+                except cl._cl.LogicError:
+                    from warnings import warn
+                    warn("Can not detect whether multiple ranks are running on the"
+                         " same GPU because you need at least pocl version 1.7.")
+                else:
+                    bus_id = dev.get_info(cl.device_info.PCI_BUS_ID_NV)
+                    slot_id = dev.get_info(cl.device_info.PCI_SLOT_ID_NV)
+
+                    dev_id = hash(tuple((domain_id, bus_id, slot_id)))
```
inducer, line 102:
```suggestion
                    dev_id = (domain_id, bus_id, slot_id)
```

mirgecom/mpi.py
```diff
@@ -78,6 +78,42 @@ def wrapped_func(*args, **kwargs):
...
+
+                    if node_rank == 0:
+                        if len(dev_ids) != len(set(dev_ids)):
+                            print(dev_ids)
+                            raise RuntimeError(
+                                  "Multiple ranks are running on the same GPU.")
+
+            else:
+                from warnings import warn
+                warn("Can not detect whether multiple ranks are running on the "
```
inducer, line 114:
```suggestion
                warn("Cannot detect whether multiple ranks are running on the "
```

mirgecom/mpi.py
```diff
@@ -78,6 +78,42 @@ def wrapped_func(*args, **kwargs):
...
+                     f" same GPU on platform '{platform_name}'.")
```
inducer, line 115: No point in issuing this warning unless we're talking to GPUs.

mirgecom/mpi.py
matthiasdiener: Fixed.

mirgecom/mpi.py
matthiasdiener: Fixed.

mirgecom/mpi.py
matthiasdiener: Fixed.

mirgecom/mpi.py
matthiasdiener: I implemented something, but not sure if that is what you had in mind.

mirgecom/mpi.py
```diff
@@ -32,6 +32,66 @@
...
+            slot_id = dev.get_info(cl.device_info.PCI_SLOT_ID_NV)
+
+            dev_id = (domain_id, bus_id, slot_id)
+
+            dev_ids = node_comm.gather(dev_id, root=0)
+
+            if node_rank == 0:
+                if len(dev_ids) != len(set(dev_ids)):
+                    raise RuntimeError(
+                          "Multiple ranks are running on the same GPU.")
```
inducer, line 87: Print an example of which?

mirgecom/mpi.py
```diff
@@ -32,6 +32,66 @@
...
+    dev = dev[0]
+    platform = dev.get_info(cl.device_info.PLATFORM)
+    platform_name = platform.get_info(cl.platform_info.NAME)
+
+    if not (dev.type & cl.device_type.GPU):
+        return
+
+    if platform_name == "Portable Computing Language":
+        try:
+            domain_id = dev.get_info(cl.device_info.PCI_DOMAIN_ID_NV)
```
inducer, line 70:
```suggestion
            domain_id = dev.pci_domain_id_nv
```
(same below)

mirgecom/mpi.py
```diff
@@ -32,6 +32,66 @@
...
+        except cl._cl.LogicError:
+            from warnings import warn
+            warn("Cannot detect whether multiple ranks are running on the"
+                 " same GPU because you need a pocl version with"
+                 " https://github.com/pocl/pocl/pull/910.")
```
inducer, line 75: Give a version number indication.

mirgecom/mpi.py
```diff
@@ -32,6 +32,66 @@
...
+    platform = dev.get_info(cl.device_info.PLATFORM)
+    platform_name = platform.get_info(cl.platform_info.NAME)
+
+    if not (dev.type & cl.device_type.GPU):
+        return
+
+    if platform_name == "Portable Computing Language":
+        try:
+            domain_id = dev.get_info(cl.device_info.PCI_DOMAIN_ID_NV)
+        except cl._cl.LogicError:
```
inducer, line 71: Wouldn't you get an `AttributeError` first, for old versions of pyopencl?

mirgecom/mpi.py
```diff
@@ -32,6 +32,66 @@
...
+
+    node_comm = MPI.COMM_WORLD.Split_type(MPI.COMM_TYPE_SHARED)
+    node_rank = node_comm.Get_rank()
+
+    cl_ctx = cl.create_some_context()
+    dev = cl_ctx.get_info(cl.context_info.DEVICES)
+
+    # No support for multi-device contexts
+    if len(dev) > 1:
+        raise RuntimeError(f"Multiple devices selected for rank {rank}.")
```
inducer, line 59: Why? Shouldn't they be unique nonetheless?

mirgecom/mpi.py
```diff
@@ -32,6 +32,66 @@
 import sys
 
 
+def _check_gpu_oversubscription():
+    """
+    Check whether multiple ranks are running on the same GPU on each node.
+
+    Only works with pocl-cuda devices currently due to the use of the
+    PCI_DOMAIN_ID_NV extension.
+    """
```
inducer, line 41: I could see maintaining this in PyOpenCL, too. What do you think of that?

mirgecom/mpi.py
```diff
@@ -32,6 +32,66 @@
...
+    from mpi4py import MPI
+    import pyopencl as cl
+
+    size = MPI.COMM_WORLD.Get_size()
+    rank = MPI.COMM_WORLD.Get_rank()
+
+    if size <= 1:
+        return
+
+    node_comm = MPI.COMM_WORLD.Split_type(MPI.COMM_TYPE_SHARED)
```
inducer, line 51: When does this communicator get destroyed?

inducer (commented):
Thanks!

matthiasdiener:
Tested on Lassen, ready for another review.

mirgecom/mpi.py
inducer: Fair enough. I mean, handling multi-device contexts here isn't hard (just add all their IDs before gathering), but this is OK. If you do leave it, please change the error message to `NotImplementedError("multi-device contexts not yet supported")`.

mirgecom/mpi.py
```diff
@@ -32,6 +32,64 @@
...
+        if node_rank == 0:
+            if len(dev_ids) != len(set(dev_ids)):
+                hostname = MPI.Get_processor_name()
+                dup = [item for item in dev_ids if dev_ids.count(item) > 1]
+
+                raise RuntimeError(
+                      f"Multiple ranks are sharing GPUs on node '{hostname}'."
+                      f" Duplicate PCIe IDs: {dup}.")
+
+    node_comm.Free()
```
inducer, line 90: This alloc/free pair should be a context manager.

mirgecom/mpi.py
```diff
@@ -32,6 +32,64 @@
...
+    node_rank = node_comm.Get_rank()
+
+    try:
+        domain_id = hex(dev.pci_domain_id_nv)
+    except (cl._cl.LogicError, AttributeError):
+        from warnings import warn
+        warn("Cannot detect whether multiple ranks are running on the"
+             " same GPU because it requires Nvidia GPUs running with"
+             " pyopencl>2021.1.1 and (Nvidia CL or pocl>1.6).")
+    else:
```
inducer, line 73: Just return from the handler and avoid the `else`. (requires the context mgr to be correct)

mirgecom/mpi.py
```diff
@@ -32,6 +32,64 @@
...
+    node_comm = MPI.COMM_WORLD.Split_type(MPI.COMM_TYPE_SHARED)
+    node_rank = node_comm.Get_rank()
+
+    try:
+        domain_id = hex(dev.pci_domain_id_nv)
+    except (cl._cl.LogicError, AttributeError):
+        from warnings import warn
+        warn("Cannot detect whether multiple ranks are running on the"
+             " same GPU because it requires Nvidia GPUs running with"
+             " pyopencl>2021.1.1 and (Nvidia CL or pocl>1.6).")
```
inducer, line 72: It will say this for AMD and Intel GPUs now. Is that intended?

matthiasdiener: Do you mean this warning should not be shown for AMD/Intel GPUs? Or is the text unclear?

mirgecom/mpi.py
matthiasdiener: Changed the error message.

mirgecom/mpi.py
inducer: OK. Nvm for now.

mirgecom/mpi.py
matthiasdiener: I didn't find much info regarding mpi4py context managers so I implemented a simple one.

mirgecom/mpi.py
matthiasdiener: Fixed.

matthiasdiener:
Ready for another review

mirgecom/mpi.py
inducer: Fair enough.

inducer (approved):

inducer:
Thanks!

PR #204, opened 2021-01-28
Use **kwags in boundary_pair method

inducer:


MTCam:
Added a tiny doc blurb about the requirements on the user func. Hope this is OK.

MTCam (approved):
lgtm.  Thanks very much for your help.

PR #206, opened 2021-01-28
Doc for flux shapes

MTCam:
This is an attempt to add some documentation for the expected shape for the flux data.  As fluxes are typically internal-only data, folks adding new capabilities (e.g. new numerical flux routines) will need to know the expected shape of the arrays.

It is easy to run afoul of the expectation.  Here's an example of that from @w-hagen:
```
vol_flux = inviscid_flux(discr, eos, q)
    print(vol_flux.shape)
    print(vol_flux[0].shape)
```
> returns (4,2), (2,) not (4,), (2,) like my arrays do

### DIFF
--- a/mirgecom/euler.py
+++ b/mirgecom/euler.py
@@ -199,12 +199,12 @@ def inviscid_flux(discr, eos, q):
     $(\rho\vec{V},(\rho{E}+p)\vec{V},\rho(\vec{V}\otimes\vec{V})
     +p\mathbf{I}, \rho{Y_s}\vec{V})$
 
-    The fluxes are returned as 2D object array with shape:
-    np.ndarray.shape = (num_equations, ndim).  Each entry in the
+    The fluxes are returned as a 2D object array with shape:
+    ``(num_equations, ndim)``.  Each entry in the
     flux array is a :class:`~meshmode.dof_array.DOFArray`.  This
     form and shape for the flux data is required by the built-in
     state data handling mechanism in :mod:`mirgecom.euler`. That
-    mechanism includes at least
+    mechanism is used by at least
     :class:`mirgecom.euler.ConservedVars`, and
     :func:`mirgecom.euler.join_conserved`, and
     :func:`mirgecom.euler.split_conserved`.
--- a/mirgecom/euler.py
+++ b/mirgecom/euler.py
@@ -202,8 +202,12 @@ def inviscid_flux(discr, eos, q):
     The fluxes are returned as 2D object array with shape:
     np.ndarray.shape = (num_equations, ndim).  Each entry in the
     flux array is a :class:`~meshmode.dof_array.DOFArray`.  This
-    form and shape for the flux data is required by
-    :mod:`mirgecom.euler`.
+    form and shape for the flux data is required by the built-in
+    state data handling mechanism in :mod:`mirgecom.euler`. That
+    mechanism includes the dataclass
+    :class:`mirgecom.euler.ConservedVars`, and module methods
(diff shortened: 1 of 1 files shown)


mirgecom/euler.py
```diff
@@ -198,6 +198,12 @@ def inviscid_flux(discr, eos, q):
     The inviscid fluxes are
     $(\rho\vec{V},(\rho{E}+p)\vec{V},\rho(\vec{V}\otimes\vec{V})
     +p\mathbf{I}, \rho{Y_s}\vec{V})$
+
+    The fluxes are returned as 2D object array with shape:
+    np.ndarray.shape = (num_equations, ndim).  Each entry in the
+    flux array is a :class:`~meshmode.dof_array.DOFArray`.  This
+    form and shape for the flux data is required by
+    :mod:`mirgecom.euler`.
```
inducer, line 206: There's an opportunity to be more precise here: by *what* in `mirgecom.euler`?

MTCam: Hrm. The discussion link didn't quite worked like i hoped (92cea52)

mirgecom/euler.py
```diff
@@ -198,6 +198,16 @@ def inviscid_flux(discr, eos, q):
...
+
+    The fluxes are returned as 2D object array with shape:
+    np.ndarray.shape = (num_equations, ndim).  Each entry in the
+    flux array is a :class:`~meshmode.dof_array.DOFArray`.  This
+    form and shape for the flux data is required by the built-in
+    state data handling mechanism in :mod:`mirgecom.euler`. That
+    mechanism includes the dataclass
+    :class:`mirgecom.euler.ConservedVars`, and module methods
+    :meth:`mirgecom.euler.join_conserved`, and
+    :meth:`mirgecom.euler.split_conserved`.
```
inducer, line 210:
```suggestion
    mechanism includes at least
    :class:`mirgecom.euler.ConservedVars`, and
    :func:`mirgecom.euler.join_conserved`, and
    :func:`mirgecom.euler.split_conserved`.
```

inducer: (The docs will tell you what you're dealing with, by clicking if nothing else.)

mirgecom/euler.py
```diff
@@ -198,6 +198,16 @@ def inviscid_flux(discr, eos, q):
     The inviscid fluxes are
     $(\rho\vec{V},(\rho{E}+p)\vec{V},\rho(\vec{V}\otimes\vec{V})
     +p\mathbf{I}, \rho{Y_s}\vec{V})$
+
+    The fluxes are returned as 2D object array with shape:
```
inducer, line 202:
```suggestion
    The fluxes are returned as a 2D object array with shape:
```

mirgecom/euler.py
```diff
@@ -198,6 +198,16 @@ def inviscid_flux(discr, eos, q):
...
+    np.ndarray.shape = (num_equations, ndim).  Each entry in the
```
inducer, line 203:
```suggestion
    ``(num_equations, ndim)``.  Each entry in the
```

mirgecom/euler.py
```diff
@@ -198,6 +198,16 @@ def inviscid_flux(discr, eos, q):
...
+    flux array is a :class:`~meshmode.dof_array.DOFArray`.  This
+    form and shape for the flux data is required by the built-in
+    state data handling mechanism in :mod:`mirgecom.euler`. That
+    mechanism includes at least
```
inducer, line 207:
```suggestion
    mechanism is used by at least
```

inducer (approved):

PR #211, opened 2021-02-03
Separate testing suites in CI

MTCam:
Separates CI pytest section into suites to work around memory intensive tests causing pytest to crash.


MTCam:
Fixes #212

MTCam:
Could be a bit better with something like this:
```
for dirname in */ ; do
python -m pytest --durations=10 --tb=native --junitxml=pytest.xml --doctest-modules -rxsw $(dirname).
done
```
And maybe a line or two to print the suite(directory) name at test time, and one to pick up any *.py files in the top-level test directory.

.github/workflows/ci.yaml
```diff
@@ -88,7 +88,9 @@ jobs:
             if [[ ${{ matrix.os }} == "macos-latest" ]]; then
               export PYOPENCL_TEST=port:pthread
             fi
-            python -m pytest --durations=10 --tb=native --junitxml=pytest.xml --doctest-modules -rxsw .
+            python -m pytest --durations=10 --tb=native --junitxml=pytest.xml --doctest-modules -rxsw ./euler
+            python -m pytest --durations=10 --tb=native --junitxml=pytest.xml --doctest-modules -rxsw ./wave
+            python -m pytest --durations=10 --tb=native --junitxml=pytest.xml --doctest-modules -rxsw ./diffusion
```
inducer, line 93: Make this a for loop so as to not repeat the command.

inducer, line 93: Won't moving the tests blow up the size of diffs in downstream PRs? (and/or cause merge conflicts?) I think we can just run each file separately for now, as a band-aid.

inducer, line 93: Add a comment linking back to the issue to explain why this is being done.

MTCam: (65be23a)

MTCam: Couple of commits for this, but finally i think is ok now.

MTCam: This will make @majosm happy, too.  FWIW: I like the subdirectories better because it provides an easily discernible separator between the tests associated with each module and/or capability.

MTCam: 1052c59

MTCam: 1052c59

MTCam (commented):
Not sure why it started me reviewing. I must have hit the wrong button.

.github/workflows/ci.yaml
```diff
@@ -88,7 +88,11 @@ jobs:
             if [[ ${{ matrix.os }} == "macos-latest" ]]; then
               export PYOPENCL_TEST=port:pthread
             fi
-            python -m pytest --durations=10 --tb=native --junitxml=pytest.xml --doctest-modules -rxsw .
+            # One test at a time to address ballooning memory.  See MIRGE-Com issue:
```
inducer, line 91:
```suggestion
            # One test at a time to address ballooning memory.
```
(Super minor: That link describes itself, so no need to state what it is.)

inducer (approved):

.github/workflows/ci.yaml
MTCam: Heh, I keep doing that.

PR #213, opened 2021-02-03
Artificial viscosity

w-hagen:
This branch implements a modal indicator based artificial viscosity approach to enable shock capturing.

### Outstanding work needed before moving forward:
- [x] Clean up boundary condition routines utilized for the artificial viscosity
- [x] Expose smoothness indicator parameters
- [x] Remove unnecessary make_obj_array calls
- [x] Bring branch up to master
- [x] Add documentation of approach
- [x] Add tests for additions


mirgecom/initializers.py
```diff
@@ -262,6 +263,95 @@ def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
...
+        rhoe = (rhoel/2.0*(actx.np.tanh(-xtanh)+1.0)
+                + rhoer/2.0*(actx.np.tanh(xtanh)+1.0))
+        u = ul/2.0*(actx.np.tanh(-xtanh)+1.0)+ur/2.0*(actx.np.tanh(xtanh)+1.0)
+        v = vl/2.0*(actx.np.tanh(-xtanh)+1.0)+vr/2.0*(actx.np.tanh(xtanh)+1.0)
+        rhou = mass*u
+        rhov = mass*v
+        energy = rhoe + 0.5*mass*(u*u + v*v)
+
+        return join_conserved(dim=self._dim, mass=mass, energy=energy,
+                              momentum=make_obj_array([rhou, rhov]))
```
MTCam, line 352:
```suggestion
        vel = make_obj_array([u, v])
        mom = mass * vel
        energy = rhoe + .5*mass*np.dot(vel, vel)

        return join_conserved(dim=self._dim, mass=mass, energy=energy, momentum=mom)
```

What about something like this?

mirgecom/initializers.py
```diff
@@ -262,6 +263,95 @@ def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
...
+        """Initialize initial condition options.
+
+        Parameters
+        ----------
+        dim: int
+           dimension of domain, must be 2
+        x0: float
+           location of shock
+        us: float
+           shock speed
```
MTCam, line 295: It might be nice to rename `x0` -> `shock_location`, `us` -> `shock_speed`

mirgecom/initializers.py
w-hagen: Yes, that is better.

mirgecom/simutil.py
```diff
@@ -93,8 +93,8 @@ def __init__(self, step, t, state):
 
 def sim_checkpoint(discr, visualizer, eos, q, vizname, exact_soln=None,
                    step=0, t=0, dt=0, cfl=1.0, nstatus=-1, nviz=-1, exittol=1e-16,
-                   constant_cfl=False, comm=None, viz_fields=None, overwrite=False,
-                   vis_timer=None):
+                   s0=None, kappa=None, constant_cfl=False, comm=None,
+                   viz_fields=None, overwrite=False, vis_timer=None):
```
MTCam, line 97: The `viz_fields` mechanism is the way to add user-defined quantities to the visualization dumps. [Here is an example](https://github.com/illinois-ceesd/mirgecom/blob/171e59fe483f11448b4d9967dd10a3512753bbcf/examples/autoignition-mpi.py#L228) of how to use it.

examples/heat-source-mpi.py
```diff
@@ -1,3 +1,5 @@
+"""Demonstrate heat source."""
+
```
MTCam, line 2: I acknowledge that this change is needed to get your linter to shut up, but it is just adding noise to the AV change set.  Not sure how others feel about this, but I'd just back this out and let it be changed by the file author, if they desire.  (cc @majosm).

requirements.txt
```diff
@@ -6,6 +6,7 @@ pyvisfile
 pymetis
 importlib-resources
 psutil
+gmsh
```
MTCam, line 9: As we discussed, this has the potential to break installs where gmsh cannot be installed by `pip`.  One platform where it *definitely* doesn't work at all is *Lassen* - because there are no compatible `gmsh` packages for that platform, and it fails to install using `pip`.   

One platform that is iffy is *Quartz* because as we've noted there it is required to be installed by `conda` and not `pip` in order to get a working install.  

The only platform that seems to be OK with `gmsh` installed via `pip` is the CI VM.

Any ideas on what to do about this?   CC: @inducer, @matthiasdiener

examples/heat-source-mpi.py
majosm: Yeah, I think we only enforce pydocstyle for the `mirgecom/` directory.

test/test_artificial_viscosity.py
```diff
@@ -0,0 +1,200 @@
...
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.parametrize("dim",  [1, 2, 3])
+@pytest.mark.parametrize("order",  [1, 5])
+def test_tag_cells(ctx_factory, dim, order):
+    """Test tag_cells.
+
+    Tests that the cells tagging properly tags cells
+    given a prescirbed solutions.
```
MTCam, line 54:
```suggestion
    given prescribed solutions.
```

mirgecom/boundary.py
```diff
@@ -62,13 +61,22 @@ def __init__(self, userfunc):
...
         boundary_discr = discr.discr_from_dd(btag)
         nodes = thaw(actx, boundary_discr.nodes())
         ext_soln = self._userfunc(nodes, **kwargs)
-        int_soln = discr.project("vol", btag, q)
-        return TracePair(btag, interior=int_soln, exterior=ext_soln)
+        return ext_soln
+
+    def av(self, discr, q, btag, **kwargs):
+        """Get the exterior solution on the boundary."""
+        return discr.project("vol", btag, q)
```
MTCam, line 79: I dig the changes that separate out the `exterior_soln` call, but I can't say the same for the AV-specific, and kind of vaguely-named `av` function. That an AV-specific function needs to go here at all seems (imo) to emit a code smell that is other than groovy.

Could we do this better with something like an "AV boundary wrapper" that inherits existing boundary objects and overloads, or replaces the boundary object methods it needs? e.g. `exterior_soln`.

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,160 @@
...
+
+    normal = thaw(actx, discr.normal(q_tpair.dd))
+
+    flux_out = np.dot(q_tpair.avg, normal)
+
+    return discr.project(q_tpair.dd, "all_faces", flux_out)
+
+
+def artificial_viscosity(discr, t, eos, boundaries, r, alpha, **kwargs):
+    r"""Compute artifical viscosity for the euler equations."""
```
MTCam, line 91: This interface should be documented.  [Here's a similar one](https://github.com/illinois-ceesd/mirgecom/blob/171e59fe483f11448b4d9967dd10a3512753bbcf/mirgecom/euler.py#L279) to give you some ideas about what to write for some of your arguments.

mirgecom/boundary.py
```diff
@@ -135,7 +159,40 @@ def boundary_pair(self, discr, q, btag, **kwargs):
...
+
+        # Grab a unit normal to the boundary
+        normal = thaw(actx, discr.normal(btag))
+
+        # Get the interior soln
+        int_soln = discr.project("vol", btag, q)
+        bndry_q = split_conserved(dim, int_soln)
+
+        # create result array to fill
+        result = np.zeros(2+dim, dtype=object)
```
MTCam, line 181: The biggest problem I have with this is the assumption regarding the necessary size of the data structure.  That this line is needed at all is already suspicious, and that it is a regular *numpy* array filled with zeros is again also odd to me. 

I feel like this whole thing could be replaced by something like (not exact here, I'm sure):

```
        bnd_cv = split_conserved(dim, int_soln)
        bnd_mom_normcomp = make_obj_array(
            [np.outer(normal, np.dot(bnd_cv.momentum, normal))[i]
            for i in range(dim)]
        )
        new_bnd_mom = 2*bnd_mom_normcomp - bnd_cv.momentum
        return join_conserved(dim=dim, mass=-bnd_cv.mass, energy=-bnd_cv.energy,
                              momentum=new_bnd_mom)
```

This would still need to be tweaked for multi-component gases, but at least it would help address some of the concerning things that stick out.

mirgecom/tag_cells.py
```diff
@@ -0,0 +1,159 @@
...
+        "result[iel,idof] = "
+        "sum(k,vec[iel,k]*vec[iel,k]*modes[k])/sum(j,"
+        " vec[iel,j]*vec[iel,j]+1.0e-12/ndiscr_nodes_in)",
+        name="smooth_comp",
+    )
+    return knl
+
+
+def smoothness_indicator(u, discr, kappa=1.0, s0=-6.0):
+    """Calculate the smoothness indicator."""
```
MTCam, line 107: Should have a documented interface.

requirements.txt
w-hagen: I can remove this if needed for now, but without it the doublemach-mpi.py case will fail CI. I added it for now to make sure it would pass CI if it had gmsh. Another option would be to include the mesh as a data file so we don't need gmsh to generate it, but I wasn't sure if we want a mesh file in the repository. Maybe the mesh file could even exist in another data repository that the CI pulls if we don't want it to live in the mirgecom repo.

MTCam (commented):
This looks terrific - seems quite close for merging!   I've left just a few comments.  

Also, is there a good reason that the utilities under `tag_cells.py` shouldn't go right into the `artificial_viscosity` module?

mirgecom/boundary.py
w-hagen: I don't think join conserved here will work for how I currently have the arrays structured (unless it has relatively recently changed). There is definitely a discussion here to have about if I should be storing my data in a different structure. Let me try this and confirm it won't work first though.

requirements.txt
MTCam: I think you should leave it until we can discuss how we should handle it. It is likely there will be something that can be done to get CI to test it as well as fix installs.

w-hagen:
No actually, I just started with it separated in how I thought about the routines. They probably do make sense to live in one file as opposed to 2.

mirgecom/boundary.py
MTCam: If it fails, it probably makes sense to track down why, and resolve that.  We should try to avoid creating and packing *numpy* arrays as it will likely result somewhere in an `object array <--> device` type issue.

mirgecom/boundary.py
w-hagen: Yeah, so av is objectively a pretty terrible function name that I should have changed to something actually useful. 
image
image
The `exterior_soln` call is then used to set the boundary conditions on Q in the calculation of R, and the terribly named `av` to set the conditions on R when calculating the RHS contribution that is made by the artificial viscosity. It's not clear to me if the conditions on R and Q are necessarily going to be the same. Particularly if more complex boundary conditions were to exist (characteristic, moving wall, ). It still might be possible to roll these into something cleaner though, I will think about it a bit and might pick your brain some later.

mirgecom/initializers.py
w-hagen: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/40364ac41ce999e79573e5980e6f671e8d20851a renames the variables as suggested.

mirgecom/simutil.py
w-hagen: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/77b02639eb730b820de6dd70f405f58451b2e5fb removes changes to simutil.py and presents how to use vis_fields in doublemach-mpi.py

examples/heat-source-mpi.py
w-hagen: Removed: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/8a50f22a38d468b311358214e4791b4cab9f47b5

mirgecom/artificial_viscosity.py
w-hagen: Added interface parameter documentation: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/218cb523e501a27c1aabb6ec56899efb9975a816

mirgecom/tag_cells.py
w-hagen: Added interface parameter documentation: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/218cb523e501a27c1aabb6ec56899efb9975a816

mirgecom/boundary.py
w-hagen: So, the suggested solution doesn't work. However, this does need to get fixed somehow so I can use join_conserved when using species arrays as well I suspect.

The root of the issue I believe to be is (this is me thinking out loud)

https://github.com/w-hagen/mirgecom/blob/cb8be6a102cb0b7089e724324761b97546480db9/mirgecom/artificial_viscosity.py#L131

discr.weak_grad() expects a DOFArray and returns and object array of size dim. Then when obj_array_vectorize is called with it, the result is a nested set of object arrays--an object array of size nvars, each containing an object array of size dim of DOFArrays. This structure is then not compatible with the join_conserved call.

It might make sense then to reshape this set of nested object arrays into a 2D object array? I think that would solve the issues here.

w-hagen: Note to self: Promoting the nested objects to an array of (nvars,ndim) causes issues when calling `grudge.eager.cross_rank_trace_pairs`

w-hagen: An attempt to clean this up can be found here: https://github.com/w-hagen/mirgecom/pull/11

This is also a bit strange with the functions needed to converter from nested objects to a 2d object and back. Maybe there is another way to handle this, but not sure I have any more ideas at the moment.

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,196 @@
...
+
+    actx = q_tpair.int.array_context
+
+    flux_dis = q_tpair.avg
+
+    normal = thaw(actx, discr.normal(q_tpair.dd))
+
+    flux_out = flux_dis * normal
+
+    return discr.project(q_tpair.dd, "all_faces", flux_out)
```
thomasgibson, line 76: Even though this is just an internal function, I am a huge stickler for at least a bit of exposition (documentation), especially for people returning to this code in the future. I think adding a short docstring stating that you are discretizing using central difference numerical fluxes (averaging) would be nice to have here. Similarly below for the numerical flux for the auxiliary equation.

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,196 @@
...
+    numpy.ndarray
+        Agglomerated object array of DOF Arrays representing the RHS associated
+        with the artificial viscosity application.
+    """
+    # Get smoothness indicator
+    indicator = smoothness_indicator(discr, r[0], **kwargs)
+
+    dflux_r = obj_array_vectorize(discr.weak_grad, r)
+
+    def my_facialflux_r_interior(q):
```
thomasgibson, line 133: This is part of the reason why I think using `r` for the state, and `q` for the auxiliary variable is potentially confusing. Please update these :)

thomasgibson (commented):
Some comments mainly on documentation/naming. I'll periodically update this as I see more. Thanks for doing all this work!

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,196 @@
...
+
+To evalutate the second order derivative the problem is recast as a set of first
+ order problems:
+
+.. math::
+
+    \partial_t \mathbf{R} = \nabla\cdot{\mathbf{Q}}
+    \mathbf{Q} = \varepsilon\nabla\mathbf{R}
+
+where $\mathbf{Q}$ is an intermediate variable.
```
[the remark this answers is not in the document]
w-hagen: Yeah, that was actually on my list to fix at one point and slipped of off it. Thanks for pointing this out!

mirgecom/artificial_viscosity.py
w-hagen: Yeah, q here is just a local function variable and I used q for both the r and q flux functions. It would be good to update the local variable to r those that relate to r to make it more readable. Will do

MTCam:
Woops :( I was trying to do a drive-by and fat-fingered it.

mirgecom/boundary.py
MTCam: Of potential interest: inducer/grudge#70

majosm:
Needs a merge @MTCam

mirgecom/artificial_viscosity.py
MTCam: How's this grab you guys? w-hagen/mirgecom#13

thomasgibson (commented):
LGTM, thanks for all the hard work. Just a couple comments

requirements.txt
MTCam: This is resolved.  pip installs gmsh on all the platforms I tried so far including Quartz and Lassen.  Time travel duplicate paradox resolved.

MTCam:
From my perspective, this is basically merge-ready, and definitely ready for another peak from all reviewers.  This does still use the `old way`:tm: for boundary treatment, but unification of boundary treatments is coming with #298.

mirgecom/fluid.py
```diff
@@ -165,6 +166,21 @@ def join_conserved(dim, mass, energy, momentum,
...
+    cv = split_conserved(dim, ary)
+    retval[0] = cv.mass
+    retval[1] = cv.energy
+    for i in range(dim):
+        retval[2+i] = ary[2+i]
+    for i in range(nspecies):
+        retval[dim+2+i] = ary[dim+2+i]
+    return retval
+
+
```
majosm, line 183: Is there a way to write `join_conserved` to handle both of these cases?

mirgecom/initializers.py
```diff
@@ -262,6 +263,100 @@ def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
                               momentum=mom)
 
 
+class DoubleMachReflection:
+    r"""Implement the double shock reflection problem.
+
+        - Woodward and Collela
+
+    The inital condition is defined
```
majosm, line 271:
```suggestion
    The initial condition is defined
```

mirgecom/initializers.py
```diff
@@ -262,6 +263,100 @@ def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
...
+        shock_speed: float
+           shock speed, Mach number
+        """
+        self._shock_location = shock_location
+        self._dim = dim
+        self._shock_speed = shock_speed
+
+    def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
+        """
+        Create the 1D Sod's shock solution at locations *x_vec*.
```
majosm, line 303: Leftover from copying `SodShock1D`?

mirgecom/initializers.py
```diff
@@ -262,6 +263,100 @@ def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
...
+
+        - Woodward and Collela
+
+    The inital condition is defined
+
+    .. math::
+
+        (\rho,u,v,P) =
+
+    This function only serves as an initial condition
```
majosm, line 277: ?

mirgecom/initializers.py
```diff
@@ -262,6 +263,100 @@ def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
...
+
+    def __init__(
+            self, dim=2, shock_location=1.0/6.0, shock_speed=4.0
+    ):
+        """Initialize initial condition options.
+
+        Parameters
+        ----------
+        dim: int
+           dimension of domain, must be 2
```
majosm, line 291: Needed?

mirgecom/initializers.py
```diff
@@ -262,6 +263,100 @@ def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
...
+        self._shock_speed = shock_speed
+
+    def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
+        """
+        Create the 1D Sod's shock solution at locations *x_vec*.
+
+        Parameters
+        ----------
+        t: float
+            Current time at which the solution is desired (unused)
```
majosm, line 308: Looks like it's used.

majosm, line 308: Looks like it's used.

mirgecom/initializers.py
```diff
@@ -262,6 +263,100 @@ def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
...
+        x_vec: numpy.ndarray
+            Nodal coordinates
+        eos: :class:`mirgecom.eos.GasEOS`
+            Equation of state class to be used in construction of soln (if needed)
+        """
+        assert self._dim == 2, "only defined for dim=2"
```
majosm, line 314: Exception instead of assert (if dim is even needed).

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,346 @@
...
+
+
+def smoothness_indicator(discr, u, kappa=1.0, s0=-6.0):
+    """Calculate the smoothness indicator.
+
+    Parameters
+    ----------
+    u
+        A DOF Array of the field that is used to calculate the
+        smoothness indicator.
```
majosm, line 279:
```suggestion
    u: meshmode.dof_array.DOFArray
        The field that is used to calculate the smoothness indicator.
```

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,346 @@
...
+
+    kappa
+        A optional argument that sets the controls the width of the
+        transition between 0 to 1.
```
majosm, line 283:
```suggestion
    kappa
        An optional argument that controls the width of the transition from 0 to 1.
```

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,346 @@
...
+    uhat = discr.empty(actx, dtype=u.entry_dtype)
+    for group in discr.discr_from_dd("vol").groups:
+        vander = vandermonde(group.basis(), group.unit_nodes)
+        vanderm1 = np.linalg.inv(vander)
+        actx.call_loopy(
+            get_kernel(),
+            mat=actx.from_numpy(vanderm1),
+            result=uhat[group.index],
+            vec=u[group.index],
+        )
```
thomasgibson, line 311: With https://github.com/inducer/grudge/pull/77, this becomes way simpler (plus it does all the caching of the Vandermonde matrices/inverses so you don't recompute)

```python
from grudge.dof_desc import DD_MODAL, DD_VOLUME

modal_map = discr.dcoll.connection_from_dds(DD_VOLUME, DD_MODAL)
uhat = modal_map(u)
```

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,346 @@
...
+    Returns
+    -------
+    meshmode.dof_array.DOFArray
+        A DOF Array containing elementwise constant values between 0 and 1
+        which indicate the smoothness of a given element.
+    """
+    assert isinstance(u, DOFArray)
+
+    def get_kernel():
+        return linear_operator_kernel()
```
thomasgibson, line 295: This can go away with https://github.com/inducer/grudge/pull/77

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,346 @@
...
+    knl = make_loopy_program(
+        """{[iel,idof,j]:
+        0<=iel<nelements and
+        0<=idof<ndiscr_nodes_out and
+        0<=j<ndiscr_nodes_in}""",
+        "result[iel,idof] = sum(j, mat[idof, j] * vec[iel, j])",
+        name="modal_decomp",
+    )
+    knl = lp.tag_array_axes(knl, "mat", "stride:auto,stride:auto")
+    return knl
```
thomasgibson, line 250: This can go away with https://github.com/inducer/grudge/pull/77

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,346 @@
...
+
+        \mbox{RHS}_{\mbox{av}} = \nabla\cdot{\varepsilon\nabla\mathbf{Q}}
+
+    Parameters
+    ----------
+    q
+        State array which expects the quantity to be limited on to be listed
+        first in the array. For the Euler equations this could be the canonical
+        conserved variables (mass, energy, mometum) for the fluid along with a
+        vector of species masses for multi-component fluids.
```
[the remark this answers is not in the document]
w-hagen: Yes. It was originally more general, but now it is fluid specific. Will update appropriately.

mirgecom/initializers.py
w-hagen: This (and a few other pieces) are left-over from that case. I will clean it up.

mirgecom/initializers.py
w-hagen: dim is really only used currently for this check so it doesn't crash when ran with 1 or 3D. Don't know if we want to assume the user knows that the case is setup to be 2D?

majosm: I think you could just remove the `dim` argument from `__init__()` (as in `Vortex2D`), and then document that it's 2D-only. @MTCam agree?

examples/doublemach-mpi.py
```diff
@@ -0,0 +1,259 @@
...
+    rank = comm.Get_rank
+
+    gen_grid = partial(get_doublemach_mesh)
+
+    local_mesh, global_nelements = create_parallel_grid(comm, gen_grid)
+
+    local_nelements = local_mesh.nelements
+
+    discr = EagerDGDiscretization(actx, local_mesh, order=order,
+                                  mpi_communicator=comm)
```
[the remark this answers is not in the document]
w-hagen: Hmm, this looks like it would require more changes. In particular it looks like `DiscretizationCollection` does not have nodes associated with it. I think all the examples would need similar modifications?

mirgecom/initializers.py
MTCam: I think it'd still be good to toss an error when used in other than 2d.  Like:

```python
# Fail if thing numdim other than 2 
if(len(x_vec)) != 2:
    raise ValueError("Blah blah error blah")
```

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,346 @@
...
+        with the artificial viscosity application.
+    """
+    dim = discr.dim
+    cv = split_conserved(dim, q)
+
+    # Get smoothness indicator based on fluid mass density
+    indicator = smoothness_indicator(discr, cv.mass, **kwargs)
+
+    # R=Grad(Q) volume part
+    grad_q_vol = join_conserved_vectors(dim, obj_array_vectorize(discr.weak_grad, q))
```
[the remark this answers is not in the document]
MTCam: No.  It does not work as you want it to when doing grad(V) where V is a vector.

mirgecom/initializers.py
w-hagen: Okay, I am working on some revisions to this case initialization I can add that. dim is also used by the join_conserved though, but that could be hardcoded to 2 as well?

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,346 @@
...
+def _facial_flux_r(discr, r_tpair):
+    """Compute facial flux for vector compontnent of grad(Q)."""
+    r_int = r_tpair.int
+    actx = r_int[0][0].array_context
+
+    normal = thaw(actx, discr.normal(r_tpair.dd))
+
+    # This uses a central vector flux along nhat:
+    # flux = 1/2 * (grad(Q)- + grad(Q)+) .dot. nhat
+    flux_out = r_tpair.avg @ normal
```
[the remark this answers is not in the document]
MTCam: From my perspective, no.  The code expressions closely match the expressions in the BR1 paper and in Hesthaven.   Hitting it with further python magic (imo) obfuscates this correspondence, and renders the function harder to read by domain users.

mirgecom/fluid.py
MTCam: :shrug: Not that I could easily figure out... thus this beauty.

examples/doublemach-mpi.py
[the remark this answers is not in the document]
MTCam: I think this is a drop-in replacement. It should *just work*.

mirgecom/artificial_viscosity.py
MTCam: Nice.

examples/doublemach-mpi.py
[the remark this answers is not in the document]
w-hagen: I could be donig something wrong (or my branch is quite working), but I get that nodes doesn't exist
```Traceback (most recent call last):
  File "/usr/workspace/hagen7/mirgecom/quartz/mirgetesting/emirge/miniforge3/envs/ceesd/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/workspace/hagen7/mirgecom/quartz/mirgetesting/emirge/miniforge3/envs/ceesd/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/workspace/hagen7/mirgecom/quartz/mirgetesting/emirge/miniforge3/envs/ceesd/lib/python3.8/site-packages/mpi4py/__main__.py", line 7, in <module>
    main()
  File "/usr/workspace/hagen7/mirgecom/quartz/mirgetesting/emirge/miniforge3/envs/ceesd/lib/python3.8/site-packages/mpi4py/run.py", line 196, in main
    run_command_line(args)
  File "/usr/workspace/hagen7/mirgecom/quartz/mirgetesting/emirge/miniforge3/envs/ceesd/lib/python3.8/site-packages/mpi4py/run.py", line 47, in run_command_line
    run_path(sys.argv[0], run_name='__main__')
  File "/usr/workspace/hagen7/mirgecom/quartz/mirgetesting/emirge/miniforge3/envs/ceesd/lib/python3.8/runpy.py", line 265, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "/usr/workspace/hagen7/mirgecom/quartz/mirgetesting/emirge/miniforge3/envs/ceesd/lib/python3.8/runpy.py", line 97, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/usr/workspace/hagen7/mirgecom/quartz/mirgetesting/emirge/miniforge3/envs/ceesd/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "./doublemach-mpi.py", line 257, in <module>
    main()
  File "/usr/WS2/hagen7/mirgecom/quartz/mirgetesting/emirge/mirgecom/mirgecom/mpi.py", line 154, in wrapped_func
    func(*args, **kwargs)
  File "./doublemach-mpi.py", line 161, in main
    nodes = thaw(actx, discr.nodes())
AttributeError: 'DiscretizationCollection' object has no attribute 'nodes'
```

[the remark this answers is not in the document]
MTCam: I stand corrected.  I have no idea.  I guess we need to pull the volume discr out somehow.

mirgecom/initializers.py
w-hagen: Removed dim argument here: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/974bcd47752630f73c6b9a536352b4bc128a5b2c

mirgecom/initializers.py
w-hagen: Updated: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/974bcd47752630f73c6b9a536352b4bc128a5b2c

mirgecom/initializers.py
w-hagen: Removed as discussed: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/974bcd47752630f73c6b9a536352b4bc128a5b2c

mirgecom/initializers.py
w-hagen: Fixed: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/974bcd47752630f73c6b9a536352b4bc128a5b2c

mirgecom/initializers.py
w-hagen: Added equations: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/974bcd47752630f73c6b9a536352b4bc128a5b2c

mirgecom/initializers.py
```diff
@@ -262,6 +263,124 @@ def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
...
+
+    .. math::
+
+        \rho_j &= \frac{(\gamma + 1) u_s^2}{(\gamma-1) u_s^2 + 2} \\
+        p_j &= \frac{2 \gamma u_s^2 - (\gamma - 1)}{\gamma+1} \\
+        u_p &= 2 \frac{u_s^2-1}{(\gamma+1) u_s}
+
+    The initial shock location is given by $x_0$ and $u_s$ is the shock speed.
+    This function serves as an initial condition as well as boundary conditions
+    for $t>0$
```
MTCam, line 300: Actually, this object just creates the double mach solution only at time 0, and never anything else.  The argument (*t*) appears to be unused.  

Technically speaking, this routine is unrelated to the boundary condition. It simply creates the solution at locations *x_vec* - full stop.  IMO, I would remove anything having to do with "boundary condition" from the documentation, and also anything having to do with *t>0*, except maybe that *t* is unused.

Even if *t>0* was used and the routine could create the solution at *t>0*, then I still would document its ability to create a solution for *t>0*, but I'd leave the boundary conditions out of it.

mirgecom/initializers.py
```diff
@@ -262,6 +263,124 @@ def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
...
+           shock speed, Mach number
+        """
+        self._shock_location = shock_location
+        self._shock_speed = shock_speed
+
+    def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
+        r"""
+        Create the initial condition for the double Mach reflection case at
+        locations *x_vec*. Also provides boundary conditions for solution at
+        times $t>0$.
```
MTCam, line 325: Echoing my last comment regarding *t>0*, and BCs callout here should be removed.  Also, and this is mostly a nitpick, but....

This routine can be used as more than an initial condition. I disagree with documenting it as an "initial condition", and instead prefer "solution initializer".  The solution initializer is used to create the double mach reflection exact solution, the end.  The user can use that functionality how they see fit, e.g. initial condition, exact soln at boundary, etc.

mirgecom/initializers.py
```diff
@@ -262,6 +263,124 @@ def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
...
+
+        Parameters
+        ----------
+        t: float
+            Current time of solution when called as a boundary condition
```
MTCam, line 330: Ditto.

mirgecom/initializers.py
```diff
@@ -262,6 +263,124 @@ def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
...
+        gmn1 = 1.0 / gm1
+        x_rel = x_vec[0]
+        y_rel = x_vec[1]
+        actx = x_rel.array_context
+
+        zeros = 0*x_rel
+
+        shock_location = zeros + self._shock_location
+        shock_speed = zeros + self._shock_speed
+        t = zeros + t
```
MTCam, line 351: This is not necessary - both `shock_location` and `y_rel` already have the correct shape and are used in the expression below to calculate `xinter`.

examples/doublemach-mpi.py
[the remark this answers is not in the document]
MTCam: Yeah, I vote we hold off on this.  It looks like this might entail sweeping changes to all of *MIRGE-Com*.

mirgecom/initializers.py
w-hagen: So it only generates the exact solution to the problem at t=0, at t>0 it is used as a prescribed boundary condition in doublemach-mpi.py (130-132). This provides the exact far field conditions, but not does not provide the exact solution for the entirety of the domain.

MTCam: It generates the solution at locations indicated by *x_vec* and at time indicated by *t*.  It itself does not care if you use it as a BC or as an initial solution.

mirgecom/initializers.py
w-hagen: Resolved: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/f3e6a6110968876e40be153abdb8fb5787b32ecd

mirgecom/initializers.py
w-hagen: Resolved: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/f3e6a6110968876e40be153abdb8fb5787b32ecd

mirgecom/initializers.py
w-hagen: Resolved: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/f3e6a6110968876e40be153abdb8fb5787b32ecd

mirgecom/initializers.py
w-hagen: Resolved: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/f3e6a6110968876e40be153abdb8fb5787b32ecd

mirgecom/artificial_viscosity.py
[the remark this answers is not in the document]
MTCam: Oh wait... yeah you might be right. I think it might already do the right thing here without `obj_array_vectorize`. Q is not a vector quantity like those that fail when used with `weak_grad`.

examples/doublemach-mpi.py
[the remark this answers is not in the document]
thomasgibson: OK I see what happened. For now, don't apply this change; keep `EagerDGDiscretization` as is. This will need some plumbing

mirgecom/artificial_viscosity.py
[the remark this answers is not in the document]
majosm: This is kind of what I had mind, FWIW: https://github.com/majosm/mirgecom/commit/80ec49f38d6859d9ebc55d73865848caef8b85de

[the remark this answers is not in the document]
MTCam: Yeah, so I do get that your way is more pythonic.  But to me, the previous construction without the fancy stuff was more straightforward and more easily understood.

mirgecom/artificial_viscosity.py
w-hagen: Done, Thanks! https://github.com/illinois-ceesd/mirgecom/pull/213/commits/7dfa8143c5162dada86b5d819469fe53d42128ff

mirgecom/artificial_viscosity.py
w-hagen: Resolved: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/7dfa8143c5162dada86b5d819469fe53d42128ff

mirgecom/artificial_viscosity.py
w-hagen: Resolved: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/7dfa8143c5162dada86b5d819469fe53d42128ff

mirgecom/artificial_viscosity.py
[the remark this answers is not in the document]
w-hagen: Changed here: https://github.com/illinois-ceesd/mirgecom/pull/213/commits/fa4ace960be6b4b80d2f64e2ff5565131908de79

mirgecom/fluid.py
thomasgibson: I also would like to try to clean this up, but I do not understand this. Could you explain what the purpose of this [`_aux_shape` function is](https://github.com/illinois-ceesd/mirgecom/blob/main/mirgecom/fluid.py#L108-L122)? I'm trying to understand why [this is happening](https://github.com/illinois-ceesd/mirgecom/blob/main/mirgecom/fluid.py#L150-L159)

thomasgibson (commented):
This is looking really good! Just some remaining questions. Then I think this is just about ready to merge (after updating against current `main`).

mirgecom/fluid.py
thomasgibson: I tried to see if `join_conserved` could be modified as @majosm suggested, and I was unable to get something working without resulting to extremely ugly hacks. However, I think _maybe_ it is worth using identical function signatures between `join_conserved` and `join_conserved_vectors`. This requires calling `split_conserved` before passing to `join_conserved_vectors`, but it's consistent with existing behavior.

I sketched out what I mean in this commit: https://github.com/illinois-ceesd/mirgecom/commit/1518e6dec0baa8b4dd4fa53262fb0b425ab4f1dc

Just a suggestion :)

MTCam: > I also would like to try to clean this up, but I do not understand this. First, I'm trying to wrap my head around `join_conserved` to see if this can be combined somehow. Could you explain what the purpose of this [`_aux_shape` function is](https://github.com/illinois-ceesd/mirgecom/blob/main/mirgecom/fluid.py?rgh-link-date=2021-04-27T23%3A14%3A59Z#L108-L122)? I'm trying to understand why [this is happening](https://github.com/illinois-ceesd/mirgecom/blob/main/mirgecom/fluid.py?rgh-link-date=2021-04-27T23%3A14%3A59Z#L150-L159)

Just a reminder about our slack convo:
https://ceesd.slack.com/archives/D01G09PTHN3/p1619574863021000

MTCam: > I tried to see if `join_conserved` could be modified as @majosm suggested, and I was unable to get something working without resorting to extremely ugly hacks. There probably exists a simple Pythonic way to maybe unify them, but at the cost of readability...? 🤷
> 
> However, I think _maybe_ it is worth using identical function signatures between `join_conserved` and `join_conserved_vectors`. This requires calling `split_conserved` before passing to `join_conserved_vectors`, but it's consistent with existing behavior.
> 
> I sketched out what I mean in this commit: [1518e6d](https://github.com/illinois-ceesd/mirgecom/commit/1518e6dec0baa8b4dd4fa53262fb0b425ab4f1dc)
> 
> Just a suggestion :)

This would be terrific, if it worked.  

The trouble is that when one goes:
```python
# for fluid q = [ mass, energy, mom1, mom2, mom3, spm1, spm2, spm3, spm4, .....] 
r_vol = obj_array_vectorize(weak_local_grad, q)
r_cv = split_conserved(dim, r_vol)
```
The arrays you get from `r_cv.momentum`, and `r_cv.species_mass` are not correct. You can see that I pick it apart manually, even after a call to `split_conserved` inside `join_conserved_vectors`.   The gradients of the scalars mass and energy come out arranged just fine, but the gradients of the vector components do not.

mirgecom/artificial_viscosity.py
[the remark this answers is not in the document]
MTCam: Re: Why the shift to a fluid-specific operator for AV.

The previous AV operator was arranged for doing an arbitrary number of independent scalar equations dropped into a general "Q".  But the fluid equations are coupled, i.e. the fluid calculations (including boundary treatments, fluxes, EOS, etc) all require ALL components of Q simultaneously for calculations.  

It is difficult to marry the general, uncoupled Q treatment of the previous AV operator/driver with the fluid-specific coupled treatment.  Some code gymnastics (both in the operator, and in the fluid boundary treatments) were being applied to handle/assemble the required flux data at element boundaries, and it also required user-knowledge of Q structure (Q[0]=mass, Q[1]=energy), etc.).

After discussion with @w-hagen and @inducer, we (at this time) are leaning towards having a separate fluid-specific operator and a different operator for AV on uncoupled scalar components.  The resulting fluid-specific AV operator is consistent with the formulation, and with the documentation (mostly), because both specifically target the Euler eqns.

mirgecom/fluid.py
majosm: [This](https://github.com/majosm/mirgecom/commit/d053e63e237f4b5badd7d58ddb07cc7c23c74277) seemed to work in my limited testing, but we should probably take a closer look to make sure.

MTCam: Works with single gas and mixtures.

MTCam: [Check this out](https://github.com/illinois-ceesd/mirgecom/pull/298/commits/c1c0024f74cf6fa2afabbe186f2175b5b452dc15), and w-hagen/mirgecom#16.

mirgecom/artificial_viscosity.py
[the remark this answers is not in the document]
MTCam: This is complete garbage of course. @majosm fix above allowed re-generalization of the AV and will show up when w-hagen/mirgecom#16 lands.

mirgecom/boundary.py
```diff
@@ -138,4 +166,39 @@ def boundary_pair(self, discr, q, btag, **kwargs):
...
+        num_equations, dim = grad_q.shape
+        num_species = num_equations - dim - 2
+        cv = split_conserved(dim, grad_q)
+        actx = cv.mass[0].array_context
+        # Grab a unit normal to the boundary
+        normal = thaw(actx, discr.normal(btag))
+
+        # Get the interior soln
+        int_soln = discr.project("vol", btag, grad_q)
+        int_cv = split_conserved(dim, int_soln)
```
majosm, line 183: I noticed this on the other PR, but I wasn't sure whether to comment there or here. The naming of `int_soln` and `int_cv` here is confusing IMO.

mirgecom/artificial_viscosity.py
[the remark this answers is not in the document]
majosm: inducer/grudge#87

mirgecom/boundary.py
MTCam: Heh, yeah.  What about this:
```python
facial_gradq = discr.project("vol", btag, grad_q)
gradq_cv = split_conserved(dim, facial_gradq)
```
The only reason I didn't go with something like `gradq_components` is due to the unwieldy length of it.

majosm: Either sounds good. I used `*_int` in the diffusion BCs: https://github.com/illinois-ceesd/mirgecom/blob/main/mirgecom/diffusion.py#L145

mirgecom/fluid.py
thomasgibson: Hooray! I like this. I think we found a winner

mirgecom/boundary.py
MTCam: w-hagen/mirgecom#17

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,325 @@
...
+
+where $\mathbf{R}$ is an intermediate variable, and the artitifial viscosity
+coefficient, $\varepsilon$, is spatially dependent and calculated using the
+smoothness indicator of [Persson_2012]_:
+
+.. math::
+
+    S_e = \frac{\langle u_{N_p} - u_{N_{p-1}}, u_{N_p} -
+        u_{N_{p-1}}\rangle_e}{\langle u_{N_p}, u_{N_p} \rangle_e}
+
```
majosm, line 32: Should this stuff be reworded post-generalization?

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,325 @@
+r""":mod:`mirgecom.artificial_viscosity` Artificial viscocity for Euler.
```
majosm, line 1:
```suggestion
r""":mod:`mirgecom.artificial_viscosity` Artificial viscosity for Euler.
```

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,325 @@
...
+
+To evalutate the second order derivative the problem is recast as a set of first
+ order problems:
+
+.. math::
+
+    \partial_t{\mathbf{Q}} &= \nabla\cdot\mathbf{R} -\nabla\cdot\mathbf{F}^I \\
+    \mathbf{R} &= \varepsilon\nabla\mathbf{Q}
+
+where $\mathbf{R}$ is an intermediate variable, and the artitifial viscosity
```
majosm, line 24:
```suggestion
where $\mathbf{R}$ is an intermediate variable, and the artificial viscosity
```

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,325 @@
...
+        u_{N_{p-1}}\rangle_e}{\langle u_{N_p}, u_{N_p} \rangle_e}
+
+where:
+
+- $S_e$ is the smoothness indicator
+- $u_{N_p}$ is the solution in modal basis at the current polynomial order
+- $u_{N_{p-1}}$ is the truncated modal represention to the polynomial order $p-1$
+- The $L_2$ inner product on an element is denoted $\langle \cdot,\cdot \rangle_e$
+
+The elementwise viscoisty is then calculated:
```
majosm, line 40:
```suggestion
The elementwise viscosity is then calculated:
```

mirgecom/artificial_viscosity.py
thomasgibson: https://github.com/w-hagen/mirgecom/pull/18 @MTCam

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,328 @@
...
+where $\varepsilon$ is the artificial viscosity coefficient.
+To evalutate the second order derivative numerically, the problem
+is recast as a set of first order problems:
+
+.. math::
+
+    \partial_t{\mathbf{Q}} + \nabla\cdot\mathbf{F} &= \nabla\cdot\mathbf{R} \\
+    \mathbf{R} &= \varepsilon\nabla\mathbf{Q}
+
+where $\mathbf{R}$ is an auxiliary variable, and the artitifial viscosity
```
w-hagen, line 27: Missed this, adding it for a a batch fix
```suggestion
where $\mathbf{R}$ is an auxiliary variable, and the artificial viscosity
```

thomasgibson:
```suggestion
where $\mathbf{R}$ is an auxiliary variable, and the artificial viscosity
```

MTCam:
I really wish that `Approve and Run` button would come pre-pressed from the factory.

mirgecom/fluid.py
```diff
@@ -6,6 +6,7 @@
 .. autoclass:: ConservedVars
 .. autofunction:: split_conserved
 .. autofunction:: join_conserved
+.. autofunction:: join_conserved_vectors
```
MTCam, line 9:
```suggestion
```

mirgecom/fluid.py
```diff
@@ -299,6 +300,21 @@ def join_conserved(dim, mass, energy, momentum,
...
+    cv = split_conserved(dim, ary)
+    retval[0] = cv.mass
+    retval[1] = cv.energy
+    for i in range(dim):
+        retval[2+i] = ary[2+i]
+    for i in range(nspecies):
+        retval[dim+2+i] = ary[dim+2+i]
+    return retval
+
+
```
MTCam, line 317:
```suggestion
```

mirgecom/initializers.py
```diff
@@ -262,6 +263,120 @@ def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
...
+        t: float
+            Time at which to compute the solution
+        x_vec: numpy.ndarray
+            Nodal coordinates
+        eos: :class:`mirgecom.eos.GasEOS`
+            Equation of state class to be used in construction of soln (if needed)
+        """
+        # Fail if numdim is other than 2
+        if(len(x_vec)) != 2:
+            raise ValueError("Case only defined for 2 dimensions")
```
MTCam, line 339: I have a question about this block.  This routine gets called from the RHS boundary treatments.  Would an `assert` actually be superior in this case so that the branch can effectively be disabled at the command line?

majosm: What's the motivation for disabling it?

MTCam: I wonder if the branch here that is frequently used in a performance sensitive construct is a performance drag - especially in the context of our code transforms and associated analysis.

If so, it might be nice if `python -O` could eliminate data/shape validation checks when running production.

thomasgibson: Isn't this only ever called once though? At the start of a simulation?

w-hagen: Looks like every time it is called. And this initial condition is also used as a prescribed boundary so it will called for in RHS boundary operations.

MTCam: It is called at every sub-step on every boundary that uses this as an exact bc.

majosm: My hunch is that it would be negligible. I think we can leave it as is until we actually know that it’s an issue.

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,328 @@
...
+
+
+def _facial_flux_q(discr, q_tpair):
+    """Compute facial flux for each scalar component of Q."""
+    flux_dis = q_tpair.avg
+    if isinstance(flux_dis, np.ndarray):
+        actx = flux_dis[0].array_context
+        flux_dis = flux_dis.reshape(-1, 1)
+    else:
+        actx = flux_dis.array_context
```
majosm, line 113: I tried to clean up this handling a little: w-hagen/mirgecom#19

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,328 @@
...
+    normal = thaw(actx, discr.normal(r_tpair.dd))
+
+    # This uses a central vector flux along nhat:
+    # flux = 1/2 * (grad(Q)- + grad(Q)+) .dot. nhat
+    flux_out = flux_dis @ normal
+
+    return discr.project(r_tpair.dd, "all_faces", flux_out)
+
+
+def av_operator(discr, t, eos, boundaries, q, alpha, **kwargs):
```
majosm, line 141: Can we document what `**kwargs` is used for?

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,328 @@
...
+
+        Time
+
+    alpha: float
+
+       The maximum artificial viscosity coefficient to be applied
+
+    eos: :class:`~mirgecom.eos.GasEOS`
+
+       Only used as a pass through to the boundary conditions.
```
majosm, line 177: Would it make sense to pass something like a `**bdry_kwargs` instead of passing the EOS explicitly?

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,328 @@
...
+        on. Values in the range $(-\infty,0]$ are allowed, where $-\infty$ results in
+        all cells being tagged and 0 results in none.
+
+    Returns
+    -------
+    meshmode.dof_array.DOFArray
+        The elementwise constant values between 0 and 1 which indicate the smoothness
+        of a given element.
+    """
+    assert isinstance(u, DOFArray)
```
majosm, line 288: Is this `assert` just a sanity check, or is it verifying user input?

majosm (commented):
Looking pretty nice, just noticed a couple of things after another read-through.

mirgecom/artificial_viscosity.py
w-hagen: Will do

mirgecom/artificial_viscosity.py
MTCam: How's come it couldn't be passed through the **kwargs?  (assuming you are suggesting the boundary should have its own kwargs struct).

w-hagen: I tried doing something like that once and I encountered some problems. That said I don't recall what they were (probably me not understanding python...). I can give something like this another  try though.

MTCam: Should time (*t*) also go into **kwargs, or whatever is used to contain args to boundary routines?

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,328 @@
...
+    indicator = discr.empty(actx, dtype=u.entry_dtype)
+    for group in discr.discr_from_dd("vol").groups:
+        mode_ids = group.mode_ids()
+        modes = len(mode_ids)
+        order = group.order
+        highest_mode = np.ones(modes)
+        for mode_index, mode_id in enumerate(mode_ids):
+            highest_mode[mode_index] = highest_mode[mode_index] * (
+                sum(mode_id) == order
+            )
```
inducer, line 308: This is probably more readable when done in a list comprehension.
`[1 if sum(mode_id) == order else 0 for mode_id in mode_ids]` or so.

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,328 @@
...
+
+    knl = make_loopy_program(
+        """{[iel,idof,j,k]:
+        0<=iel<nelements and
+        0<=idof<ndiscr_nodes_out and
+        0<=j<ndiscr_nodes_in and
+        0<=k<ndiscr_nodes_in}""",
+        "result[iel,idof] = "
+        "sum(k,vec[iel,k]*vec[iel,k]*modes[k])/sum(j,"
+        " vec[iel,j]*vec[iel,j]+1.0e-12/ndiscr_nodes_in)",
```
inducer, line 261: The denominator here is not an accurate approximation of the L^2 norm.

mirgecom/boundary.py
```diff
@@ -138,4 +165,28 @@ def boundary_pair(self, discr, q, btag, **kwargs):
                                     momentum=ext_mom,
                                     species_mass=int_cv.species_mass)
 
-        return TracePair(btag, interior=int_soln, exterior=bndry_soln)
+        return bndry_soln
+
+    def exterior_grad_q(self, discr, grad_q, btag, **kwargs):
```
inducer, line 170: It'd be nice if this could make use of @MTCam's upcoming BC work that's usable for both diffusion and fluid PDEs.

requirements.txt
inducer, line 9: This cannot stay. Installing `gmsh` via pip is not an option for us. It won't work on Lassen, and it may clash with our conda packages.

inducer:

@w-hagen To help me save time, could you wait until you've got an approval from one of the core devs before requesting another review?

MTCam:
> Unsubscribing... @-mention or request review once it's ready for a look or needs attention.
> 
> @w-hagen To help me save time, could you wait until you've got an approval from one of the core devs before requesting another review?

This was totally my fault.  I told Wyatt he could go ahead and request your eyes on it.

mirgecom/artificial_viscosity.py
thomasgibson, line 261: I'm having a hard time see how this kernel is doing this:
image
Am I just being dense? Shouldn't it look something like this:
```python
"{[iel]: 0 <= iel < nelements}",
"{[idof]: 0 <= idof < ndiscr_nodes_in}",
"{[jdof]: 0 <= jdof < ndiscr_nodes_in}",
"{[kdof]: 0 <= kdof < ndiscr_nodes_in}"
],
"""
    result[iel, idof] = sum(kdof, (u[iel, kdof] - uhat[iel, kdof])    \
                                  * (u[iel, kdof] - uhat[iel, kdof])  \
                                  * weights[kdof]) /                  \
                        sum(jdof, u[iel, jdof] * u[iel, jdof]         \
                                  * weights[jdof])
"""
```

inducer: Actually, I got that wrong. Since `vec` consists of modal coefficients wrt an orthonormal basis, this is OK.

inducer, line 261:
```suggestion
        """
        result[iel,idof] = (
          sum(k,vec[iel,k]*vec[iel,k]*modes[k])
          / sum(j, vec[iel,j]*vec[iel,j]+1.0e-12/ndiscr_nodes_in))
        """,
```
No functional changes (I hope), just better formatting.

inducer: Since it's being fed modal data, it can use [Parseval's identity](https://en.wikipedia.org/wiki/Parseval%27s_theorem) to compute L^2 norms without needing weights. And `u-uhat` is being written `u*mode_flag`.

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,328 @@
...
+    from meshmode.array_context import make_loopy_program
+
+    knl = make_loopy_program(
+        """{[iel,idof,j,k]:
+        0<=iel<nelements and
+        0<=idof<ndiscr_nodes_out and
+        0<=j<ndiscr_nodes_in and
+        0<=k<ndiscr_nodes_in}""",
+        "result[iel,idof] = "
+        "sum(k,vec[iel,k]*vec[iel,k]*modes[k])/sum(j,"
```
inducer, line 260: Maybe rename `modes` to `mode_active_flag`?

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,328 @@
...
+    """Compute the smoothness indicator for all elements."""
+    from meshmode.array_context import make_loopy_program
+
+    knl = make_loopy_program(
+        """{[iel,idof,j,k]:
+        0<=iel<nelements and
+        0<=idof<ndiscr_nodes_out and
+        0<=j<ndiscr_nodes_in and
+        0<=k<ndiscr_nodes_in}""",
+        "result[iel,idof] = "
```
inducer, line 259: The RHS of this assignment is being computed repeatedly, once for each DOF. IMO it would be better to compute it once per element.

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,328 @@
...
+            interior=discr.project("vol", btag, q),
+            exterior=boundaries[btag].exterior_q(discr, btag=btag, t=t,
+                                                 q=q, eos=eos))
+        q_flux_db = q_flux_db + _facial_flux_q(discr, q_tpair=q_tpair)
+    # Total Q flux across element boundaries
+    q_bnd_flux = q_flux_int + q_flux_pb + q_flux_db
+
+    # Compute R
+    r = discr.inverse_mass(
+        -alpha * indicator * (grad_q_vol - discr.face_mass(q_bnd_flux))
```
inducer, line 214: When I did related work, I found it essential to use a "smooth" viscosity (otherwise strong AV might produce more nonsmoothness than it would remove). I got acceptable results using the following steps:
image

(https://doi.org/10.1051/mmnp/20116303, Section 5.2)

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,328 @@
...
+            btag,
+            interior=discr.project("vol", btag, r),
+            exterior=boundaries[btag].exterior_grad_q(discr, btag=btag, t=t,
+                                                      grad_q=r, eos=eos))
+        r_flux_db = r_flux_db + _facial_flux_r(discr, r_tpair=r_tpair)
+    # Total grad(Q) flux element boundaries
+    r_flux_bnd = r_flux_int + r_flux_pb + r_flux_db
+
+    # Return the AV RHS term
+    return discr.inverse_mass(-div_r_vol + discr.face_mass(r_flux_bnd))
```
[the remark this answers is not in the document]
majosm: w-hagen/mirgecom#21

I had to refactor the diffusion operator somewhat so that it could take in fluid BC objects. @MTCam tells me the BC APIs are currently in flux (heh), so this may need further tweaking to be compatible with that.

Otherwise it seems like it works... The AV tests pass, and the doublemach-mpi case looks visually similar (though not numerically identical) after 100 steps. I suspect the differences come down to the AV implementation and the diffusion operator using different conventions for the auxiliary variable (`alpha*grad(u)` vs. `grad(u)`). Are there any more tests we can run to make sure this is ok @w-hagen?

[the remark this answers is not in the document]
w-hagen: Great. Thanks Matt, I have a couple of ideas rolling around in my head at the moment on how we could possibly add some more testing to this. Let me think on it a bit more and maybe try to implement some additional tests.

MTCam:
CI failures due to inducer/grudge#83 moving stuff around. @thomasgibson.  Fixed these in NS with:
(e451ea4, e4cb256).

thomasgibson:
> CI failures due to [inducer/grudge#83](https://github.com/inducer/grudge/pull/83) moving stuff around. @thomasgibson. Fixed these in NS with:
> ([e451ea4](https://github.com/illinois-ceesd/mirgecom/commit/e451ea4622f37546b585e3de898d047a10e23bf8), [e4cb256](https://github.com/illinois-ceesd/mirgecom/commit/e4cb256ca00d689dc7c9bc71b9ec87f6378aed9f)).

@MTCam I'll take care of these updates and I'll file a mirgecom PR and one against @w-hagen's branch

MTCam:
> > CI failures due to [inducer/grudge#83](https://github.com/inducer/grudge/pull/83) moving stuff around. @thomasgibson. Fixed these in NS with:
> > ([e451ea4](https://github.com/illinois-ceesd/mirgecom/commit/e451ea4622f37546b585e3de898d047a10e23bf8), [e4cb256](https://github.com/illinois-ceesd/mirgecom/commit/e4cb256ca00d689dc7c9bc71b9ec87f6378aed9f)).
> 
> @MTCam I'll take care of these updates and I'll file a mirgecom PR and one against @w-hagen's branch

Thanks a ton, @thomasgibson!

mirgecom/artificial_viscosity.py
w-hagen: So my thinking was that the smoothness_indicator can be called directly by the user (mostly for the user to add an output field of the tagged cells). This was meant to ensure the user passed a DOFArray and not a numpy object array of the state, but it may not be necessary?

majosm: Ah, ok. I think we want to use exceptions instead of assertions when it's checking what the user passed in.

```suggestion
    if not isinstance(u, DOFArray):
        raise ValueError("u argument must be a DOFArray.")
```

requirements.txt
majosm: I tried to replace this with a gmsh entry in `conda-env.yml`, but I was having some trouble getting it to work on lassen (I was getting dependency conflicts with cantera). @matthiasdiener mentioned that this might go away with conda-forge/gmsh-feedstock#44.

mirgecom/artificial_viscosity.py
majosm: w-hagen/mirgecom#23

mirgecom/artificial_viscosity.py
```diff
@@ -0,0 +1,346 @@
...
+                vec=uhat[grp.index],
+                modes_active_flag=highest_mode(grp))["result"]
+            for grp in discr.discr_from_dd("vol").groups
+        )
+    )
+    indicator = actx.np.log10(indicator + 1.0e-12)
+
+    # Compute artificial viscosity percentage based on indicator and set parameters
+    yesnol = indicator > (s0 - kappa)
+    yesnou = indicator > (s0 + kappa)
```
matthiasdiener, line 338:
```suggestion
    yesnol = actx.np.greater(indicator, (s0 - kappa))
    yesnou = actx.np.greater(indicator, (s0 + kappa))
```

@kaushikcfd

MTCam:
Closing this as stale.  We merged AV in #582

PR #225, opened 2021-02-07
Remove source term documentation in Euler

MTCam:
This just removes the documentation of sources from Euler, per the plan to leave source terms up to the user.

### DIFF
--- a/mirgecom/euler.py
+++ b/mirgecom/euler.py
@@ -220,7 +220,7 @@ def inviscid_flux(discr, eos, q):
     return join_conserved(dim,
             mass=mom,
             energy=mom * (cv.energy + p) / cv.mass,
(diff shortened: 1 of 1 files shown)


inducer:
I'm not totally sold.

I like
```
f(x) + p
```
better than
```
f2(x, add_this=p)
```

MTCam:
> ```
> (f(x) + p) > f2(x, add_this=p)
> ```

I can get on-board with that.  Should we also tweak the documentation to be consistent?  Both our module-level and RHS-level docs mention sources.

inducer:
> Should we also tweak the documentation to be consistent? Both our module-level and RHS-level docs mention sources.

Yep, sounds good.

MTCam:
To explain: We will no longer handle sources inside the Euler module.  Any source handling will be exported to the driver or some other component TBD.

inducer (approved):

PR #232, opened 2021-02-08
Add CTI file management infrastructure for thermochemistry configs

MTCam:
This adds some infrastructure for managing CTI files that are used in the configuration of `Cantera`.  This infrastructure (or similar) is needed to support the addition of `Pyrometheus`-based EOS to `MIRGE-Com`.

Generally, to use this, the user only needs to do something like:
```
from mirgecom.mechanisms import get_mechanism_cti
mech_cti = get_mechanism_cti("uiuc")
cantera_soln = Cantera.Solution(source=mech_cti)
```

The use of `importlib` in this infrastructure is supposed to work transparently regardless of whether `MIRGE-Com` is installed to ZIP.  The infrastructure was developed/worked in #207, previously.


doc/support/thermochem.rst
```diff
@@ -0,0 +1,4 @@
+Thermochemistry mechanism management
```
inducer, line 1: If this file is being added, it should probably be added to a toctree as well.

mirgecom/mechanisms/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+    # function, so use the PyPI version:
+    import importlib_resources
+else:
+    # importlib.resources has files(), so use that:
+    import importlib.resources as importlib_resources
+
+
+def get_mechanisms_pkgname() -> str:
+    """Get a qualified package name for the location of mechanism data."""
+    return("mirgecom.mechanisms")
```
inducer, line 45:
```suggestion
    return "mirgecom.mechanisms"
```
(same below)

mirgecom/mechanisms/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+    Returns
+    -------
+    :class:`importlib.abc.Traversable`
+        Object of type :class:`importlib.abc.Traversable` representing the container
+        (think directory) of the thermochemistry mechanism data (think CTI files).
+    """
+    return importlib_resources.files(get_mechanisms_pkgname())
+
+
+def get_mechanism_cti(mechanism_name: str) -> str:
```
inducer, line 65: Add a test for this?

doc/support/thermochem.rst
MTCam: I added this under *support*, and in the toctree there already.  Is there an additional place to add it?

mirgecom/mechanisms/__init__.py
MTCam: Before I add a dedicated test for this, there are subsequent PRs that have this construct used extensively in tests and examples.  Sorry - the way I have it split up, it isn't apparent that this function will ever be used.

mirgecom/mechanisms/__init__.py
MTCam: Grr.  I just cannot get my fingers to stop doing that!

mirgecom/mechanisms/__init__.py
inducer: Sure. Even so, can't hurt to have a test shows that this function can be called and won't crash.

MTCam: Sure! Now if I can just figure out how to verify that the thing is reading properly without using the thing itself to get the verification data :)

MTCam: (7cbb272) ?

MTCam: Is there any way to add a test to see if it still works even through a ZIP install?

inducer: :shrug: Good enough for me.

doc/support/thermochem.rst
inducer: Duh. Missed that. (despite it being right above this in the diff) Sorry! All good.

inducer (approved):

MTCam:
Turns out adding this simple, innocuous seeming test caught a major problem.

requirements.txt
```diff
@@ -4,6 +4,7 @@ numpy
 pytest
 pyvisfile
 pymetis
+importlib-resources>=1.1.0; python_version < '3.9'
```
matthiasdiener, line 7: This change probably broke emirge.

MTCam: Oh hrm, I haven't had any troubles with it, but maybe I would need --branch-name=<something> to catch it?

matthiasdiener: > Oh hrm, I haven't had any troubles with it, but maybe I would need --branch-name= to catch it? 

Any fresh installation of emirge will fail, see e.g. here https://github.com/illinois-ceesd/emirge/runs/1865330076

> (edit: pip thinks it's ok).

Emirge doesn't use pip to parse this file.

What is the intention here, why not install this package for python 3.9 too?

MTCam: For python >= 3.9, importlib-resources.files is baked-in.  See e.g.
https://jwodder.github.io/kbits/posts/pypkg-data/#accessing-package-data-at-runtime

matthiasdiener: Sure, but installing the package won't hurt, right?

MTCam: 🤷 Elephino

PR #239, opened 2021-02-09
Fix CTI file reading test failures

MTCam:
The CTI test isn't working because `b'test string' != 'test string'`.  Not sure of the best way to handle when fp.read() returns a stream of bytes.  Appears to be platform, or at least Python version dependent.

Should note: `Cantera` itself doesn't care.  It will take a string or the bytes and work the same. Only the test actually cares about this.


matthiasdiener:
cc #232. Are the "normal" CTI files (not the test CTI file) just text files?

MTCam:
> cc #232. Are the "normal" CTI files (not the test CTI file) just text files?

Yes. They use 1 of 4 formats, YAML (preferred), CTI (deprecated), XML (ancient and deprecated), and CK (input format for proprietary ChemKin package).  Our infrastructure can easily support any of these - but all of our current mechanisms are in CTI format (which is indeed a text-based format).

matthiasdiener:
I'm unable to reproduce this issue at all on my local system. Why would it open the file in binary mode? Due to the missing line ending in `testme.cti`?

MTCam:
> I'm unable to reproduce this issue at all on my local system. Why would it open the file in binary mode? Due to the missing line ending in `testme.cti`?

It seems platform or Python dependent.  It fails in CI on ubuntu but seems to work in MacOS.

matthiasdiener (approved):
Looks fine. Imho, it would have been easier to just compare (e.g.) the first line of a "real" cti file, instead of adding a new file, but it's fine like this too.

MTCam:
> Looks fine. Imho, it would have been easier to just compare (e.g.) the first line of a "real" cti file, instead of adding a new file, but it's fine like this too.

Well let's fix it up; I like your idea better.

matthiasdiener (approved):
~Should adjust the title of this PR.~ Nvm

MTCam:
> ~Should adjust the title of this PR.~ Nvm

I *really* need to re-think my PR naming approach.  I'm not sure I've ever had one with a stable title.

test/test_cti.py
```diff
@@ -30,5 +30,7 @@
 
 def test_cti_reader():
     """Quick test of CTI reader."""
-    test_cti = get_mechanism_cti("testme")
-    assert test_cti == "This is just for testing. Do not remove."
+    test_cti = get_mechanism_cti("uiuc")
+    test_string = test_cti.partition("\n")[0]
```
inducer, line 34:
```suggestion
    test_string, _, _ = test_cti.partition("\n")
```

MTCam: df0fa0a

PR #216, opened 2021-02-04
Undo pytest memory bandage.

MTCam:
Just a placeholder CI test to demonstrate the issue until #212 resolution.

Issue #212, opened 2021-02-03
Pytest crashes in CI after diffusion tests


MTCam:
[This diffusion test](https://github.com/illinois-ceesd/mirgecom/blob/ce4e0229aa32897d3be84062392b24993dfc95b7/test/test_diffusion.py#L241) that uses over-integration seems to require a great deal of memory in CI, causing subsequent crash of `pytest` without error messages in #95. 

Here's a plot of the memory usage of `pytest` with and without the diffusion accuracy test. The blue data has the diffusion accuracy test disabled.  When the diffusion accuracy test is disabled, CI in #95 runs to completion with all tests passing.  When the diffusion test is enabled, `pytest` crashes inexplicably in the next suite/file. 

![Screen Shot 2021-02-03 at 8 33 05 AM](https://user-images.githubusercontent.com/8116930/106761453-74c18100-65fa-11eb-98df-a7b8b8ea7679.png)

MTCam:
Appears to be fixed by #211.

inducer:
#211 manages to paper over this, sure, but I wouldn't call this a fix. #150 reports a similar memory growth issue. I propose we devote some time to finding out what's going on. There's no reason why memory use should grow like this.

majosm:
I'm seeing different behavior on my laptop:

![Figure_2](https://user-images.githubusercontent.com/7232918/106780261-6d57a300-660d-11eb-8966-755baeaac807.png)

This is with my old diffusion dev environment though. I'm currently re-running with a fresh emirge.

majosm:
Hmmm, nope. No memory blowup with fresh emirge either:

![Figure_1](https://user-images.githubusercontent.com/7232918/106781022-35049480-660e-11eb-959b-560f923fdf7f.png)

inducer:
Hmm... interesting. Are you both running the same thing and getting different results?

MTCam:
I'm running it again, it's just taking a long time. Update as soon as the data is in.

majosm:
Seems like it. I'm running pytest with `-sv`, but I don't see why that would make a difference. I can try a run without just to make sure.

MTCam:
> Seems like it. I'm running pytest with `-sv`, but I don't see why that would make a difference. I can try a run without just to make sure.

A little redundant, but just wanted to make sure I run exactly the same thing.  This was the diffusion test only on fresh emirge/master. 

[...]

PR #240, opened 2021-02-09
Eliminate silent failures in CI.

### LINKED ISSUE #241: CI fails silently
I am sorry to report this:

CI is currently failing silently.  Turns out that papering over the memory issue (can be undone with #216) was a bad idea, or at least needs tweaking.  Looking down through the CI capture for `pytest` sections, I saw a failing test for CTI file reads, but a green checkmark on the CI dashboard.

The CTI file read issue appears to be fixed (by #239), so it is no longer a good example, unless we can dig up CI history on it.

The fix for CI may be as simple as #240, or similar. Either way, paper can only go so far and it is tearing.

MTCam:
Trying to get CI to speak up when individual `pytest` invocation has failing tests.  Individual `pytest` running was introduced to paper over a memory issue (#212).  The papering can be undone with #216.

The current example where CI fails silently is in [this CI run](https://github.com/illinois-ceesd/mirgecom/runs/1860379003?check_suite_focus=true) of #239. Unfortunately, this is already happening in master as a test in #232 causes a failure, but CI dashboard has a green checkmark.

* The failing test from #232 is now fixed by #239.
* _This_ PR (#240) successfully gets CI to fail when individual files are tested with `pytest`. It should be merged to keep the paper-over working properly. 
* The CI for #240 should only pass when all the tests in master are passing

Fixes #241


matthiasdiener:
Why not just `python ... || exit 1`?

majosm:
Silent CI failures?

MTCam:
Kind of strange... If it *fails* that means it's working, and is ready for merge.
:thinking:

majosm:
May want to do `exit $?` so we get the right error code.

This still seems weird to me... I don't get why the explicit `exit` should be necessary.

matthiasdiener:
> May want to do `exit $?` so we get the right error code.

👍 

> This still seems weird to me... I don't get why the explicit `exit` should be necessary.

I think because the CI checks the return code of the whole script. Bash usually just continues executing even when an error occurs midway through the script. (this is why options such as `set -o errexit` exist)

majosm:
Ohhh. That's nasty. Is it possible/advisable to do `set -e`?

inducer:
> Is it possible/advisable to do `set -e`?

- IMO: That should always be the first line of any shell script.
- AFAIK, Github CI runs its scripts with `set -e` by default.

matthiasdiener:
> Ohhh. That's nasty. Is it possible/advisable to do `set -e`?

I think that's risky. Sometimes you want it to continue on failure. E.g., the following does not work:

```bash
set -o errexit

ret=$(false)

# Not executed:
echo $ret
```

matthiasdiener:
> * AFAIK, Github CI runs its scripts with `set -e` by default.

[Yes](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell), but we override it.

majosm:
> I think that's risky. Sometimes you want it to continue on failure. E.g., the following does not work:

I think when you want to continue on failure you can do:

```bash
<command> || result=$?

echo $result
```

Exit on error seems like a safer default to me.

matthiasdiener:
> > I think that's risky. Sometimes you want it to continue on failure. E.g., the following does not work:
> 
> I think when you want to continue on failure you can do:
> 
> ```shell
> <command> || result=$?
> 
> echo $result
> ```
> 
> Exit on error seems like a safer default to me.

Sure, but you might want to run external scripts that rely on the default error behavior.

matthiasdiener:
> > Exit on error seems like a safer default to me.
> 
> Sure, but you might want to run external scripts that rely on the default error behavior.

Just to be clear, I am not against setting -e on these CI scripts, just that the rules for errexit are really arcane (see https://stackoverflow.com/questions/44819831/bash-errexit-with-arithmetic-expansion for a fun one) and not very reliable.

matthiasdiener (approved):
(cc #239)

MTCam:
Which should be merged first?  My current plan would be to do #239 firs, then #240 should start passing, and I'll merge it. Do we need further discussion or final review @inducer?

.github/workflows/ci.yaml
```diff
@@ -91,7 +91,7 @@ jobs:
             # One test at a time to address ballooning memory.
             # https://github.com/illinois-ceesd/mirgecom/issues/212
             for filename in test_*.py ; do
-              python -m pytest --durations=10 --tb=native --junitxml=pytest.xml --doctest-modules -rxsw ./"$filename"
+              python -m pytest --durations=10 --tb=native --junitxml=pytest.xml --doctest-modules -rxsw ./"$filename" || exit $?
```
inducer, line 94: > > * AFAIK, Github CI runs its scripts with `set -e` by default.
> 
> [Yes](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell), but we override it.
> 
> Edit: here is where we override that:
> 
> https://github.com/illinois-ceesd/mirgecom/blob/d5c6b0ddf8da1778268e13df169174216a5aef3c/.github/workflows/ci.yaml#L82-L83

My strong preference would be to undo the shell override instead and leave the default of `set -e` in place.

inducer (changes requested):

.github/workflows/ci.yaml
MTCam: Wanna take a crack at this @matthiasdiener?

MTCam: Successfully failed. :tada:

matthiasdiener: its rare to be so happy about a failing CI run :-D

MTCam:
I guess our changes to the CTI file reader produced another non-working version:
```
_______________________________ test_cti_reader ________________________________
  File "/home/runner/work/mirgecom/mirgecom/test/test_cti.py", line 34, in test_cti_reader
    first_line, _, _ = test_cti.partition("\n")
TypeError: a bytes-like object is required, not 'str'
--- generated xml file: /home/runner/work/mirgecom/mirgecom/test/pytest.xml ----
============================= slowest 10 durations =============================
```

.github/workflows/ci.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
           with:
             python-version: '3.x'
         - name: Install
-          shell: bash -l {0}
+          shell: bash -e -l {0}
```
inducer, line 39:
```suggestion
```
Or, put, differently, what's the advantage to having this? (Sorry, we may have had this discussion. I just don't remember.)

matthiasdiener: I dont remember either. For some reason, we must have thought of using a login shell here.

inducer: Kill it?

inducer (approved):

PR #242, opened 2021-02-09
fix importlib-resources for emirge

matthiasdiener:


matthiasdiener:
Confirmed to fix the issue introduced in #232.

MTCam (approved):
LGTM!

requirements.txt
```diff
@@ -4,7 +4,7 @@ numpy
 pytest
 pyvisfile
 pymetis
-importlib-resources>=1.1.0; python_version < '3.9'
+importlib-resources
```
inducer, line 7: This change strikes me as incorrect. Why is this right?

matthiasdiener: Emirge can't parse this line and just does nonsense. Installing this package even for Py3.9 seems to have no ill effect (see e.g. https://pypi.org/project/importlib-resources/)

MTCam: At risk of adding unwanted noise here.. I think that this change might be OK simply because of this:
```
if sys.version_info < (3, 9):
    # importlib.resources either doesn't exist or lacks the files()
    # function, so use the PyPI version:
    import importlib_resources
else:
    # importlib.resources has files(), so use that:
    import importlib.resources as importlib_resources
```
This snippet seems to be OK even if both versions co-exist?

inducer: Fair enough. Also, #244.

PR #244, opened 2021-02-09
Point out syntax restrictions in requirements.txt

inducer:


matthiasdiener (approved):

matthiasdiener:
Imho, at least for the medium-term, we should probably abandon the requirements.txt parsing completely, and just read the file line-by-line, and feeding each line to pip directly. With a special case for the development packages such as loopy, where we just add a `--src` argument to pip.

inducer:
> Imho, at least for the medium-term, we should probably abandon the requirements.txt parsing completely, and just read the file line-by-line, feeding each line to pip directly. With a special case for the development packages such as loopy, where we just add a `--src` argument to pip.

Not opposed. Want to create an emirge issue?

PR #245, opened 2021-02-09
Fix failing CTI test 

MTCam:
Continued from #239 - which did not fix the problem.  This time #240 is used as the base so that we can make sure that it fixes before sending it to master.

Edit: This one passed with #240 as the base - and inspection of the CI capture revealed that it really did pass.   So.. switched the base back to master.


MTCam:
This really does fix the CTI read, we missed the failure before because it was silent (i.e. the thing we're fixing in #240.

mirgecom/mechanisms/__init__.py
```diff
@@ -67,5 +67,9 @@ def get_mechanism_cti(mechanism_name: str) -> str:
     mech_data = import_mechdata()
     mech_file = mech_data / get_mechanism_file_name(mechanism_name)
     with mech_file.open() as fp:
-        mech_cti = fp.read()
+        read_data = fp.read()
+    try:
+        mech_cti = read_data.decode()
+    except (UnicodeDecodeError, AttributeError):
+        return read_data
```
inducer, line 74: This looks super fishy to me. How come Cantera needs us to do unicode decoding on its files?

MTCam: Cantera doesn't (seem to) care whether we pass in bytes or a string. The test cares, because it calls a string-specific built-in.  The real weird thing is that on ubuntu VM in CI, without this in-place, this routine returns a byte stream. (!)

inducer: Could you add comments to state these things?

Also, maybe
```
if isinstance(read_data, bytes):
```
to clarify the intention?

MTCam: (b97443d) I'm with you, this seems fishy still.

inducer (approved):

PR #246, opened 2021-02-10
Use jemalloc in Linux CI

inducer:


MTCam:
It is interesting to me that this make the CI pass.  Apparently the memory issue #212 only crashes the linux CI and then the mac one just stops?

matthiasdiener:
> It is interesting to me that this make the CI pass. Apparently the memory issue #212 only crashes the linux CI and then the mac one just stops?

Currently, CI tests within the same matrix strategy are cancelled if one of them fails. We could disable this behavior with the `fail-fast: false` directive (as done here for example: #247)

inducer:
> Currently, CI tests within the same matrix strategy are cancelled if one of them fails. We could disable this behavior with the `fail-fast: false` directive (as done here for example: #247)

IMO it's OK to leave this at its default. In a sense, we're spending Github's money here (i.e. what they pay for the compute time), and it's on us to be good stewards of that.

doc/running/memory.rst
```diff
@@ -0,0 +1,38 @@
...
+    LD_PRELOAD=$CONDA_PREFIX/lib/libjemalloc.so.2 python myscript.py
+
+or (macOS):
+
+    DYLD_INSERT_LIBRARIES=$CONDA_PREFIX/lib/libjemalloc.2.dylib python myscript.py  # untested
+
+.. note::
+
+    The macOS allocator appears to be less prone to fragmentation, so this may
+    be less necessary on Macs.
```
MTCam, line 38: I first noticed the issue on my Mac. On my laptop, the memory spirals out of control and my machine starts swapping, making the tests take a really long time.  This could also contribute to why the mac os CI consistently takes 2x longer than the ubuntu one.

inducer: OIC, good to know! Let me know what you find. Maybe we should remove this note then.

MTCam: :frowning_face:

Well, I could have done something wrong.

inducer: a23f73a, see also https://github.com/illinois-ceesd/mirgecom/issues/212#issuecomment-776939831

inducer:
FYI: I aimed this PR at master instead of #216, since it appears @matthiasdiener has been experimenting there.

PR #247, opened 2021-02-10
ci: test emirge install

matthiasdiener:
We could change the "normal" CI tests to use emirge too, but I would like to observe how this behaves for a week or so before doing that change.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -144,7 +144,7 @@ jobs:
             [[ $(uname) == Linux ]] && sudo apt-get install -y openmpi-bin libopenmpi-dev
             [[ $(uname) == Darwin ]] && brew install mpich
             cd ..
-            git clone https://github.com/illinois-ceesd/emirge
+            git clone --branch add-cantera https://github.com/illinois-ceesd/emirge
             cd emirge
             sed -i.bak '/fetch-mirgecom/d' install.sh
             cp -a ../mirgecom .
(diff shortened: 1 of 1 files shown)


MTCam (commented):
This is a good idea I think.  

If we did make `emirge` the installation method for the tests, then I guess we'd need `--branch-name=` to work properly.  Is there any value to adding a branch-specific install test or so to round it out?

matthiasdiener:
> This is a good idea I think.
> 
> If we did make `emirge` the installation method for the tests, then I guess we'd need `--branch-name=` to work properly. Is there any value to adding a branch-specific install test or so to round it out?

`--branch` won't work currently with forks, which is why this CI test is copying over the checked-out mirgecom files into the emirge folder to use the mirgecom version that is under review. This should work with PRs from branches in this repo as well as forked repos. Unfortunately, there seems to be no easy way to do this without the copying step.

MTCam:
> `--branch` won't work currently with forks, which is why this CI test is copying over the checked-out mirgecom files into the emirge folder to install the mirgecom version that is under review. This should work with PRs from branches in this repo as well as forked repos. Unfortunately, there seems to be no easy way to do this without the copying step.

oic. the copy seems like a good fix for that issue. tested?

matthiasdiener:
> > `--branch` won't work currently with forks, which is why this CI test is copying over the checked-out mirgecom files into the emirge folder to install the mirgecom version that is under review. This should work with PRs from branches in this repo as well as forked repos. Unfortunately, there seems to be no easy way to do this without the copying step.
> 
> oic. the copy seems like a good fix for that issue. tested?

test works with https://github.com/illinois-ceesd/emirge/pull/110

MTCam (commented):
I'm not afraid if you're not 👍

MTCam (approved):
Looks good to me :)

PR #236, opened 2021-02-09
Add Pyrometheus thermochemistry EOS

MTCam:
Supersedes #147

Adds a Pyrometheus-based mixture EOS and chemistry source term interface. It requires the MIRGE-compatible generator in this [Pyrometheus PR](https://github.com/ecisneros8/pyrometheus/pull/7), and it requires Cantera.

Several tests of both the generated Pyrometheus code object, and the Pyrometheus-based EOS interface ensure that both agree very closely with Cantera. Reactions are tested against a homogeneous reactor in Cantera. All tests are operated in the "MIRGE context" - i.e. with DOFArrays, and MIRGE-Com-specific data structures.

PR #248, opened 2021-02-10
Test emirge installation of mirgecom branch

MTCam:
This is one branch that requires different package installations than `master`.  Hopefully this will expose any issues there might be with a local branch when using #247.


.test-conda-env-py3.yml
```diff
@@ -2,6 +2,7 @@ name: test-conda-env-py3
 channels:
 - conda-forge
 - nodefaults
+- cantera/label/dev
```
inducer, line 5: This should no longer be necessary. Cantera is in conda forge proper.

MTCam: This is something I should address in #236

MTCam:
Closing this up.  It was just a tool.

PR #252, opened 2021-02-15
annotate dev packages with --editable

matthiasdiener:
Complements https://github.com/illinois-ceesd/emirge/pull/113

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,3 @@
-# /!\ Careful! Not all aspects of the pip requirement specifier format [1] are supported here.
-# See emirge's parsing code [2] for additional restrictions.
-#
-# [1] https://pip.pypa.io/en/stable/reference/pip_install/#requirement-specifiers
-# [2] https://github.com/illinois-ceesd/emirge/blob/master/parse_requirements.sh
-
 islpy
 mpi4py
 numpy
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,15 +12,16 @@ pyvisfile
 pymetis
 importlib-resources
 
-git+https://github.com/inducer/pymbolic.git#egg=pymbolic
-#git+https://github.com/inducer/pyopencl.git#egg=pyopencl
-git+https://github.com/inducer/loopy.git#egg=loopy
-git+https://github.com/inducer/dagrt.git#egg=dagrt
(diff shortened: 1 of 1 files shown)


inducer:
I don't know that this is a good idea, as it impacts the reusability of mirgecom. What if these packages are already installed? AFAICT, if I were to use this in a virtualenv that already has all those packages, those would get uninstalled and replaced by fresh checkouts under mirgecom.

matthiasdiener:
> I don't know that this is a good idea, as it impacts the reusability of mirgecom. What if these packages are already installed? AFAICT, if I were to use this in a virtualenv that already has all those packages, those would get uninstalled and replaced by fresh checkouts under mirgecom.

Hmm, wouldn't that be the intention, to install the exact versions that mirgecom requires?

inducer:
> Edit: Do you mean this in the context of when someone is installing mirgecom without emirge?

Yes. (That's me, most of the time.)

inducer:
(OTOH, IMO this is minor. I won't oppose it.)

MTCam:
Insomuch as this is designed to support hand-off of installation responsibility of development packages from `emirge` -> `pip` (via https://github.com/illinois-ceesd/emirge/pull/113), I'm all for it.  

I am a little more unsure about the potential subtle issues or unintended consequences that might impact others' use of `emirge` as a result of these changes, including those in https://github.com/illinois-ceesd/emirge/pull/113.

But just the fact that we need to be soo cautious about changing this simple behavior of `emirge` and the potential fallout; to me is an indicator that `emirge` is too complex for a `MIRGE-Com` installation tool, and anything we can do to reduce its complexity would be a good move.  Exporting `requirements.txt` parsing to `pip` is a kind of low-hanging fruit pursuant to that goal.

matthiasdiener:
> I don't know that this is a good idea, as it impacts the reusability of mirgecom. What if these packages are already installed? AFAICT, if I were to use this in a virtualenv that already has all those packages, those would get uninstalled and replaced by fresh checkouts under mirgecom.

I've done a few tests, and afaics the behavior with and without `--editable` is the same: it uninstalls a previously installed package (whether that was installed through git or not), and installs the new version.

MTCam (approved):

inducer:
I can always delete the `--editable`, so I won't oppose this.

PR #253, opened 2021-02-16
remove gmsh from conda

matthiasdiener:
background: the gmsh package doesn't exist for power9, so installs on Lassen fail at the moment. If we need the package for CI in the future, it should probably be installed in the CI scripts.


MTCam (approved):
Yes, please 👍

PR #238, opened 2021-02-09
Add autoignition combustion example.

MTCam:

PR #257, opened 2021-02-22
Refactor checkpoint/simutils (attempt #2)

### LINKED ISSUE #167: Final solution dump after exception can run into "file exists" error

majosm:
@MTCam's comment this morning ([link](https://github.com/illinois-ceesd/mirgecom/pull/238#discussion_r580368331)) reminded me that I had this semi-fixed in an old PR (https://github.com/illinois-ceesd/mirgecom/pull/168). I took another look at it today and decided to make a second attempt at cleaning up the simutils/io code.

This PR makes the following changes:
1) It makes simutils and I/O independent of Euler (I think this will be useful going forward with the wall/multiphysics stuff coming in).
2) It separates the visualization code from the checkpoint call so it can be called independently.
3) It adds visualization output when an exact solution mismatch error occurs.

Unlike the previous attempt, this one doesn't add much extra logic on the driver side.

Thoughts?

(Fixes #167, among other things.)


majosm:
(Tests are expected to fail for the time being because I haven't converted all of the examples over, just pulse-mpi and vortex-mpi. Wanting to get some input before I do the rest.)

MTCam:
> Thoughts?
> 
> (Fixes #167, among other things.)

A few thoughts...

- This does not seem to address the issue that the final dump was intended to address.  What happens when the code is done stepping but there has been no output on the final step?  This happens for reasons other than abnormal stepping exit.  For example, if t_final/(nviz * dt) != 0.
- The amount of code that the user needs to get right in the driver has increased in this proposed approach
- If the actual discretization norms were parallel-enabled, then all ranks should successfully detect the mismatch on any rank, obviating special treatment for this case
- I think our solution for this issue should involve a way to detect and handle rank-specific error states.

majosm:
> * This does not seem to address the issue that the final dump was intended to address.  What happens when the code is done stepping but there has been no output on the final step?  This happens for reasons other than abnormal stepping exit.  For example, if t_final/(nviz * dt) != 0.

I implemented this in 19a929a724c90 but decided to change it back. I'm on the fence about whether this is actually desirable behavior. Seems like it would make visualization more annoying since the dumps would not be at fixed intervals. Any thoughts @anderson2981?

> * The amount of code that the user needs to get right in the driver has increased in this proposed approach

Not significantly (IMO). There are a few more function calls, but the calls themselves are simpler. Monolithic `sim_checkpoint` was doing a lot of unrelated things, and being tied to Euler made it unusable for other types of simulations (heat, wave).

> * If the actual discretization norms were parallel-enabled, then all ranks should successfully detect the mismatch on any rank, obviating special treatment for this case

Sounds like a possibility. Are there any downsides? Is there any reason why someone might want to compute norms separately on each rank?

> * I think our solution for this issue should involve a way to detect and handle rank-specific error states.

Fraught with peril, I think. If the error state is not broadcast immediately and one of the non-erroring ranks enters the next blocking operation, the code will hang. (But if you just mean reporting on which rank the error occurred, that sounds reasonable to me.)

MTCam:
> > * This does not seem to address the issue that the final dump was intended to address.  What happens when the code is done stepping but there has been no output on the final step?  This happens for reasons other than abnormal stepping exit.  For example, if t_final/(nviz * dt) != 0.
> 
> I implemented this in [19a929a](https://github.com/illinois-ceesd/mirgecom/commit/19a929a724c90eb15d5f789543079040affff630) but decided to change it back. I'm on the fence about whether this is actually desirable behavior. Seems like it would make visualization more annoying since the dumps would not be at fixed intervals. Any thoughts @anderson2981?
> 

It is wrong to let the code exit at the end of stepping (regardless of how stepping ends) and not make a dump before exiting (unless, clearly, a dump at that step has already been performed). That is the only thing the original check was intended to do.  If this is not done, then any time since the last regular dump is lost, wasting expensive computing time. Additionally, in the case of a crash, it is best to have a dump at the point of the crash.  

In the case of an out-of-output-band stop, only exactly one dump is out-of-band; the last one.


> > * The amount of code that the user needs to get right in the driver has increased in this proposed approach
> 
> Not significantly (IMO). There are a few more function calls, but the calls themselves are simpler. Monolithic `sim_checkpoint` was doing a lot of unrelated things, and being tied to Euler made it unusable for other types of simulations (heat, wave).

Agreed.  Sim checkpoint was originally designed to be generic; just visualize the user-specified fields.  After changes to how CV and DV work in the code, it now looks specific to the fluid module.  

> 
> > * If the actual discretization norms were parallel-enabled, then all ranks should successfully detect the mismatch on any rank, obviating special treatment for this case
> 
> Sounds like a possibility. Are there any downsides? Is there any reason why someone might want to compute norms separately on each rank?

Not usually. For most simulations like ours, not ever.  We would not currently have any use for a rank-specific field stats.  *location*-specific on-the-other-hand would be nice.  As would being able to have the physical location of the min and max of a field.
 
> 
> > * I think our solution for this issue should involve a way to detect and handle rank-specific error states.
> 
> Fraught with peril, I think. If the error state is not broadcast immediately and one of the non-erroring ranks enters the next blocking operation, the code will hang. (But if you just mean reporting on which rank the error occurred, that sounds reasonable to me.)

It's not as dangerous as you might think, and actually avoids the issue you just described.  It amounts to putting in a reduction of the error code at any point where all the ranks should be done before moving on. For example, at the end of every time step, you can just reduce the error code so that you know whether any rank experienced an error during the step (this is what it used to do).  Once an error is detected on any rank, then the code can exit gracefully.

The hang you just mentioned where one rank goes into an error state and the rest enter into a blocking operation; is what happens when parallel-enabled error checking is *not* done.

mirgecom/euler.py
```diff
@@ -34,6 +34,13 @@
 
 .. autofunction:: get_inviscid_timestep
 .. autofunction:: get_inviscid_cfl
+
+Status Output
+^^^^^^^^^^^^^
+
+.. autofunction:: get_extra_init
+.. autofunction:: get_extra_status
```
MTCam, line 42: What would you think about changing this ~ `_extra_` -> `_custom_` or `_user_`?

mirgecom/mpi.py
```diff
@@ -154,3 +156,11 @@ def wrapped_func(*args, **kwargs):
...
 
     return wrapped_func
+
+
+def comm_any(comm, local_value):
+    """Return True if *local_value* is True on any rank in *comm*."""
+    reduced_value = np.array([int(local_value)])
+    from mpi4py import MPI
+    comm.Allreduce(MPI.IN_PLACE, reduced_value, op=MPI.MAX)
+    return bool(reduced_value[0])
```
MTCam, line 166: This is exactly the facility I was talking about when I said we need the ability to handle rank-specific errors.  An error on any rank can be detected using this snippet.  We should be careful to do collective error detection and handling in other places, too.

mirgecom/euler.py
```diff
@@ -373,3 +380,27 @@ def get_inviscid_timestep(discr, eos, cfl, q):
...
 #    return c*dt_ngf*dt_gf/max_v
+
+
+def get_extra_init(cfl, constant_cfl, initname, eosname):
+    """Get Euler-specific init messages."""
+    return (
+        f"CFL:             {cfl}\n"
+        f"Constant CFL:    {constant_cfl}\n"
+        f"Initialization:  {initname}\n"
+        f"EOS:             {eosname}\n")
```
MTCam, line 391: Related to the previous comment about `_extra_` vs. `_custom_` or similar...

These aren't exactly `Euler`-specific, but fluid-specific. Just so happens that `Euler` is our only flow module atm.  Maybe `get_flow_init`?

We should think of the quantities that would be good to have when a user asks for support to fix their simulation.  The init message should (by default) provide at a glance the essential parameters in the flow setup.  (Might be nice if it would *validate* it too, but that's another can of worms).  These are what I came up with - but there will be a lot more with viscous flows of reactive mixtures.

Some hook to allow user customization of the init message (i.e. to *add* things to it) would be good.  The whole init msg handling is pretty clunk right now, and made a bit better by your changes... but let's go all the way?

mirgecom/euler.py
```diff
@@ -373,3 +380,27 @@ def get_inviscid_timestep(discr, eos, cfl, q):
...
+    # seems potentially misleading...
+    cv = split_conserved(discr.dim, state)
+    dv = eos.dependent_vars(cv)
+    from functools import partial
+    _min = partial(discr.nodal_min, "vol")
+    _max = partial(discr.nodal_max, "vol")
+    return (
+        f"------- {cfl=}\n"
+        f"------- P({_min(dv.pressure):.3g}, {_max(dv.pressure):.3g})\n"
+        f"------- T({_min(dv.temperature):.3g}, {_max(dv.temperature):.3g})")
```
MTCam, line 406: I think we intend to replace this by logging facilities.  Those already address the rank-local issue, and allow customization. 

Health check would still be useful, however.  For example, not just *reporting* min/max of fields, but tossing an error when they stray from an accepted value range.

majosm:
> It is wrong to let the code exit at the end of stepping (regardless of how stepping ends) and not make a dump before exiting (unless, clearly, a dump at that step has already been performed). That is the only thing the original check was intended to do. If this is not done, then any time since the last regular dump is lost, wasting expensive computing time. Additionally, in the case of a crash, it is best to have a dump at the point of the crash.
> 
> In the case of an out-of-output-band stop, only exactly one dump is out-of-band; the last one.

I'm in favor of dumping on a crash. I don't think that would lead to any issues visualization-wise since that final dump will likely be deleted before the next run anyway.

I'm still not really loving the idea of forcing a final dump, though. Tempted to say I prefer the old CM/PC2 style of integrating for a fixed number of steps instead of a fixed amount of time. It avoids having to deal with some of the weirder edge cases. For example, in pulse-mpi, if you force a write at exit you end up with the last 2 dumps being ~1e-16s apart due to rounding error. Bleh 🙁 

> It's not as dangerous as you might think, and actually avoids the issue you just described. It amounts to putting in a reduction of the error code at any point where all the ranks should be done before moving on. For example, at the end of every time step, you can just reduce the error code so that you know whether any rank experienced an error during the step (this is what it used to do). Once an error is detected on any rank, then the code can exit gracefully.
> 
> The hang you just mentioned where one rank goes into an error state and the rest enter into a blocking operation; is what happens when parallel-enabled error checking is _not_ done.

I think for this to work you need to allow all of the ranks to keep going even though some of them have encountered an error (up until they can all synchronize their error state). This is slightly sketchy, but it probably works often enough. Python doesn't typically handle errors this way, though, and I have a feeling we would encounter some push-back if we tried to circumvent this. 🙂

mirgecom/euler.py
majosm: Yeah, I guess what I was thinking is that we would have a pair of these routines for each set of governing equations (Euler, NS, heat) and let the user mix and match based on what they're running. Having even more customization than that would be cool, but I don't really have a clear vision for how to set that up.

MTCam: Sure. I agree with the intent. Euler and NS will mostly overlap however, meaning that every time you use NS one - you need to stack the Euler one and the NS one if you don't intend to repeat code.  That leads me to suggest we have a unified "flow" init msg that just spews the minimal set of flow configuration.

Re: customization - I think we could handle it exactly like the viz customization.  If the user has junk to add, just create one's own `make_status_message`.

MTCam:
> I'm still not really loving the idea of forcing a final dump, though. Tempted to say I prefer the old CM/PC2 style of integrating for a fixed number of steps instead of a fixed amount of time. It avoids having to deal with some of the weirder edge cases. For example, in pulse-mpi, if you force a write at exit you end up with the last 2 dumps being ~1e-16s apart due to rounding error. Bleh 🙁
> 

PC2 performs this last dump. I maintain that if the simulation state has changed before exit (i.e. and since the last dump), then we should record that change with a dump. It is OK imo to have an out-of-band dump at the end.  For most simulation configurations, it won't be triggered. 

Since this capability does no harm, and offers the user the option of being able to specify a stopping criterion based on simulated time, or perhaps convergence, rather than number-of-steps; seems like win-win.  



> > (...) reduction of the error code at any point where all the ranks should be done before moving on.
> 
> I think for this to work you need to allow all of the ranks to keep going even though some of them have encountered an error (up until they can all synchronize their error state). This is slightly sketchy, but it probably works often enough. Python doesn't typically handle errors this way, though, and I have a feeling we would encounter some push-back if we tried to circumvent this. 🙂

Certainly this is true, and (imo) is a good practice for parallel codes. Either we don't let ranks toss exceptions except at certain places, or we toss exceptions that are parallel-aware.  The former approach is the only one I've ever pursued and it is pretty easy to design the code blocks so that the errors are detected.  Upon error, any given rank falls through to the end of the processing block where the error reduction occurs. This is basically the approach you have taken here, I'm only saying we should peruse the code and try to catch the cases where ranks can exit or go into exception without notifying the rest of the parallel world.

anderson2981:
Detecting an error state and exiting/dumping immediately would be a huge help when running simulations. Being able to look at the results and see where an error occurred would be immensely helpful. Here I think the best course of action would be to trigger a dump/exit flag and let the solver finish the current timestep, or substep at least. I agree I feel like it would be difficult/perilous to interrupt the control flow and back all the way out. Having the code hang it way worse than having it just continue merrily along producing garbage.

I think for the timestepping, we need some more advanced ways of setting the timestep.  I can certainly envision scenarios where I want to dump at a fixed time interval (debugging/early runs) and also ones where dumping at fixed step intervals is handy (when I want to maximize throughput). In the later, we want to move towards an adaptive time step that's based on a CFL-like criteria, where the timestep can change. In this case, you would adjust the timesteps leading up to dump/simulation end time to hit them exactly.

majosm:
> PC2 performs this last dump. I maintain that if the simulation state has changed before exit (i.e. and since the last dump), then we should record that change with a dump. It is OK imo to have an out-of-band dump at the end.

With the PC2-style "fixed number of steps" approach you could set the number of steps to be a multiple of the visualization interval to avoid this. I don't think this is realistically possible with the "fixed time interval" approach due to rounding error.

> For most simulation configurations, it won't be triggered.

Won't this *always* be triggered?

MTCam:
> > PC2 performs this last dump. I maintain that if the simulation state has changed before exit (i.e. and since the last dump), then we should record that change with a dump. It is OK imo to have an out-of-band dump at the end.
> 
> With the PC2-style "fixed number of steps" approach you could set the number of steps to be a multiple of the visualization interval to avoid this. I don't think this is realistically possible with the "fixed time interval" approach due to rounding error.
> 
> > For most simulation configurations, it won't be triggered.
> 
> Won't this _always_ be triggered?

Only when the final step was not dumped:

```py
# Check to see if current state has been written
if not (check_step(current_step, ndump)):
  write_final_dump()
```

MTCam:
> > PC2 performs this last dump. I maintain that if the simulation state has changed before exit (i.e. and since the last dump), then we should record that change with a dump. It is OK imo to have an out-of-band dump at the end.
> 
> With the PC2-style "fixed number of steps" approach you could set the number of steps to be a multiple of the visualization interval to avoid this. I don't think this is realistically possible with the "fixed time interval" approach due to rounding error.

In PC2 any of several conditions can cause the stepping to exit.  This same "final_dump" check is performed in PC2.

majosm:
> Only when the final step was not dumped:
> 
> ```python
> # Check to see if current state has been written
> if not (check_step(current_step, ndump)):
>   write_final_dump()
> ```

In principle I'm not against this, but in the current "fixed interval of time" implementation this will almost always be triggered. With rounding error, even if the time interval is a multiple of the timestep, the stepper will generally end up doing a tiny extra final step, e.g. (pulse-mpi):

```
Status: step=10 t=0.09999999999999999 dt=1.3877787807814457e-17
------- cfl=1.0
------- P(1, 1.11)
------- T(0.00348, 0.00419)
Checkpointing final state ...
Status: step=11 t=0.1 dt=0
------- cfl=1.0
------- P(1, 1.11)
------- T(0.00348, 0.00419)
```

MTCam:
> In principle I'm not against this, but in the current "fixed interval of time" implementation this will almost always be triggered. With rounding error, even if the time interval is a multiple of the timestep, the stepper will generally end up doing a tiny extra final step, e.g. (pulse-mpi):

I hadn't actually noticed it was doing this (I'm almost sure I've never actually seen it do this), but I would assert that if your stopping criterion is simulated-time-based, then this is exactly the correct (and desired) behavior.

We need to be able to stop based on nsteps, too. If that wasn't in there already, it was an oversight.

majosm:
> We need to be able to stop based on nsteps, too. If that wasn't in there already, it was an oversight.

majosm/mirgecom#1

mirgecom/euler.py
MTCam: Another quick thought about customization: maybe the init/status msg handling routine should just take an argument that is the user's customized init/status msg function.  Then the msg handling routine could just call that an expect a return string that is the user's customized msg addition.

examples/vortex-mpi.py
```diff
@@ -110,52 +116,55 @@ def main(ctx_factory=cl.create_some_context):
...
+        initname=initializer.__class__.__name__,
+        eosname=eos.__class__.__name__)
+    init_message = make_init_message(dim=dim, order=order, casename=casename,
                                      nelements=local_nelements,
                                      global_nelements=global_nelements,
                                      dt=current_dt, t_final=t_final, nstatus=nstatus,
-                                     nviz=nviz, cfl=current_cfl,
-                                     constant_cfl=constant_cfl, initname=initname,
-                                     eosname=eosname, casename=casename)
+                                     nviz=nviz, extra_init=extra_init)
```
inducer, line 128: In a way, this is all information that should also be recorded in logpyle's constants, so that these run properties are recorded. Maybe this "init message" thing can be combined with/based on that facility?

cc @matthiasdiener

mirgecom/euler.py
inducer: > I think we intend to replace this by logging facilities. Those already address the rank-local issue, and allow customization.

Agree.

> Health check would still be useful, however. For example, not just reporting min/max of fields, but tossing an error when they stray from an accepted value range.

True, though probably driver-specific.

mirgecom/mpi.py
inducer: I'm not sure. I recall we discussed this before, but I don't recall what the resolution was. Currently, my thinking is that we should probably rely on mpi4py to kill the job if one rank exits with an exception.

inducer (commented):
Thanks for working on this. Some thoughts below.

inducer:
Forgot to say: I generally like where this is going.

MTCam:
> True, though probably driver-specific.

FWIW, much of it is driver-independent.  For example, if any of temp, press, or speed of sound go negative, or if specific volume goes ~infinite, we should stop, instead of continuing to compute NANs as we currently do.  These are all common situations which arise frequently, and which we generally don't catch until post-run analysis when a portion (often a majority) of the dumps are all NANs.

A health-check facility would be useful for this.  Allowing customization would also be nice as users could then add any driver-specific health-watches needed.

inducer:

majosm:
I added a few more updates:

1) I merged in the logging changes. I ended up tweaking a few things in `logging_quantities.py` to try to remove that pesky EOS from the interface. I also moved the Euler-specific logging setup to a function inside `euler.py`. cc @matthiasdiener 

2) I reworked the structure of `advance_state` a little to support alternate termination criteria and to fix the final dump issue. Now `checkpoint` is called at the end of each timestep instead of at the beginning, and it returns a boolean value telling the stepper whether it should continue or not. I added an explicit checkpoint call to the drivers before stepping starts to do the initial dump (only if current_step == 0, i.e., not a restart). cc @MTCam 

3) I removed the old status output (now fixes #258).

Still not sure what needs to be done (if anything) about `make_init_message`.

examples/vortex-mpi.py
```diff
@@ -149,54 +150,57 @@ def main(ctx_factory=cl.create_some_context, use_profiling=False, use_logmgr=Fal
...
+
+    def checkpoint(step, t, dt, state):
+        done = t >= t_final
+        exact_state = initializer(nodes, t=t)
+        if comm_any(comm, discr.norm(state - exact_state, np.inf) > exittol):
+            write_vis(step, t, state)
+            raise RuntimeError(f"Exact solution mismatch at step {step}.")
+        if check_step(step, nviz) or done:
+            write_vis(step, t, state)
+        return done
```
MTCam, line 187: I do like how this is shaping up - but I am still concerned that this leaves a lot to the user to get right in every driver.  I don't think I found a default or checkpoint helper. The same seems to be true for the `write_viz` functionality above.  

Can we provide some built-in defaults for these so that the user just needs to "overload" them if they need to customize?

majosm:
More updates:

1. I added handling for timestep computation errors (which don't exist yet, but can occur in the constant CFL case).
2. I added support for fixed-time dump intervals (demonstrated [here](https://github.com/majosm/mirgecom/commit/145aad604c34da4a20eaacaa8f5608b755164dde)).
3. I added back `sim_checkpoint` and a couple of other helper functions to help reduce the amount of driver code.

@MTCam Any more thoughts?

majosm:
Paraphrasing @MTCam from slack discussing (feel free to correct/add more detail): Seems ok; not sure about `get_timestep`. Would also like to have more health-checking capabilities.

What do you think of the current API @inducer?

examples/vortex-mpi.py
```diff
@@ -147,56 +151,57 @@ def main(ctx_factory=cl.create_some_context, use_profiling=False, use_logmgr=Fal
...
 
-    def my_rhs(t, state):
+    def get_timestep(step, t, state):
+        try:
+            dt_max = get_inviscid_timestep(discr=discr, q=state, cfl=current_cfl,
+                eos=eos) if constant_cfl else current_dt
+        except InviscidTimestepError:
+            write_vis(step, t, state)
+            raise
+        return get_sim_timestep(dt_max, t, t_final=t_final)
```
MTCam, line 176: This seems like it leaves too much for the user to get right in order to have timestepping work correctly.  That is my only real concern about this.

MTCam:
>  Would also like to have more health-checking capabilities.
> 

Regarding health checking:

I envision that the user would optionally add "accepted limits" to the logger's `watched quantities`.  So, for example, if "pressure" is watched, then limits could be `[0, 1e8]`. Then if some watched quantity violates the indicated limit, it could get flagged. Maybe we could check it periodically (like at checkpoint time, but ideally at every step).

Setting no limits would be the current behavior... just keep on trucking even though we're computing NANs.

This is just one possible mechanism.

MTCam:
FWIW, `make_init_message` is just what I personally wanted to see at the beginning of every run.  Most users will want to see a summary similar to this because it is typical for users to have many different sets of things they are running.  Having this summary spew out in your stdout is just a nice-to-have.  

It is not required by any means, so we shouldn't spend any time trying to formalize it.

majosm:
Any thoughts on the [current API](https://github.com/illinois-ceesd/mirgecom/blob/a0b64ebee5f5e8f5a16b4ae24e0a924d46ce1927/examples/pulse-mpi.py) (and the [fixed-time visualization variant](https://github.com/majosm/mirgecom/commit/145aad604c34da4a20eaacaa8f5608b755164dde)) @inducer? Does this look ok?

inducer:
Some reactions while I was chatting with @majosm:

- Is `checkpoint` misnamed? (vs. its responsibility of "let the user look at the current state and decide what to do next")
  - Maybe not?
- If something raises, do I get a final checkpoint? (Is that even possible?)
- Having to manually checkpoint step 0 is a bit clunky? Is that avoidable?
- We should make a logger that only logs on rank 0. (by configuration (IMO preferable), or by wrapping it)
- We (effectively) throw around triples (step number, time, state) a lot. Maybe stick them in a dataclass?
- If you have multiple fixed-interval vis cycles going, how will you know which one you've just hit/which one limited the time step size? Maybe `get_sim_timestep` could return a (list of?) things that limited the time step size?

majosm:
> * We (effectively) throw around triples (step number, time, state) a lot. Maybe stick them in a dataclass?

Maybe something like [this](https://github.com/majosm/mirgecom/blob/81659c0125d13be5f278c100401703a182ad40d2/mirgecom/simutil.py#L48-L68)? Kind of redefining the usage of 'state' here, but it does make the driver look nice ([example](https://github.com/majosm/mirgecom/blob/81659c0125d13be5f278c100401703a182ad40d2/examples/pulse-mpi.py)).

> * Having to manually checkpoint step 0 is a bit clunky? Is that avoidable?

The trouble I ran into with this is that if I put the step 0 handling inside `advance_state`, I end up with potentially inconsistent behavior for the `step == 0` vs. `step > 0` cases:

```python
    if istep == 0:
        done = checkpoint(step, t, state)
    else:
        # Have to guess done value because we can't checkpoint on this step a
        # second time
        done = False
```

If `step == 0` I can call `checkpoint` and potentially exit without taking any steps, but if `step > 0` I can't call `checkpoint` (because it has already been called for that step), so I have to assume `done == False` and take at least one step. To avoid this I think I would need to separate the done checking from the rest of the checkpointing somehow. (`checkpoint` would still need to know the value of `done`, though, in order to write the final output.)

inducer:
> Maybe something like [this](https://github.com/majosm/mirgecom/blob/81659c0125d13be5f278c100401703a182ad40d2/mirgecom/simutil.py#L48-L68)? Kind of redefining the usage of 'state' here, but it does make the driver look nice ([example](https://github.com/majosm/mirgecom/blob/81659c0125d13be5f278c100401703a182ad40d2/examples/pulse-mpi.py)).

Not loving `State` as the name. Otherwise yep, that looks great.

> I end up with potentially inconsistent behavior for the step == 0 vs. step > 0 cases:

Maybe a `continuing=True/False` flag?

majosm:
> Not loving `State` as the name. Otherwise yep, that looks great.

Is `<something>State` ok? Or should it not be called state at all?

> Maybe a continuing=True/False flag?

I think maybe I'll just make it an error to call `advance_state` with a checkpoint function that returns true on the initial state:

```python
    if state.step == 0:
        done = checkpoint(state)
        if done:
            raise ValueError("Checkpoint returned 'True' before stepping.")

    done = False
```

MTCam:
> > Not loving `State` as the name. Otherwise yep, that looks great.
> 
> Is `<something>State` ok? Or should it not be called state at all?
> 

Is something like `loop_state` acceptable?


> I think maybe I'll just make it an error to call `advance_state` with a checkpoint function that returns true on the initial state:

We often want to initialize the solution and dump and then take no steps. Can that still be supported?

majosm:
> We often want to initialize the solution and dump and then take no steps. Can that still be supported?

Yeah, this seems to be broken. I think we really do either need to separate the done checking from the rest of the checkpoint or implement multiple 'degrees' of checkpointing so that we can safely call it multiple times.

I made an attempt at the latter approach [here](https://github.com/majosm/mirgecom/commit/7c18f5753d0ac72f8fda644ef011ca68e3b5e2c9), which fixes the zero-steps issue. Not crazy about it, but I still think I prefer something like this over separating out the done checking.

MTCam:
> > We often want to initialize the solution and dump and then take no steps. Can that still be supported?
> Yeah, this seems to be broken. I think we really do either need to separate the done checking from the rest of the checkpoint or implement multiple 'degrees' of checkpointing so that we can safely call it multiple times.

Couldn't the user do this if they wanted it with a driver snippet that goes like:
```
dump_step = 0
if current_step == dump_step:
  dump()
```
... where dump() is just viz writing or even restart writing?  I don't think the viz dump has to absolutely be tied to checkpoint does it?

majosm:
> Couldn't the user do this if they wanted it with a driver snippet that goes like:
> 
> ```
> dump_step = 0
> if current_step == dump_step:
>   dump()
> ```
> 
> ... where dump() is just viz writing or even restart writing? I don't think the viz dump has to absolutely be tied to checkpoint does it?

Yeah, they could just call `write_vis` and comment out `advance_state`. I guess that's good enough?

inducer:
- Not loving "weak" as the name, especially not without defined semantics.
- I wonder if much of this problem space went away if the underlying vis writer offered a mode of "only write this if it doesn't currently exist"?
- As for naming, "state" is tempting, but overused IMO. `TimedSolutionSnapshot`? `StateWithTime`?

@ mention me next time you'd like me to look.

MTCam:
> > Couldn't the user do this if they wanted it with a driver snippet that goes like: <blah blah blah>
> 
> Yeah, they could just call `write_vis` and comment out `advance_state`. I guess that's good enough?

Sure. But I think they wouldn't even need to comment out stepping if t_final = 0.

majosm:
Superseded by #381 (mostly; I have some stuff I'd still like to add from here, but those will come in other PRs).

PR #262, opened 2021-02-25
Discontinuous initializer

anderson2981:
Add initialization routine for a discontinuous field, smoothed by a hyperbolic tangent.


mirgecom/initializers.py
```diff
@@ -923,20 +929,18 @@ def __call__(self, t, x_vec, eos=IdealSingleGas()):
...
-        rhou = mass*u
-        energy = rhoe + 0.5*mass*(u*u)
+        xtanh = 1.0 / sigma * (x_rel - x0)
+        mass = (rhol / 2.0 * (actx.np.tanh(-xtanh) + 1.0)
+              + rhor / 2.0 * (actx.np.tanh(xtanh) + 1.0))
+        rhoe = (rhoel / 2.0 * (actx.np.tanh(-xtanh) + 1.0)
+              + rhoer / 2.0 * (actx.np.tanh(xtanh) + 1.0))
+        u = (ul / 2.0 * (actx.np.tanh(-xtanh) + 1.0)
+           + ur / 2.0 * (actx.np.tanh(xtanh) + 1.0))
+        rhou = mass * u
```
MTCam, line 939:
```suggest
velocity = make_obj_array([u, zeros])
mom = rho * velocity
```

mirgecom/initializers.py
```diff
@@ -923,20 +929,18 @@ def __call__(self, t, x_vec, eos=IdealSingleGas()):
...
-            [
-                0*x_rel
-                for i in range(self._dim)
-            ]
-        )
-        mom[0]=rhou
-
-        return flat_obj_array(mass, energy, mom)
+        mom = make_obj_array([0 * x_rel for i in range(self._dim)])
+        mom[0] = rhou
```
MTCam, line 943:
```suggest
```

mirgecom/initializers.py
```diff
@@ -841,3 +842,103 @@ def __call__(self, x_vec, eos, *, t=0.0):
...
+        rhoer = zeros + self._pr/gm1
+
+        xtanh = 1.0 / sigma * (x_rel - x0)
+        mass = (rhol / 2.0 * (actx.np.tanh(-xtanh) + 1.0)
+              + rhor / 2.0 * (actx.np.tanh(xtanh) + 1.0))
+        rhoe = (rhoel / 2.0 * (actx.np.tanh(-xtanh) + 1.0)
+              + rhoer / 2.0 * (actx.np.tanh(xtanh) + 1.0))
+        u = (ul / 2.0 * (actx.np.tanh(-xtanh) + 1.0)
+           + ur / 2.0 * (actx.np.tanh(xtanh) + 1.0))
+mom = mass * make_obj_array([u, zeros])
```
anderson2981, line 939:
```suggestion
        mom = mass * make_obj_array([u, zeros])
```

mirgecom/initializers.py
```diff
@@ -840,3 +841,149 @@ def __call__(self, x_vec, eos, *, t=0.0):
 
         return join_conserved(dim=self._dim, mass=mass, energy=energy,
                               momentum=mom, species_mass=specmass)
+
+
+class Discontinuity:
+    r"""Solution initializer for flow with a discontinuity.
+
+    This initializer creates a physics-consistent mixture solution
```
MTCam, line 849:
```suggestion
    This initializer creates a physics-consistent flow solution
```

mirgecom/initializers.py
```diff
@@ -840,3 +841,149 @@ def __call__(self, x_vec, eos, *, t=0.0):
...
+    given an initial thermal state (pressure, temperature) and a
+    mixture-compatible EOS.
```
MTCam, line 851:
```suggestion
    given an initial thermal state (pressure, temperature) and an EOS.
```

mirgecom/initializers.py
```diff
@@ -840,3 +841,149 @@ def __call__(self, x_vec, eos, *, t=0.0):
...
+
+    The solution varies across a planar interface defined by a tanh fucntion
```
MTCam, line 853:
```suggestion
    The solution varies across a planar interface defined by a tanh function.
```

mirgecom/initializers.py
```diff
@@ -840,3 +841,149 @@ def __call__(self, x_vec, eos, *, t=0.0):
...
+            mass = eos.get_density(pressure, temperature, y)
+        else:
+            mass = pressure/temperature/eos.gas_const()
+
+        specmass = mass * y
+        mom = mass * velocity
+        if self._nspecies:
+            internal_energy = eos.get_internal_energy(temperature, y)
+        else:
+            internal_energy = pressure/mass/(eos.gamma() - 1)
```
MTCam, line 983: We need to make the EOS interface uniform to avoid branches like these.

mirgecom/initializers.py
```diff
@@ -840,3 +841,149 @@ def __call__(self, x_vec, eos, *, t=0.0):
...
+    .. automethod:: __call__
+    """
+
+    def __init__(
+            self, *, dim=3, xdir=0, x0=0, nspecies=0,
+            tl=300.0, tr=600.0,
+            pl=1.e5, pr=2.e5,
+            ul=None, ur=None,
+            yl=None, yr=None,
+            uc=None, sigma=0.5
```
MTCam, line 866: I _like_ the brevity, but experience indicates we should be more descriptive with parameter names - at least in the actual user-facing interface.  Internally, I think it is less important because only those who (think they) know what they are doing will monkey with it internally.

mirgecom/initializers.py
```diff
@@ -840,3 +841,149 @@ def __call__(self, x_vec, eos, *, t=0.0):
...
+        x_rel = x_vec[self._xdir]
+        actx = x_rel.array_context
+        zeros = 0*x_rel
+        x0 = zeros + self._uc[self._xdir]*t + self._x0
+        ones = zeros + 1
+
+        pl = self._pl*ones
+        tl = self._tl*ones
+        ul = make_obj_array([self._ul[i]*ones for i in range(self._dim)])
+        yl = make_obj_array([self._yl[i]*ones for i in range(self._nspecies)])
```
MTCam, line 958: Real question: Are these needed? Or would something like this suffice:
```
ul = self._ul * ones
```

mirgecom/initializers.py
majosm: I agree. Look at all that nice empty space you have there to fill up. 🙂

mirgecom/initializers.py
```diff
@@ -840,3 +841,149 @@ def __call__(self, x_vec, eos, *, t=0.0):
...
+        yl: numpy.ndarray
+            species mass fractions to the left of the discontinuity
+        pr: float
+            pressure to the right of the discontinuity
+        tr: float
+            temperaure to the right of the discontinuity
+        ur: numpy.ndarray
+            velocity (vector) to the right of the discontinuity
+        yr: numpy.ndarray
+            species mass fractions to the right of the discontinuity
```
majosm, line 893: Seems like this would be a little easier to read if the corresponding `<something>l` and `<something>r` were next to each other?

mirgecom/initializers.py
```diff
@@ -840,3 +841,149 @@ def __call__(self, x_vec, eos, *, t=0.0):
...
+        self._tr = tr
+        self._yl = yl
+        self._yr = yr
+        self._xdir = xdir
+        if self._xdir >= self._dim:
+            self._xdir = self._dim - 1
+
+    def __call__(self, x_vec, eos, *, t=0.0):
+        """
+        Create the mixture state at locations *x_vec* (t is ignored).
```
majosm, line 931:
```suggestion
        Create the mixture state at locations *x_vec* (*t* is ignored).
```

mirgecom/initializers.py
majosm: Actually, it doesn't seem like you would need these arrays at all, maybe? I think the expressions below should work even if `ul`, `ur`, etc. are just floats.

mirgecom/initializers.py
```diff
@@ -840,3 +841,149 @@ def __call__(self, x_vec, eos, *, t=0.0):
 
         return join_conserved(dim=self._dim, mass=mass, energy=energy,
                               momentum=mom, species_mass=specmass)
+
+
+class Discontinuity:
```
majosm, line 846: Is there a more descriptive name we could use? `PlanarDiscontinuity`? `PlanarShock`?

mirgecom/initializers.py
```diff
@@ -840,3 +841,149 @@ def __call__(self, x_vec, eos, *, t=0.0):
...
+    r"""Solution initializer for flow with a discontinuity.
+
+    This initializer creates a physics-consistent mixture solution
+    given an initial thermal state (pressure, temperature) and a
+    mixture-compatible EOS.
+
+    The solution varies across a planar interface defined by a tanh fucntion
+    located at x=xloc for density, velocity, and pressure
```
majosm, line 854:
```suggestion
    located at x=x0 for density, velocity, and pressure
```

mirgecom/initializers.py
anderson2981: PlanarDiscontinuity would be the most descriptive, It's not guaranteed to be a shock, depending on the left and right states.

mirgecom/initializers.py
MTCam: I sort of wonder if they should come in as tuples?

mirgecom/initializers.py
anderson2981: I agree, by default I use left and right, this is obvious when the discontinuity is in the x-direction, but less so in y and z? I'm ok with it if the group is though. I like some directionality over things like t1 and t2 for example. in this case it's not clear which side of the interface is 1 or 2. I could also be talked into plus and minus.

How verbose should/can we be?
temperature_<left/right>, pressure<left/right>, velocity<left/right>, species_mass_<left/right>

I think preserving the shortened names helps readability in the call routine.

mirgecom/initializers.py
anderson2981: I struggle with how to arrange these, it feels natural  to group them by left and right. but grouping by pressure/temperature, etc. makes sense as well.

mirgecom/initializers.py
MTCam: What do we think about something like this?
Tweak to interface: anderson2981/mirgecom#1
Corresponding call site: anderson2981/CEESD-Y1_Nozzle#1

anderson2981: For me as a user, this seems less intuitive. What are we gaining by using tuples? I could see the advantage if it was a common data structure in the code, but this is a weird mix of conserved and dependent variables. If we could agree on a uniform API for all the initializers, then that would be something to consider further I think. We would need to think hard about how the data should be passed in, the needs vary between the different initializers.

MTCam: > For me as a user, this seems less intuitive. What are we gaining by using tuples? I could see the advantage if it was a common data structure in the code, ...

It is a little bit messy having two non-descriptive variables for each left/right thing.  The tuples make the call smaller, and better associate the things that are coming in pairs.  I do agree that the changed interface is harder to "deal with" from both the user and init-developer perspective. 

For this particular init, I think it is fine if the variable names are just changed to be descriptive. In general, it would be nice if we had a natural programming dialect for things that come in pairs; a left and right thing, or a plus and minus thing.  

> ... but this is a weird mix of conserved and dependent variables. If we could agree on a uniform API for all the initializers, then that would be something to consider further I think. We would need to think hard about how the data should be passed in, the needs vary between the different initializers.

I am not sure it is entirely relevant whether the data are conserved and or dependent variables, or a mix of them.  The init routines need to support arbitrary input data, as their data needs are sharply different. That said, if we had a nice way to pass in arbitrary parameters to a routine (like **kwargs) or so, it'd be a snap to unify the init API.  

We should think about what init API unification would really buy us, however.  One would not be able to pass just any parameters willy-nilly and expect it to work, the user will still need to read the init doc (or use an example) to know the data requirements for the init.

anderson2981: I think the goal is that the API should allow the user to look at an existing driver and tell exactly what the input parameters do. The descriptive variables names accomplish that I think. I'll make the change and push something.

mirgecom/initializers.py
MTCam: My vote would be to arrange by physical quantity, not by left vs. right.  Either way is perfectly understandable, but I do agree with @majosm's arrangement suggestion.

mirgecom/initializers.py
```diff
@@ -840,3 +842,136 @@ def __call__(self, x_vec, eos, *, t=0.0):
...
+        pressure_right: float
+            pressure to the right of the discontinuity
+        temperature_right: float
+            temperaure to the right of the discontinuity
+        velocity_right: numpy.ndarray
+            velocity (vector) to the right of the discontinuity
+        species_mass_right: numpy.ndarray
+            species mass fractions to the right of the discontinuity
+        sigma: float
+           sharpness parameter
```
inducer, line 897: Maybe specify the `tanh` blend curve so that the semantics for this can be defined?

mirgecom/initializers.py
```diff
@@ -840,3 +842,136 @@ def __call__(self, x_vec, eos, *, t=0.0):
...
+        species_mass_left: numpy.ndarray
+            species mass fractions to the left of the discontinuity
+        pressure_right: float
+            pressure to the right of the discontinuity
+        temperature_right: float
+            temperaure to the right of the discontinuity
+        velocity_right: numpy.ndarray
+            velocity (vector) to the right of the discontinuity
+        species_mass_right: numpy.ndarray
+            species mass fractions to the right of the discontinuity
```
inducer, line 895: Maybe create a dataclass (or just use `ConservedVars`?) for the left/right states instead of this pile'o'variables?

inducer (commented):
Thanks! A few thoughts below.

inducer:

anderson2981:
@MTCam let's get this into main.

PR #263, opened 2021-03-01
Add the 14-stage, 4th order LSRK method of Niegemann, Diehl, and Busch (2012)

thomasgibson:
Howdy!

I've just installed mirgecom and noticed ya'll have the LSRK 5-stage, 4th order method by Kennedy and Carpenter, so I thought I'd add another LSRK method. The method I have added is a 14-stage, 4th order method summarized in the following paper (Table 3): https://www.sciencedirect.com/science/article/pii/S0021999111005213?via%3Dihub

It possesses an optimized stability region for advection-dominated flows, and is capable of taking much larger time-steps than the RK4 or LSRK4 methods. I've seen nice results in terms of time-to-solution using this 14-stage method in other DG codes I have experimented with, and I thought I would add this to mirgecom too. This hopefully can help speed up the time-to-solution for people running tests and simulations.

As a quick example, I ran the `examples/wave-eager.py` example in 2D, and the 14-stage method can take a larger timestep (6 times larger) than the 5-stage 4th order method. I also rearranged some of the code a bit to perhaps make it a bit neater/organized!

### DIFF
--- a/mirgecom/integrators.py
+++ b/mirgecom/integrators.py
@@ -37,7 +37,8 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-from mirgecom.butcher_tableau import *
+import mirgecom.butcher_tableau as bt
+
 
 def rk4_step(state, t, dt, rhs):
     """Take one step using 4th order Runge-Kutta."""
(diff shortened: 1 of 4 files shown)


mirgecom/integrators.py
```diff
@@ -79,8 +59,25 @@ def lsrk4_step(state, t, dt, rhs):
...
+
+    return p
+
+
+def lsrk144_step(state, t, dt, rhs):
+    """
+    Take one step using the low storage 14-stage 4th order Runge-Kutta method.
+
+    LSRK coefficients are summarized in Table 3 of Niegemann, Diehl, and
+    Busch (2012): https://doi.org/10.1016/j.jcp.2011.09.003.
```
MTCam, line 73: Thanks a ton, Thomas!  Could you please add the reference to `doc/misc.rst` and reference it here?

thomasgibson: Will do!

MTCam (approved):
Lgtm!  👍

inducer (commented):
Whoops. Bit late with my review. Oh well, here goes anyway.

First of all, thanks for working on this! It's interesting and immediately useful.

One additional bit of perspective I'll add is that, possibly, we might engage [leap](https://github.com/inducer/leap) at some later point for "fancier" time integration needs. (also [here](https://documen.tician.de/leap/))

PR #265, opened 2021-03-02
Update mirgecom time-stepping module

thomasgibson:
Howdy!

This PR adds the `integrators` module within mirgecom, which is responsible for all timestepping needs.


mirgecom/timesteppers/lsrk.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def step(self, state, t, dt, rhs):
+        """Take one step using a low-storage Runge Kutta method."""
+
+        k = 0.0
+        for i in range(self.nstages):
+            k = self.A[i]*k + dt*rhs(t + self.C[i]*dt, state)
+            state += self.B[i]*k
+
+        return state
```
majosm, line 29: Seems like you might be able to achieve this without resorting to a class hierarchy, by doing something like:

```python
@dataclass(frozen=True)
class LSRK4Coefs:
    A: np.ndarray
    B: np.ndarray
    C: np.ndarray


def lsrk4_step(coefs, state, t, dt, rhs):
     k = 0.0
     for i in range(len(coefs.A)):
         k = coefs.A[i]*k + dt*rhs(t + coefs.C[i]*dt, state)
         state += coefs.B[i]*k
     return state
```

```python
LSRK54CarpenterKennedyCoefs = LSRK4Coefs(
    A=[...],
    B=[...],
    C=[...])


lsrk4_carpenter_kennedy_step = partial(lsrk4_step, LSRK54CarpenterKennedyCoefs)
```
Thoughts?

majosm (commented):
Added a few thoughts.

mirgecom/timesteppers/lsrk.py
majosm: Could even do:

```python
lsrk54_carpenter_kennedy_step = partial(lsrk_step, LSRKCoefs(
    A=[...],
    B=[...],
    C=[...])
```
🙂

majosm: Then again, if you're going to do that maybe you don't even need the dataclass:

```python
def lsrk_step(A, B, C, state, t, dt, rhs):
    ...
```

```python
lsrk54_carpenter_kennedy_step = partial(lsrk_step, 
    A=[...],
    B=[...],
    C=[...])
```

I'm rambling here, sorry. 🤦

examples/heat-source-mpi.py
```diff
@@ -34,7 +34,7 @@
 from grudge import sym as grudge_sym
 from grudge.shortcuts import make_visualizer
 from grudge.symbolic.primitives import QTAG_NONE
-from mirgecom.integrators import rk4_step
+from mirgecom.timesteppers import RK4Classical
```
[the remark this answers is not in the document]
thomasgibson: Ah, this is a very good point! Having `steppers` and `timesteppers` is indeed confusing. I will revert back to `integrators` to keep it more easily distinguishable.

As for where to put `advance_state`, I think putting it in `simutil.py` could work. What do others think? @inducer @MTCam ?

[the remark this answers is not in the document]
MTCam: I'm not opposed to moving `advance_state`, but I am not sure `simutil.py` is where it should land.  Is there any reason it shouldn't get dropped into the same module as the time integration methods?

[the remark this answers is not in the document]
thomasgibson: @MTCam I just took a gander in `simutil.py` and I'm inclined to agree with you; it seems to be a bit out of place compared to the other utility functions in there, and given the role of `advance_state`, I think it makes sense to move it into the timestepping module. Will update!

mirgecom/timesteppers/lsrk.py
thomasgibson: @majosm I like the pythonics :smiley: I will try out your suggestions and see how it "feels" and "looks", then update! Out of curiosity, are there some subtle disadvantages to using a class hierarchy that I may be overlooking? Perhaps it's a bit unnecessary since the steppers are relatively simple algorithms

examples/heat-source-mpi.py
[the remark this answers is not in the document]
majosm: > I'm not opposed to moving `advance_state`, but I am not sure `simutil.py` is where it should land.

Why not? It's a driver helper function like everything else in there.

> Is there any reason it shouldn't get dropped into the same module as the time integration methods?

`advance_state` sits at a higher level in the mirgecom abstraction hierarchy. I don't think it would make sense to package it together with the integrators.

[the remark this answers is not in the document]
MTCam: > > I'm not opposed to moving `advance_state`, but I am not sure `simutil.py` is where it should land.
> 
> Why not? It's a driver helper function like everything else in there.
> 
> 
> `advance_state` sits at a higher level in the mirgecom abstraction hierarchy. I don't think it would make sense to package it together with the integrators.

I don't disagree with anything you've said here.  My only reluctance to drop it into `simutils.py` is that placing it there would be an _elevation_ of a take-it-or-leave-it, "Random collection-o-tools" as it is documented, module to a higher "status" so-to-say.

mirgecom/timesteppers/lsrk.py
majosm: Mainly just extra boilerplate code I think. Classes make sense when you have a more complicated interface, but for something that is effectively just a function, you might as well make it one. 🙂

examples/heat-source-mpi.py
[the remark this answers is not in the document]
majosm: > I don't disagree with anything you've said here. My only reluctance to drop it into `simutils.py` is that placing it there would be an _elevation_ of a take-it-or-leave-it, "Random collection-o-tools" as it is documented, module to a higher "status" so-to-say.

That's fair. I think if we want to avoid elevating `simutils.py` to that status my preference would be to leave things as they are now with `steppers.py` and `integrators.py`.

mirgecom/timesteppers/lsrk.py
majosm: One potential downside I just noticed to what I suggested above: we won't get a unique docstring for each integrator function. Maybe it would be better to do (with the dataclass again):

```python
LSRK54CarpenterKennedyCoefs = LSRKCoefs(
    A=[...],
    B=[...],
    C=[...])


def lsrk54_carpenter_kennedy_step(state, t, dt, rhs):
    """Blah blah blah."""
    return lsrk_step(LSRK54CarpenterKennedyCoefs, state, t, dt, rhs)
```

thomasgibson: @majosm I decided to go with your dataclass suggestion. With regards to the docstring, would
```python
foo = partial(func, args)
foo.__doc__ = ...
```
Be a viable alternative? I just now saw the discussion after I already pushed some changes. I don't mind updating the function definitions to what you're suggesting here if this is preferrable!

examples/heat-source-mpi.py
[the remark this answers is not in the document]
thomasgibson: Sorry I just now saw this discussion. I moved `advance_state` to the time-stepping (now called `integrators`) module in the file `tsutils.py` (similar to the style `simutils.py`). I don't mind keeping it `steppers.py` if that is preferable. But
```python
from mirgecom.integrators import *
```
will include `advance_state` now, as well as all the timestepping functions.

mirgecom/timesteppers/lsrk.py
majosm: Hmm, maybe -- not sure. @inducer?

thomasgibson: I've decided to go with your suggestion anyways because I think it's more clear!

mirgecom/integrators/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .explicit_rk import *             # noqa: F401, F403
```
thomasgibson, line 1: Are people okay with this? My reasoning here is that all relevant imports through * are determined via `__all__` in the relevant python file. It just makes it so that you don't have to always modify the init when ever someone adds a new timestepper.

examples/heat-source-mpi.py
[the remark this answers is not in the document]
majosm: Yeah, personally I think it'd be best to keep `advance_state` in `steppers`, and let `integrators` remain blissfully unaware that things like drivers, logging, etc. exist.

thomasgibson:
Thank you everyone for your comments. I believe _now_ it's officially ready! Just waiting on CI results to make sure.

examples/lump-mpi.py
```diff
@@ -46,8 +46,7 @@
 from mirgecom.io import make_init_message
 from mirgecom.mpi import mpi_entry_point
 
-from mirgecom.integrators import rk4_step
-from mirgecom.steppers import advance_state
+from mirgecom.integrators import rk4_step, advance_state
```
majosm, line 49: Need to change these back.

mirgecom/integrators/__init__.py
majosm: No idea. 🙂 @inducer

mirgecom/integrators/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+"""
+
+from .explicit_rk import *             # noqa: F401, F403
+from .lsrk import *                    # noqa: F401, F403
+
+__doc__ = """
+.. automodule:: mirgecom.integrators.lsrk
+.. automodule:: mirgecom.integrators.explicit_rk
```
majosm, line 32:
```suggestion
.. automodule:: mirgecom.integrators.explicit_rk
.. automodule:: mirgecom.integrators.lsrk
```
Let's swap these.

majosm (approved):
Apart from the one suggestion, looks good as far as I can tell. We should give @inducer a chance to look before merging though.

mirgecom/integrators/__init__.py
thomasgibson: Done!

mirgecom/integrators/__init__.py
inducer: No, sorry, I don't like `*` imports much. Here, I suspect it's defensible, but in modules containing code, it removes Flake8's ability to figure out what names are defined. It's noisy, but I'd prefer making the list of imported things explicit.

mirgecom/integrators/explicit_rk.py
```diff
@@ -37,50 +27,15 @@
...
 """
 
-import mirgecom.butcher_tableau as bt
+
+__all__ = ("rk4_step", )
 
 
 def rk4_step(state, t, dt, rhs):
-    """Take one step using 4th order Runge-Kutta."""
+    """Take one step using 4th-order Classical Runge-Kutta method."""
```
inducer, line 35:
```suggestion
    """Take one step using the fourth-order Classical Runge-Kutta method."""
```

mirgecom/integrators/lsrk.py
```diff
@@ -0,0 +1,157 @@
...
+    return state
+
+
+LSRKEulerCoefs = LSRKCoefficients(
+    A=np.array([0.]),
+    B=np.array([1.]),
+    C=np.array([0.]))
+
+
+def lsrkeuler_step(state, t, dt, rhs):
```
inducer, line 68: Low-storage Euler, eh? :)

I get the consistency angle here, but it reads funny to me. (vs just `euler`) What do you think?

inducer (approved):
Thanks for working on this!

To clarify protocol: "Approval with comments" to me means, "Provided the comments below are addressed in one way or another (I trust you there), I'm good with merging it and don't need to look at this again."

mirgecom/integrators/lsrk.py
thomasgibson: It is a bit silly looking at it now lol. I just wanted to keep it consistent (`lsrkX_step`). But I am totally happy with keeping it as before (`euler_step`). Though in all honestly, we should NOT be using explicit Euler in production ;)

PR #272, opened 2021-03-08
Add fluid module with resident vel grad function.

MTCam:
Simple module add introduces a generic "fluid" module that will soon encapsulate our operator-independent functions and data structures supporting general CFD. 

The module brings in a primitive function needed to compute the viscous stress tensor, `compute_local_velocity_gradient`, and tests for that function.


mirgecom/fluid.py
```diff
@@ -0,0 +1,62 @@
...
+    -------
+    numpy.ndarray
+        object array of :class:`~meshmode.dof_array.DOFArray`
+        representing $\partial_j{v_i}$.
+    """
+    dim = discr.dim
+    velocity = cv.momentum/cv.mass
+    dmass = discr.grad(cv.mass)
+    dmom = [discr.grad(cv.momentum[i]) for i in range(dim)]
+    return [(dmom[i] - velocity[i]*dmass)/cv.mass for i in range(dim)]
```
inducer, line 62: My one concern with this is that, in the presence of solutions with discontinuities (as you're likely to find with DG), this is not a consistent gradient. Some type of flux term is needed.

MTCam: Sure. Is there an existing `grad` utility that can help me with that, or do I need to add an interface-aware grad?  Is there a grudge-oriented example I could look at?

inducer: To "just move past this", a central flux ("just average both sides") would at least be consistent, but IMO not desirable. Upwinding is a good form of dissipation to have, after all. :) So there's a whole design space here that depends on how you'd like to discretize the overall operator, and the choice may be different for viscous and inviscid terms. (but do note that the volume derivatives can be reused) For perspective, note how local Lax-Friedrichs is just central + a penalty. For second order operators, "unstabilized central", "interior penalty", and "LDG" are the options that are worth considering IMO.

This is DG as a scheme being "not nice" to the code-driven desire to build abstractions (like the one you're proposing here).

Summoning @thomasgibson, maybe he has additional perspective.

cc @lukeolson

MTCam: Just so we're clear, this part is supposed to be only the volume piece evaluated at the nodes of the discretization.  The fluxes I plan on constructing with a symmetric tensor built from this function.

thomasgibson: So just so I understand, is this function just returning the cell-wise gradient of velocity? Meaning that it's not really a "DG gradient"? In which case, this should be made clear in the docstring that this is basically computing `\nabla|_K u`, where `K` is some cell. There are no integrals here?

mirgecom/fluid.py
```diff
@@ -0,0 +1,62 @@
...
+import numpy as np  # noqa
+from meshmode.dof_array import DOFArray  # noqa
+from mirgecom.euler import ConservedVars
+
+
+def compute_local_velocity_gradient(discr, cv: ConservedVars):
+    r"""
+    Compute the cell-local gradient of fluid velocity.
+
+    Computes the cell-local gradient of fluid velocity from:
```
thomasgibson, line 38: Minor comment here! 
```suggestion
    Computes the cell-local gradient of fluid velocity from:
```

MTCam: The trouble is that the first sentence is required by pydocstyle to be one-line sentence in imperative mood.  The detailed description follows.

thomasgibson: Oh sorry I forgot about that. Nevermind then!

mirgecom/fluid.py
```diff
@@ -0,0 +1,62 @@
...
+    -------
+    numpy.ndarray
+        object array of :class:`~meshmode.dof_array.DOFArray`
+        representing $\partial_j{v_i}$.
+    """
+    dim = discr.dim
+    velocity = cv.momentum/cv.mass
+    dmass = discr.grad(cv.mass)
+    dmom = [discr.grad(cv.momentum[i]) for i in range(dim)]
+    return [(dmom[i] - velocity[i]*dmass)/cv.mass for i in range(dim)]
```
majosm, line 62: Doc says it returns a numpy array, but it's currently returning a list.

Could maybe do something along the lines of:

```suggestion
    dmom = np.array([discr.grad(cv.momentum[i]) for i in range(dim)], dtype=object)
    return dmom - np.outer(velocity, dmass)/cv.mass
```
?

mirgecom/fluid.py
```diff
@@ -0,0 +1,62 @@
...
+    ----------
+    discr: grudge.eager.EagerDGDiscretization
+        the discretization to use
+    cv: mirgecom.euler.ConservedVars
+        the fluid conserved variables
+    Returns
+    -------
+    numpy.ndarray
+        object array of :class:`~meshmode.dof_array.DOFArray`
+        representing $\partial_j{v_i}$.
```
majosm, line 56: Should probably specify the structure of the array, since it's not obvious.

test/test_fluid.py
```diff
@@ -0,0 +1,162 @@
...
+    grad_v = compute_local_velocity_gradient(discr, cv)
+    tol = 1e-12
+    for i in range(dim):
+        grad_v_comp = grad_v[i]
+        for j in range(dim):
+            if i == j:
+                exp_value = 1.0
+            else:
+                exp_value = 0.0
+            assert discr.norm(grad_v_comp[j] - exp_value, np.inf) < tol
```
majosm, line 128: Possible to avoid test code duplication by parametrizing the test on `cv` and `expected_grad_v` (or functions that generate them given a discretization, etc.)?

test/test_fluid.py
```diff
@@ -0,0 +1,162 @@
...
+        comp_err = make_obj_array([discr.norm(grad_v[i] - discr.grad(velocity[i]),
+                                              np.inf) for i in range(dim)])
+        max_err = comp_err.max()
+        eoc.add_data_point(h, max_err)
+
+    print(eoc)
+    assert (
+        eoc.order_estimate() >= order - 0.5
+        or eoc.max_error() < 1e-9
+    )
```
majosm, line 162: Maybe make this a separate test.

test/test_fluid.py
```diff
@@ -0,0 +1,162 @@
+"""Test the generic fluid helper functions."""
+
+__copyright__ = """
+Copyright (C) 2020 University of Illinois Board of Trustees
```
majosm, line 4:
```suggestion
Copyright (C) 2021 University of Illinois Board of Trustees
```

mirgecom/fluid.py
MTCam: I had to tweak this slightly back to a clunkier form (ae88854), but it may have just been because the whole array needed to be divided by `cv.mass`. At least the returned structure is right now.  I'll try to bring it back closer to your expression after addressing the other comments.

mirgecom/fluid.py
MTCam: (ae88854)

test/test_fluid.py
MTCam: (f3d0c98)

test/test_fluid.py
MTCam: (f3d0c98)

mirgecom/fluid.py
MTCam: (d7f3dc9)

mirgecom/fluid.py
```diff
@@ -0,0 +1,65 @@
...
+        representing $\partial_j{v_i}$. e.g. for 2D:
+        $\left( \begin{array}{cc}
+        \partial_{x}\mathbf{v}_{x}&\partial_{y}\mathbf{v}_{x} \\
+        \partial_{x}\mathbf{v}_{y}&\partial_{y}\mathbf{v}_{y} \end{array} \right)$
+    """
+    dim = discr.dim
+    velocity = cv.momentum/cv.mass
+    dmass = discr.grad(cv.mass)
+    dmom = np.array([discr.grad(cv.momentum[i]) for i in range(dim)], dtype=object)
+    return (dmom - np.outer(velocity, dmass))/cv.mass
```
inducer, line 65: In addition to the missing surface part, this here also still makes me quite nervous, TBH. As it evaluates a fairly nonlinear function (division) on a non-upsampled grid. (And there's no facility here to get onto one.)

Could you cross-reference this with what is done in the DG book? If they do the same thing, it would make me less nervous.

thomasgibson: @inducer I don't think there needs to be any surface term here because this is just computing a term inside the viscous flux, not a weak divergence or numerical flux. I.e. this is a utility function designed to help construct the flux, which will later be integrated by parts. Then we will run into the question of how to handle the numerical flux (on element boundaries) afterward.

mirgecom/fluid.py
```diff
@@ -0,0 +1,65 @@
...
+        object array of :class:`~meshmode.dof_array.DOFArray`
+        representing $\partial_j{v_i}$. e.g. for 2D:
+        $\left( \begin{array}{cc}
+        \partial_{x}\mathbf{v}_{x}&\partial_{y}\mathbf{v}_{x} \\
+        \partial_{x}\mathbf{v}_{y}&\partial_{y}\mathbf{v}_{y} \end{array} \right)$
+    """
+    dim = discr.dim
+    velocity = cv.momentum/cv.mass
+    dmass = discr.grad(cv.mass)
+    dmom = np.array([discr.grad(cv.momentum[i]) for i in range(dim)], dtype=object)
```
thomasgibson, line 64: Incoming (potentially) dumb question :) Is there a reason we can't just take momentum, divide by density, and then compute the gradient?

mirgecom/fluid.py
MTCam: @thomasgibson is correct about the intended use for this routine. However, maybe I should just remove it from this PR and then re-submit when its use makes the intent apparent.

mirgecom/fluid.py
MTCam: The formula I used follows the form given by Hesthaven and Warburton (section 7.5.2) and a nice feature/advantage is that we are taking the gradient using conserved quantities only, instead of a derived quantity (i.e. velocity). 

If I implement the form of the gradient they indicate, then using the conserved quantities has a distinct advantage since each gradient requires domain boundary data that we cannot have without applying the boundary conditions a-priori.  Unfortunately, Hesthaven says precious little about how to deal with the domain boundary, and what they do say borders on irrelevant for our particular needs.

test/test_fluid.py
```diff
@@ -0,0 +1,138 @@
...
+    assert grad_v_err < tol
+
+
+@pytest.mark.parametrize("dim", [1, 2, 3])
+def test_velocity_gradient_eoc(actx_factory, dim):
+    """Test that the velocity gradient converges at the proper rate."""
+    from mirgecom.fluid import compute_local_velocity_gradient
+    actx = actx_factory()
+
+    nel_1d = 16
```
majosm, line 92:
```suggestion
```

test/test_fluid.py
```diff
@@ -0,0 +1,138 @@
...
+        energy = zeros + 2.5
+
+        mass = nodes[dim-1]*nodes[dim-1]
+        velocity = make_obj_array([actx.np.cos(nodes[i]) for i in range(dim)])
+
+        def exact_grad_row(xdata, gdim, dim):
+            actx = xdata.array_context
+            zeros = 0 * xdata
+            exact_grad_row = make_obj_array([zeros for _ in range(dim)])
+            exact_grad_row[gdim] = actx.np.sin(xdata)
```
majosm, line 121:
```suggestion
            exact_grad_row[gdim] = -actx.np.sin(xdata)
```

test/test_fluid.py
```diff
@@ -0,0 +1,138 @@
...
+            return exact_grad_row
+
+        mom = mass*velocity
+        q = join_conserved(dim, mass=mass, energy=energy, momentum=mom)
+        cv = split_conserved(dim, q)
+        grad_v = compute_local_velocity_gradient(discr, cv)
+        comp_err = make_obj_array([
+            discr.norm(grad_v[i] + exact_grad_row(nodes[i], i, dim), np.inf)
```
majosm, line 129:
```suggestion
            discr.norm(grad_v[i] - exact_grad_row(nodes[i], i, dim), np.inf)
```

mirgecom/fluid.py
majosm: Since I also wondered about this before stumbling upon @thomasgibson's question and your reply here, it might be a good idea to add a comment in the code explaining it.

majosm (commented):
Just a few little things.

mirgecom/fluid.py
MTCam: Updated this function (091b432) so that it is what it ended up being from NS #298.  No more sketchy _local grad_ is used internally. In practice in NS, a _dg grad_ is used as input.
   
@inducer, @thomasgibson

test/test_fluid.py
```diff
@@ -0,0 +1,149 @@
...
+            exact_grad_row[gdim] = -actx.np.sin(xdata)
+            return exact_grad_row
+
+        comp_err = make_obj_array([
+            discr.norm(grad_v[i] - exact_grad_row(nodes[i], i, dim), np.inf)
+            for i in range(dim)])
+        err_max = comp_err.max()
+        eoc.add_data_point(h, err_max)
+
+    print(eoc)
```
thomasgibson, line 145: Do we want print statements in the test?

test/test_fluid.py
```diff
@@ -0,0 +1,149 @@
...
+    assert (
+        eoc.order_estimate() >= order - 0.5
+        or eoc.max_error() < 1e-9
+    )
```
thomasgibson, line 149: I'm actually not sure it makes sense to do a convergence test for this. Because the velocity gradient is not a weak (DG) gradient. In fact, what you're computing is sometimes referred to as the "broken gradient"; that is, the gradient operator restricted to cell volumes.

I think rather than run this convergence test, let's just choose a suitable resolution and just compute the norm of the broken gradient with the exact cell-wise gradient. @inducer @lukeolson Do you agree or disagree?

thomasgibson (commented):
Thanks for this! I think this looks good to me. We'll just note in the future that we may need to modify these functions whenever we add hooks in for over-integration!

I do have a comment about the test though

test/test_fluid.py
MTCam: I agree.  I think the test you suggest already in #298.  I didn't get this test properly updated to reflect the changes in its underlying implementation.

test/test_fluid.py
MTCam: removed

thomasgibson (approved):
Looks good. Thanks!

thomasgibson (approved):

PR #273, opened 2021-03-08
Move ConservedVars to fluid module.

MTCam:
Builds on #272 and moves the operator-independent ConservedVars dataclass to the new fluid module.

PR #275, opened 2021-03-08
Add inviscid module with inviscid-specific functions

MTCam:
Builds on Navier-Stokes refactor in (#272, #273, #274).


majosm (commented):
Do we need to keep both `inviscid.py` and `euler.py`? Why not toss everything into inviscid?

MTCam:
> Do we need to keep both `inviscid.py` and `euler.py`? Why not toss everything into inviscid?

I do kind of like the modular separation between the operator driver and the inviscid flux module, but not wildly so.  

Consistency might demand we include the Navier-Stokes driver under `viscous.py`, and I'm not terribly fond of that.  My thinking was that Euler module uses inviscid flux module, and NS module uses both inviscid and viscous modules.  

If I failed to convince you with that argument, then I am open to moving the Euler operator under inviscid.

mirgecom/inviscid.py
```diff
@@ -0,0 +1,105 @@
...
+
+    # This roughly reproduces the timestep AK used in wave toy
+    dt = (1.0 - 0.25 * (dim - 1)) / (nel_1d * order ** 2)
+    return cfl * dt
+
+#    dt_ngf = dt_non_geometric_factor(discr.mesh)
+#    dt_gf  = dt_geometric_factor(discr.mesh)
+#    wavespeeds = compute_wavespeed(w,eos=eos)
+#    max_v = clmath.max(wavespeeds)
+#    return c*dt_ngf*dt_gf/max_v
```
thomasgibson, line 99: I don't know what's going on here, but I like the commented-out code way more :smile: Are there any issues related to extracting geometric factors and the like? I honestly think this should probably become a higher priority since adaptive time-stepping and CFL monitoring (which @anderson2981 mentioned before) will need these.

@inducer any sense for what is missing / what work is involved to get this information?

thomasgibson (approved):
I left a comment mainly for clarification, but I think that shouldn't stop this from merging right now (since this is just refactoring)

PR #276, opened 2021-03-09
Add flux module, inviscid/Euler refactor

MTCam:
Builds on (#272, #273, #274, #275) - refactors numerical fluxes into flux-specific module.


mirgecom/flux.py
```diff
@@ -0,0 +1,77 @@
+""":mod:`mirgecom.flux` provides inter-facial flux routines.
+
+Numerical Flux Routines
+^^^^^^^^^^^^^^^^^^^^^^^
+.. autofunction:: lfr_flux
+"""
+
+__copyright__ = """
+Copyright (C) 2020 University of Illinois Board of Trustees
```
majosm, line 9:
```suggestion
Copyright (C) 2021 University of Illinois Board of Trustees
```

mirgecom/flux.py
```diff
@@ -0,0 +1,77 @@
...
+    The Lax-Friedrichs/Rusanov flux is calculated as:
+
+    .. math::
+
+        f_{\mathtt{LFR}} = \frac{1}{2}(\mathbf{f}^{+} + \mathbf{f}^{-}) \cdot
+        \hat{n} + \frac{\lambda}{2}(q^{-} - q^{+}),
+
+    where $f^-, f^+$, and $q^-, q^+$ are the fluxes and scalar solution components on
+    the interior and the exterior of the face on which the LFR flux is to be
+    calculated. The The face normal is $\hat{n}$, and $\lambda$ is the user-supplied
```
majosm, line 45:
```suggestion
    calculated. The face normal is $\hat{n}$, and $\lambda$ is the user-supplied
```

mirgecom/flux.py
```diff
@@ -0,0 +1,77 @@
...
+    compute_flux:
+
+        function should return ambient dim-vector fluxes given *q* values
+
+    normal: numpy.ndarray
+
+        object array of :class:`meshmode.dof_array.DOFArray` with outward-pointing
+        normals
+
+    lam: :class:`meshmode.dof_array.DOFArray`
```
majosm, line 64: Necessary to abbreviate? `lambda` is not that long. 🙂

mirgecom/flux.py
```diff
@@ -0,0 +1,77 @@
...
+    where $f^-, f^+$, and $q^-, q^+$ are the fluxes and scalar solution components on
+    the interior and the exterior of the face on which the LFR flux is to be
+    calculated. The The face normal is $\hat{n}$, and $\lambda$ is the user-supplied
+    jump term coefficient.
+
+    Parameters
+    ----------
+    q_tpair:
+
+        Trace pair (grudge.symbolic.TracePair) for the face upon which flux
```
majosm, line 52:
```suggestion
        Trace pair (:class:`grudge.symbolic.TracePair`) for the face upon which flux
```

mirgecom/flux.py
```diff
@@ -0,0 +1,77 @@
...
+
+    lam: :class:`meshmode.dof_array.DOFArray`
+
+        lambda parameter for Lax-Friedrichs/Rusanov flux
+
+    Returns
+    -------
+    numpy.ndarray
+
+        object array of meshmode.dof_array.DOFArray with the Lax-Friedrichs/Rusanov
```
majosm, line 72:
```suggestion
        object array of :class:`meshmode.dof_array.DOFArray` with the Lax-Friedrichs/Rusanov
```

mirgecom/flux.py
MTCam: This one gives me trouble.  Somehow the docs build doesn't find the class ref?

mirgecom/flux.py
MTCam: My editor highlights the keyword "lambda" and that made me fear I was fumbling around with a built-in symbol. No?

majosm: Ahh, right, I forgot about that. Better keep it `lam` then.

mirgecom/flux.py
majosm: Hmm. I forget, do you need to import it first in order to reference it in the docstring?

MTCam: That didn't do it, even after using the proper `grudge.symbolic.primitives.TracePair`.  Odd, the same reference works fine in `euler.py`.

MTCam: It's `grudge.sym.TracePair`

thomasgibson:
I like this a lot. Do you think we should add a central flux function in this module for the viscous terms here? Not necessarily in this PR, but just thinking about NS. Alternatively, we can add central flux functions here and use them in `diffusion.py`

MTCam:
> I like this a lot. Do you think we should add a central flux function in this module for the viscous terms here? Not necessarily in this PR, but just thinking about NS. Alternatively, we can add central flux functions here and use them in `diffusion.py`

Yes, I was hoping we could generalize the various flux routines here.  I do think keeping this PR small is the way to go, then add using additional PRs. There is no good reason not to roll dependent PR(s) that adds flux routines as we need them, however :)

mirgecom/euler.py
```diff
@@ -80,25 +82,18 @@ def _facial_flux(discr, eos, q_tpair, local=False):
...
         compute_wavespeed(dim, eos=eos, cv=split_conserved(dim, q_tpair.int)),
         compute_wavespeed(dim, eos=eos, cv=split_conserved(dim, q_tpair.ext))
     )
-
     normal = thaw(actx, discr.normal(q_tpair.dd))
-    flux_weak = (
-        flux_avg @ normal
-        - 0.5 * lam * (q_tpair.ext - q_tpair.int))
+
+    # todo: user-supplied flux routine
```
thomasgibson, line 95: Should we modify the arguments to `_facial_flux` to accept an incoming flux function? Something like
```python
def _facial_flux(discr, eos, q_tpair, numerical_flux_fct, local=False):
    ...
    kwargs = {
        "compute_flux": euler_flux,
        "normal": normal,
        "lam": lam
    }
    flux_weak = numerical_flux_fct(q_tpair, **kwargs)
    ...
```
This is more or less opening up a pathway to drop in other choices of numerical fluxes for the inviscid terms, such as HLLC or Roe.

MTCam: Yes, but not in *this* PR. That change (i.e. the equivalent one) is coming in a later PR.

mirgecom/flux.py
```diff
@@ -0,0 +1,77 @@
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+"""
+
+
+def lfr_flux(q_tpair, compute_flux, normal, lam):
```
majosm, line 34: It's a little bit confusing to me to see two different `flux`es, where one is a scalar and one is a vector. Would it make sense to absorb the dot product with the normal into `compute_flux`?

MTCam: I'm not sure how to do that here.  Is this true?

```
(f@normal - lam*dq) == (f - lam*dq)@normal
```

mirgecom/euler.py
```diff
@@ -80,25 +82,18 @@ def _facial_flux(discr, eos, q_tpair, local=False):
...
     actx = q_tpair[0].int.array_context
+    dim = discr.dim
 
-    flux_int = inviscid_flux(discr, eos, q_tpair.int)
-    flux_ext = inviscid_flux(discr, eos, q_tpair.ext)
-
-    # Lax-Friedrichs/Rusanov after [Hesthaven_2008]_, Section 6.6
-    flux_avg = 0.5*(flux_int + flux_ext)
-
+    euler_flux = partial(inviscid_flux, discr, eos)
```
majosm, line 88: I meant something more like this:
```suggestion
    def euler_flux(q):
        return inviscid_flux(discr, eos, q) @ normal
```

mirgecom/flux.py
```diff
@@ -0,0 +1,77 @@
...
+    Returns
+    -------
+    numpy.ndarray
+
+        object array of :class:`meshmode.dof_array.DOFArray` with the
+        Lax-Friedrichs/Rusanov flux.
+    """
+    flux_avg = 0.5*(compute_flux(q_tpair.int)
+                    + compute_flux(q_tpair.ext))
+    return flux_avg @ normal - 0.5*lam*(q_tpair.ext - q_tpair.int)
```
majosm, line 77: And
```suggestion
    return flux_avg - 0.5*lam*(q_tpair.ext - q_tpair.int)
```

mirgecom/euler.py
majosm: Though this is also has some vector/scalar ambiguity due to `inviscid_flux`...

mirgecom/flux.py
majosm, line 77: How about:
```suggestion
def lfr_flux(q_tpair, flux_func, normal, lam):
    r"""Compute Lax-Friedrichs/Rusanov flux after [Hesthaven_2008]_, Section 6.6.

    The Lax-Friedrichs/Rusanov flux is calculated as:

    .. math::

        f_{\mathtt{LFR}} = \frac{1}{2}(\mathbf{F}(q^-) + \mathbf{F}(q^+)) \cdot
        \hat{n} + \frac{\lambda}{2}(q^{-} - q^{+}),

    where $q^-, q^+$ are the scalar solution components on the interior and the
    exterior of the face on which the LFR flux is to be calculated, $\mathbf{F}$ is
    the vector flux function, $\hat{n}$ is the face normal, and $\lambda$ is the
    user-supplied jump term coefficient.

    Parameters
    ----------
    flux_func:

        function should return ambient dim-vector fluxes given *q* values

    q_tpair: :class:`grudge.sym.TracePair`

        Trace pair for the face upon which flux calculation is to be performed

    normal: numpy.ndarray

        object array of :class:`meshmode.dof_array.DOFArray` with outward-pointing
        normals

    lam: :class:`meshmode.dof_array.DOFArray`

        lambda parameter for Lax-Friedrichs/Rusanov flux

    Returns
    -------
    numpy.ndarray

        object array of :class:`meshmode.dof_array.DOFArray` with the
        Lax-Friedrichs/Rusanov flux.
    """
    flux_avg = 0.5*(flux_func(q_tpair.int)
                    + flux_func(q_tpair.ext))
    return flux_avg @ normal - 0.5*lam*(q_tpair.ext - q_tpair.int)
```

(Mostly just changing the notation in the documentation, and renaming `compute_flux` to `flux_func` to make it marginally less jarring.)

majosm: (Easier to read version: [link](https://github.com/majosm/mirgecom/commit/cecd3d9e72e71c66146265cab6d6bdcb36bf310c))

thomasgibson (approved):
Looks good!

PR #277, opened 2021-03-09
Rename inviscid_operator -> euler_operator

MTCam:
Renames the inviscid_operator -> euler_operator.  This is part of the fluid refactor to support Navier-Stokes development.  Follows (#272, #273, #274, #275, #276)


thomasgibson (approved):
I personally like inviscid operator a little more, but euler is fine too :smile:

MTCam:
> I personally like inviscid operator a little more, but euler is fine too 😄

I agree; `inviscid_operator` is OK. Then, for consistency, we might be led to rename the Navier-Stokes one to `viscous_operator`, which is also OK.   I considered this briefly (e.g. even renamed `euler_operator` to `inviscid_operator` for a while), but still think that aligning the operator name with the given name of the equations they intend to solve is a superior approach. 

I am open to being convinced otherwise, however :)

PR #285, opened 2021-03-15
Rename parallel mesh widget

### LINKED ISSUE #284: create_parallel_grid is poorly named
- "parallel" is meaningless in this context. Execution in mirgecom is, for the most part, always parallel. "distributed" would be better.
- There is no existing concept called "grid". It returns a meshmode `Mesh`. It should be named for it.
- This also applies to the `generate_grid` argument that this function takes.

Easy fix: Rename, add a compat wrapper with a `DeprecationWarning` and a mid-2021 removal target date.

[Source for reference](https://raw.githubusercontent.com/illinois-ceesd/mirgecom/ba5ae2e0be9f67460b915949b692a80c6fc375ef/examples/boundary-bug.py)

@MTCam, could you take care of that?

MTCam:
Closes #284

### DIFF
--- a/.ci-support/install.sh
+++ b/.ci-support/install.sh
@@ -1,13 +1,15 @@
 if [ "$(uname)" = "Darwin" ]; then
-PLATFORM=MacOSX
-brew install open-mpi
-brew install octave
+    PLATFORM=MacOSX
+    brew install open-mpi
+    brew install octave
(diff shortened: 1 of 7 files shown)


inducer (approved):
Thanks! (I like the new name!)

PR #287, opened 2021-03-15
Make nel_1d mean what it says

majosm:
The `n` parameter in `generate_regular_rect_mesh` is the number of points in 1D, so `nel_1d` is currently kind of misnamed. This PR fixes it.

### DIFF
--- a/.ci-support/install.sh
+++ b/.ci-support/install.sh
@@ -1,11 +1,13 @@
 if [ "$(uname)" = "Darwin" ]; then
 PLATFORM=MacOSX
+brew update
+brew upgrade
 brew install open-mpi
 brew install octave
 else
 PLATFORM=Linux
 sudo apt-get update
-sudo apt-get -y install openmpi-bin libopenmpi-dev
+sudo apt-get -y install libmpich-dev mpich
 sudo apt-get -y install octave
 fi
 MINIFORGE_INSTALL_DIR=.miniforge3
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -139,7 +139,7 @@ jobs:
         - name: Install emirge
           run: |
             [[ $(uname) == Linux ]] && sudo apt-get install -y openmpi-bin libopenmpi-dev
-            [[ $(uname) == Darwin ]] && brew install mpich
+            [[ $(uname) == Darwin ]] && brew update && brew install mpich
             cd ..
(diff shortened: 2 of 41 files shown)


MTCam:
A thousand thumbs up on this, but: shouldn't it be renamed to `npts_1d` instead of left as `nel_1d` just to be corrected at the call sites where it's used?

inducer:
Whoops! I feel (at least partly) responsible here. Thanks @majosm for cleaning this up! (Could you do the same in the grudge and meshmode examples, if that's not asking too much?) FWIW I agree with @MTCam that renaming the variable to `npts_1d` is probably better.

majosm:
I still like thinking of the problem size in terms of numbers of elements instead of points. It makes it easier to multiply the mesh resolution. Also makes code like this simpler (from #272):

```diff
 nel_1d_0 = 5
 for hn1 in [1, 2, 3, 4]:

-    nel_1d = hn1 * (nel_1d_0 - 1) + 1
-    h = 1/(nel_1d-1)
+    nel_1d = hn1 * nel_1d_0
+    h = 1/nel_1d
```

If the `+1`s are awkward, could we possibly add an `nel=` parameter to `generate_regular_rect_mesh` (and then maybe deprecate `n=`?)?

inducer:
> I still like thinking of the problem size in terms of numbers of elements instead of points.

Actually, that's a fair point.

> If the +1s are awkward, could we possibly add an `nel=` parameter to `generate_regular_rect_mesh` (and then maybe deprecate `n=`?)?

I like that idea! We *could* (not saying we should) offer `npts=` and `nel=` to make the interface clearer. Or just `nel`. It's not like it's hard to convert.

MTCam:
> > I still like thinking of the problem size in terms of numbers of elements instead of points.
> 
> Actually, that's a fair point.
> 
> > If the +1s are awkward, could we possibly add an `nel=` parameter to `generate_regular_rect_mesh` (and then maybe deprecate `n=`?)?
> 
> I like that idea! We _could_ (not saying we should) offer `npts=` and `nel=` to make the interface clearer. Or just `nel`. It's not like it's hard to convert.

+1's to everything here (even if it's awkward).  I also prefer to think about number of elements, and would love it to have this distinction acknowledged by the interface.

MTCam (approved):
Still not a giant fan of that it changes the previous grids, but still prefer it over the previous thing.

inducer (approved):
Thanks!

PR #293, opened 2021-03-19
set pocl cache dir

### LINKED ISSUE #292: Parallel runs often fail to spawn on Lassen
This issue has presented itself on Lassen, and is pretty severe lately.

Upon spawning parallel jobs on lassen with something like:
> jsrun -g 1 -a 1 -n 32 python -O -u -m mpi4py nozzle.py

The job often fails during setup (before stepping) with this message:
`CUDA_ERROR_FILE_NOT_FOUND: file not found`

The error is typically hidden in several pages of output from grudge, meshmode, and loopy:
```
< ... pages of similar messages ... >
grudge_assign_0: preprocess done
grudge_assign_0: preprocess start
grudge_assign_0: preprocess done
CUDA_ERROR_FILE_NOT_FOUND: file not found
grudge_assign_0: preprocess start
grudge_assign_0: preprocess start
grudge_assign_0: preprocess start
< ... more similar messages ... >
ERROR:  One or more process (first noticed rank 3) terminated with signal 6
```

The issue is more troublesome at higher node counts.  For the example given above, the job had to be attempted 13 times before it finally ran properly without failing due to this message. 

For now, I just want to submit this issue in case someone has seen this, or has any ideas.  I do not know if the issue is limited to the Y1 Nozzle case (I can't imagine that it is), but I have not tested it for other cases.

matthiasdiener:
Fixes error of the type:
CUDA_ERROR_FILE_NOT_FOUND: file not found

Fixes #292

### DIFF
--- a/examples/lassen.bsub.sh
+++ b/examples/lassen.bsub.sh
@@ -28,6 +28,9 @@ jsrun_cmd="jsrun -g 1 -a 1 -n $nproc"
 # on why this is important
 export XDG_CACHE_HOME="/tmp/$USER/xdg-scratch"
 
+# Fixes https://github.com/illinois-ceesd/mirgecom/issues/292
+export POCL_CACHE_DIR="/tmp/$USER/pocl-cache"
+
 # Print task allocation
 $jsrun_cmd js_task_info
 
--- a/examples/lassen.bsub.sh
+++ b/examples/lassen.bsub.sh
@@ -29,6 +29,7 @@ jsrun_cmd="jsrun -g 1 -a 1 -n $nproc"
 export XDG_CACHE_HOME="/tmp/$USER/xdg-scratch"
 
 # Fixes https://github.com/illinois-ceesd/mirgecom/issues/292
+# (each rank needs its own POCL cache dir)
 export POCL_CACHE_DIR_ROOT="/tmp/$USER/pocl-cache"
 
 # Print task allocation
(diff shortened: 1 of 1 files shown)


MTCam (approved):
Outstanding. Thanks a bunch! ( cc @anderson2981 )

examples/lassen.bsub.sh
```diff
@@ -28,6 +28,9 @@ jsrun_cmd="jsrun -g 1 -a 1 -n $nproc"
 # on why this is important
 export XDG_CACHE_HOME="/tmp/$USER/xdg-scratch"
 
+# Fixes https://github.com/illinois-ceesd/mirgecom/issues/292
```
inducer, line 31: Could you add a comment on why this addresses the issue? Is Nvidia's locking code so bad that one rank's cache updates trample on what another expects to find?

MTCam: Also, I spoke too soon (shock).   It does not seem to sufficiently fix the issue.  We're still iterating on it.

matthiasdiener: I think it's the pocl code that trips over itself. It turns out this doesn't fix the issue completely, multiple ranks on a single node still cause failures.

matthiasdiener: > Could you add a comment on why this addresses the issue? Is Nvidia's locking code so bad that one rank's cache updates trample on what another expects to find?

(Kindof) added something.

inducer (approved):

MTCam (approved):
This does appear to fix the issue for me. :)

PR #296, opened 2021-03-29
Add option to return q from diffusion_operator

majosm:
Just something I added while investigating discontinuous coefficients. Seems vaguely useful.


MTCam (approved):
👍

PR #297, opened 2021-03-29
Add support for discontinuous alpha in diffusion_operator

### LINKED ISSUE #291: Diffusion operator doesn't handle discontinuous diffusivity
Making an issue to organize the discussion about this.

majosm:
Redefines `q` to be `grad(u)` instead of `sqrt(alpha)*grad(u)`. I opted to just call it `grad_u` instead of `q`. Seems a little more readable that way.

Depends on #296 and inducer/grudge#60. Closes #291.


inducer (approved):
LGTM. Thanks for working on this!

PR #298, opened 2021-03-29
Navier-Stokes monster

MTCam:
This is a stab at NS in MIRGE-Com.  This PR is not meant to merge.  It is here for discussion of the NS code design, and depends on outstanding developments in PRs (#272 - #277) + PR #236.


mirgecom/flux.py
```diff
@@ -29,6 +29,46 @@
...
+        object array of `meshmode.dof_array.DOFArray` with the central scalar flux
+        for each scalar component.
+    """
+    tp_avg = trace_pair.avg
+    ncomp = 1
+    if isinstance(tp_avg, np.ndarray):
+        ncomp = len(tp_avg)
+    if ncomp > 1:
+        return make_obj_array([tp_avg[i]*normal for i in range(ncomp)])
+    return trace_pair.avg*normal
```
inducer, line 71:
```suggestion
    tp_avg = trace_pair.avg
    if isinstance(tp_avg, np.ndarray):
        tp_avg = tp_avg.reshape(-1, 1)
    return tp_avg*normal
```

MTCam:
(3cdc3e2) Can be backed out after inducer/grudge#70.

examples/nsmix-mpi.py
```diff
@@ -56,7 +56,7 @@
 from mirgecom.steppers import advance_state
 from mirgecom.boundary import (  # noqa
     AdiabaticSlipBoundary,
-    IsothermalNoSlip,
+    IsothermalNoSlipBoundary,
```
MTCam, line 59: Yeah, I had this before, but replaced it in favor of the version without the "Boundary" designation.  Do we really want to put "Boundary" on the end of every boundary condition?  It is OK by me if this is your preference, but I was sort of leaning the other way with it (i.e. removing the "Boundary" designation.)

majosm:
So I thought about the BC interface stuff for a while last night, and here's what I came up with. Rather than trying to create a single all-encompassing API for BCs, I would propose letting each operator define its own BC interface, e.g.:

```python
# navierstokes.py

class NSBoundary(metaclass=abc.ABCMeta):
    @abc.abstractmethod
    def get_ns_flux(self, discr, btag, eos, q, grad_q, grad_t, **kwargs):
        raise NotImplementedError
```

```python
# artificial_viscosity.py

class AVBoundary(metaclass=abc.ABCMeta):
    @abc.abstractmethod
    def get_av_gradient_flux(self, discr, dd, alpha, q, **kwargs):
        raise NotImplementedError

    @abc.abstractmethod
    def get_av_flux(self, discr, dd, alpha, grad_q, **kwargs):
        raise NotImplementedError
```

Then when you create a boundary condition, inherit/implement accordingly for compatible operators:

```python
# boundary.py

from navierstokes import NSBoundary
from artificial_viscosity import AVBoundary

class IsothermalNoSlipBoundary(NSBoundary, AVBoundary):
    def get_ns_flux(self, discr, btag, eos, q, grad_q, grad_t, **kwargs):
        ...

    def get_av_gradient_flux(self, discr, dd, alpha, q, **kwargs):
        ...

    def get_av_flux(self, discr, dd, alpha, grad_q, **kwargs):
        ...
```

I would actually be tempted to take this one step further and try to implement the BCs inside the corresponding operator's module when possible (in order to keep related code/numerics together). Something like:

```python
# navierstokes.py

class NSIsothermalNoSlipBoundary(NSBoundary):
    def get_ns_flux(self, discr, btag, eos, q, grad_q, grad_t, **kwargs):
        ...
```

```python
# artificial_viscosity.py

class AVDirichletBoundary(AVBoundary):
    def get_av_gradient_flux(self, discr, dd, alpha, q, **kwargs):
        ...

    def get_av_flux(self, discr, dd, alpha, grad_q, **kwargs):
        ...

class AVNeumannBoundary(AVBoundary):
    def get_av_gradient_flux(self, discr, dd, alpha, q, **kwargs):
        ...

    def get_av_flux(self, discr, dd, alpha, grad_q, **kwargs):
        ...

# BC that wraps a list of scalar AV dirichlet/neumann/whatever boundaries
class AVAggregateBoundary(AVBoundary):
    ...
```

```python
# boundary.py

from navierstokes import NSIsothermalNoSlipBoundary
from artificial_viscosity import (
    AVDirichletBoundary,
    AVNeumannBoundary,
    AVAggregateBoundary,
)

class IsothermalNoSlipBoundary(NSIsothermalNoSlipBoundary, AVAggregateBoundary):
    def __init__():
        component_av_boundaries = [...]
        super(IsothermalNoSlipBoundary, self).__init__(component_av_boundaries)
```
I'm not super familiar with inheritance in Python, so I'm not sure what the pitfalls of an approach like this would be. @inducer?

inducer:
@majosm I think as far as inheritance is concerned, the scheme you describe would work just fine.

One question: Are these BCs actually mathematically different? I.e. if the BC prescribed the "external state" rather than the flux, would that permit us to avoid the distinction between AV and NS?

(This is one factor why I'm starting to view BCs returning fluxes (as opposed to external states) in a not-so-favorable light. Another factor is that it violates separation of concerns: The PDE operator specifies which Riemann solver is in use, and if the BC returns a flux, it ultimately overrides that choice.)

majosm:
> One question: Are these BCs actually mathematically different? I.e. if the BC prescribed the "external state" rather than the flux, would that permit us to avoid the distinction between AV and NS?

I think there are still some differences in terms of _what_ external state is needed. For example, in addition to the external `q`, I think NS also needs external temperature (@MTCam?). And AV needs external `grad_q`.

inducer:
> I think there are still some differences in terms of what external state is needed.

Ah, fair point! But the *shared* bits of external state should agree though? If not, I'd be scratching my head a bit.

> I just don't know if it's possible to construct an external state for every possible BC?

I don't know either, but I think it's possible to try. FWIW, if we have a flux, it shouldn't be super hard to reverse-engineer an external state, perhaps using sympy.

FWIW, I've also added this to the agenda for Friday morning. (But I'm not saying that to end the discussion here!)

MTCam:
> So I thought about the BC interface stuff for a while last night, and here's what I came up with. Rather than trying to create a single all-encompassing API for BCs, I would propose letting each operator define its own BC interface, e.g.:
> 
> ```python
> # navierstokes.py
> class NSBoundary(metaclass=abc.ABCMeta):
>     @abc.abstractmethod
>     def get_ns_flux(self, discr, btag, eos, q, grad_q, grad_t, **kwargs):
> # artificial_viscosity.py 
> class AVBoundary(metaclass=abc.ABCMeta):
>     @abc.abstractmethod
>     def get_av_gradient_flux(self, discr, dd, alpha, q, **kwargs):
>     @abc.abstractmethod
>     def get_av_flux(self, discr, dd, alpha, grad_q, **kwargs):
> ```

I like this. I am using an ABC in NS and Euler here (#298 and #332), but my abstract interface includes the calls needed by all of Euler, NS, and AV.  If I understand correctly, this approach would be naturally more clear which API calls go with which operators.  

> Then when you create a boundary condition, inherit/implement accordingly for compatible operators:
> 
> ```python
> # boundary.py
> from navierstokes import NSBoundary
> from artificial_viscosity import AVBoundary
> class IsothermalNoSlipBoundary(NSBoundary, AVBoundary):
>     def get_ns_flux(self, discr, btag, eos, q, grad_q, grad_t, **kwargs):
>     def get_av_gradient_flux(self, discr, dd, alpha, q, **kwargs):
>     def get_av_flux(self, discr, dd, alpha, grad_q, **kwargs):
> ```
> 

Your proposed approach is not terribly different than what we are currently doing, except for the explicit split/definition of AV and NS boundaries.  Each BC (including IsothermalNoSlip) just implements the calls needed by each operator (AV, NS, and Euler).  It had been made more complicated by that interface defined by IsothermalNoSlip (and other BCs) is not the same between AV, Euler, and Navier-Stokes operators, but should be unified now when NS lands. 

> I would actually be tempted to take this one step further and try to implement the BCs inside the corresponding operator's module when possible (in order to keep related code/numerics together)....

I like the things you are proposing, but I have some difficulties:

- Fluid domain boundaries have 2 major boundary considerations, inviscid, and viscous, and splitting the specification for those across two operators (i.e. Euler, and Navier-Stokes) is quite awkward. Perhaps we could have a unified "fluid" boundary construct to deal with that.

- Boundary treatments are more naturally thought of as domain-specific than they are operator-specific.  The operator must dictate the API, sure, but the *values* or the content of the boundary data must be consistently dictated by the domain in which the operator is used. If I use multiple operators in a fluid-domain, for example AV and Euler, the boundary treatments need to be consistent with the inviscid *fluid* boundaries.  The inviscid fluid BC calculations result in the component flux or even Q+, and those must be used by the AV boundary treatments; they are not independent. If we implement the boundaries in each operator, then AV, for example needs to re-implement all the fluid boundary treatments for the AV-specific calls.

- Duplicate functionality.  This is related to the previous point, but a different aspect of it.  All of AV, Diffusion, and Navier-Stokes require `grad(Q)` to do their work.  We can obviously get at that many ways, and as I've noted above, getting the value of Q+, for example, must be consistent with the fluid domain boundaries. All of these operators can easily share the `get_q_flux`, or `get_external_q` (routine names are example-only).   

I think whatever we come up with for boundary treatment unification should include a way for operator-specific boundary treatments to piggy-back off of the fluid-domain boundaries. Perhaps that's just it; maybe our boundary treatment abstractions should be specific to the fluid domain, and then the operator-specific boundary objects can inherit from, and use the fluid boundary objects.

majosm:
> Ah, fair point! But the _shared_ bits of external state should agree though? If not, I'd be scratching my head a bit.

Could there be differences resulting from different overintegration schemes? Though I guess you could probably just pass in the quadrature tag...

MTCam:
> > I think there are still some differences in terms of what external state is needed.
> 
> Ah, fair point! But the _shared_ bits of external state should agree though? If not, I'd be scratching my head a bit.

Yes @majosm and yes @inducer. The boundary data needed are disparate (and sometimes the same!) across the operators (e.g. some need Q+, some need grad(Q)+), but also need to be consistent across all the operators/boundary treatments.

> 
> > I just don't know if it's possible to construct an external state for every possible BC?
> 
> I don't know either, but I think it's possible to try. FWIW, if we have a flux, it shouldn't be super hard to reverse-engineer an external state, perhaps using sympy.
> 
> FWIW, I've also added this to the agenda for Friday morning. (But I'm not saying that to end the discussion here!)

I love thinking about all the boundaries being able to provide (Q+), or whatever external state is required.  It would make everything so much easier!  This notion is inconsistent with the physics, however, and this bears itself out in the literature (e.g. see BR1).  For many physical boundaries, the model dictates that we prescribe (or calculate) (Q+) for some components, while we prescribe or calculate *flux* for other solution components.  For all operators (that we currently have in MIRGE-Com), the final thing that is actually used to integrate the element is *flux*.

inducer:
> This notion is inconsistent with the physics, however, and this bears itself out in the literature (e.g. see BR1).

Could you spell this out a bit more? I'm not sure I totally follow.

> Could there be differences resulting from different overintegration schemes? Though I guess you could probably just pass in the quadrature tag...

I agree that the BC should just adapt to the discretization prescribed by the operator.

MTCam:
> > This notion is inconsistent with the physics, however, and this bears itself out in the literature (e.g. see BR1).
> 
> Could you spell this out a bit more? I'm not sure I totally follow.
> 

For a wall boundary, for example, one want to prescribe the fluxes for density and species masses, and the values for momenta. An interface that does not allow us to specify these fluxes, require us to craft a Q+ that will result in the desired flux. For (subsonic) in/outflow boundaries, we can use Q- to construct Q+ and use the numerical flux function to get the fluxes. 

The varied situations are covered in a very cursory manner by BR1, and more completely by other references such as Poinsot and Lele, Cook, and [this Toulopoulos paper](https://www.sciencedirect.com/science/article/pii/S0021999111002324).

majosm:
> * Fluid domain boundaries have 2 major boundary considerations, inviscid, and viscous, and splitting the specification for those across two operators (i.e. Euler, and Navier-Stokes) is quite awkward. Perhaps we could have a unified "fluid" boundary construct to deal with that.

Would it be possible to implement the NS BCs by using the Euler BCs internally, similar to what I'm doing with AV/diffusion [here](https://github.com/majosm/mirgecom/commit/4220b1a7831ccac4e005054840e61c1cda7c5d0f)?

> * Boundary treatments are more naturally thought of as domain-specific than they are operator-specific.  The operator must dictate the API, sure, but the _values_ or the content of the boundary data must be consistently dictated by the domain in which the operator is used. If I use multiple operators in a fluid-domain, for example AV and Euler, the boundary treatments need to be consistent with the inviscid _fluid_ boundaries.  The inviscid fluid BC calculations result in the component flux or even Q+, and those must be used by the AV boundary treatments; they are not independent. If we implement the boundaries in each operator, then AV, for example needs to re-implement all the fluid boundary treatments for the AV-specific calls.

Right. I'm picturing the boundaries that live in `boundary.py` as the "real" boundaries, and the interfaces are just giving them a way to provide the needed fluxes (or external values) to each operator in a way that's consistent with the boundary condition that they represent.

> * Duplicate functionality.  This is related to the previous point, but a different aspect of it.  All of AV, Diffusion, and Navier-Stokes require `grad(Q)` to do their work.  We can obviously get at that many ways, and as I've noted above, getting the value of Q+, for example, must be consistent with the fluid domain boundaries. All of these operators can easily share the `get_q_flux`, or `get_external_q` (routine names are example-only).

I don't see it as being _too_ egregious to just do something along the lines of:

```python
class ExampleBoundary(Op1Boundary, Op2Boundary):
    def _get_external_q(self, discr, dd, q, **kwargs):
        ...

    def get_op1_external_q(self, discr, dd, q, **kwargs):
        return _get_external_q(self, discr, dd, q, **kwargs)

    def get_op2_external_q(self, discr, dd, q, **kwargs):
        return _get_external_q(self, discr, dd, q, **kwargs)
```

but maybe there's a way to combine them, I'm not sure.

> I think whatever we come up with for boundary treatment unification should include a way for operator-specific boundary treatments to piggy-back off of the fluid-domain boundaries. Perhaps that's just it; maybe our boundary treatment abstractions should be specific to the fluid domain, and then the operator-specific boundary objects can inherit from, and use the fluid boundary objects.

I would prefer not to make things fluid-centric. It would be nice if we could make a single convention for BC handling that extends to the wall part of the simulation as well.

MTCam:
> > * Fluid domain boundaries have 2 major boundary considerations, inviscid, and viscous, and splitting the specification for those across two operators (i.e. Euler, and Navier-Stokes) is quite awkward. Perhaps we could have a unified "fluid" boundary construct to deal with that.
> 
> Would it be possible to implement the NS BCs by using the Euler BCs internally, similar to what I'm doing with AV/diffusion [here](https://github.com/majosm/mirgecom/commit/4220b1a7831ccac4e005054840e61c1cda7c5d0f)?
> 

Yes, I do think so.  My Python-fu just isn't such that a viable solution jumped out at me.

> > * Boundary treatments are more naturally thought of as domain-specific than they are operator-specific.  (...)
> 
> Right. I'm picturing the boundaries that live in `boundary.py` as the "real" boundaries, and the interfaces are just giving them a way to provide the needed fluxes (or external values) to each operator in a way that's consistent with the boundary condition that they represent.
> 

Ah, in that case I may have misinterpreted the implications of the proposed approach.

> > * Duplicate functionality.  (...)
> 
> I don't see it as being _too_ egregious to just do something along the lines of:
> 
> ```python
> class ExampleBoundary(Op1Boundary, Op2Boundary):
>     def _get_external_q(self, discr, dd, q, **kwargs):
>     def get_op1_external_q(self, discr, dd, q, **kwargs):
>         return _get_external_q(self, discr, dd, q, **kwargs)
>     def get_op2_external_q(self, discr, dd, q, **kwargs):
>         return _get_external_q(self, discr, dd, q, **kwargs)
> ```
> 
> but maybe there's a way to combine them, I'm not sure.
> 

Agreed. I think this would be just fine.  

> > I think whatever we come up with for boundary treatment unification should include a way for operator-specific boundary treatments to piggy-back off of the fluid-domain boundaries. (...)
> 
> I would prefer not to make things fluid-centric. It would be nice if we could make a single convention for BC handling that extends to the wall part of the simulation as well.

Yikes. This I don't follow.  Something in there *must* implement the actual fluid boundary treatments.  I am not suggesting that we tie the *interface* itself to fluids-only, only that we implement the actual domain-specific boundary treatments somewhere, and then use your proposed (or similar) approach to let the operator-specific boundary constructs piggy-back the domain-specific boundary treatment implementations.   If I understand correctly, then I think this is what you are already suggesting.

majosm:
> I would actually be tempted to take this one step further and try to implement the BCs inside the corresponding operator's module when possible (in order to keep related code/numerics together). Something like:
> 
> ```python
> # navierstokes.py
> 
> class NSIsothermalNoSlipBoundary(NSBoundary):
>     def get_ns_flux(self, discr, btag, eos, q, grad_q, grad_t, **kwargs):
>         ...
> ```
> 
> ```python
> # artificial_viscosity.py
> 
> class AVDirichletBoundary(AVBoundary):
>     def get_av_gradient_flux(self, discr, dd, alpha, q, **kwargs):
>         ...
> 
>     def get_av_flux(self, discr, dd, alpha, grad_q, **kwargs):
>         ...
> 
> class AVNeumannBoundary(AVBoundary):
>     def get_av_gradient_flux(self, discr, dd, alpha, q, **kwargs):
>         ...
> 
>     def get_av_flux(self, discr, dd, alpha, grad_q, **kwargs):
>         ...
> 
> # BC that wraps a list of scalar AV dirichlet/neumann/whatever boundaries
> class AVAggregateBoundary(AVBoundary):
>     ...
> ```
> 
> ```python
> # boundary.py
> 
> from navierstokes import NSIsothermalNoSlipBoundary
> from artificial_viscosity import (
>     AVDirichletBoundary,
>     AVNeumannBoundary,
>     AVAggregateBoundary,
> )
> 
> class IsothermalNoSlipBoundary(NSIsothermalNoSlipBoundary, AVAggregateBoundary):
>     def __init__():
>         component_av_boundaries = [...]
>         super(IsothermalNoSlipBoundary, self).__init__(component_av_boundaries)
> ```

I've changed my mind a little about the second part of my original suggestion (quoted above). I still like the interface idea, but as @MTCam pointed out the boundary conditions are more related to the overall governing equations and domain than to the specific operators, so it might be preferable to keep their implementations separate instead of putting them in the operator modules like I suggested above.

> Yikes. This I don't follow. Something in there must implement the actual fluid boundary treatments. I am not suggesting that we tie the interface itself to fluids-only, only that we implement the actual domain-specific boundary treatments somewhere, and then use your proposed (or similar) approach to let the operator-specific boundary constructs piggy-back the domain-specific boundary treatment implementations. If I understand correctly, then I think this is what you are already suggesting.

I'm still not 100% sure I'm understanding what you're describing, but: I think I prefer the current approach of defining the boundary condition code in `boundary.py`. The operators are kind of a step lower in the hierarchy than the governing equations / domain, so they shouldn't know too much about the overall problem (e.g., the diffusion operator shouldn't know whether it's part of a heat equation, a Helmholtz equation, etc.). The operator can just provide a standardized way for a BC to tell it what to do at the boundaries, and it doesn't need to know more than that.

majosm:
Also, I'm open to calling them something other than `<operatorname>Boundary` if that's confusing. `<operatorname>BoundaryInterface`? `<operatorname>BoundarySpecifier`? Hopefully someone can come up with something catchier than these. 🙂

majosm:
So I started trying to implement the `AVBoundary` interface in the fluid BCs [here](https://github.com/majosm/mirgecom/commit/93242d8237763dbcd85ddf94f3837d4d4af5a334). I ran into a snag though -- I need to get the indicator values when computing the boundary fluxes. I'll look at it more tomorrow and see if there's a sensible way to to do that.

MTCam:
> So I started trying to implement the `AVBoundary` interface in the fluid BCs [here](https://github.com/majosm/mirgecom/commit/93242d8237763dbcd85ddf94f3837d4d4af5a334). I ran into a snag though -- I need to get the indicator values when computing the boundary fluxes. I'll look at it more tomorrow and see if there's a sensible way to to do that.

Thanks so much for your attention on this.  I feel like this will shape up into something much better than what we have.

Re: your snag... did you have a look at the current AV boundary treatments, (e.g. [this one for slip boundary](https://github.com/w-hagen/mirgecom/blob/326caf8a6dfd579f4692862a31d5cd4b87ed0bf0/mirgecom/boundary.py#L170))? The other BCs are relevant too; they are just simpler.  These are already working in inviscid flows, and they do not directly access the smoothness indicator as a separate construct.  How is he getting by without it?

MTCam:
> Also, I'm open to calling them something other than `<operatorname>Boundary` if that's confusing. `<operatorname>BoundaryInterface`? `<operatorname>BoundarySpecifier`? Hopefully someone can come up with something catchier than these. 🙂

I vote `operatorNameBoundaryInterface`.  I think just Boundary is fine too, but BoundaryInterface is more specific and almost instantly understood by the reader.

MTCam:
> Also, I'm open to calling them something other than `<operatorname>Boundary` if that's confusing. `<operatorname>BoundaryInterface`? `<operatorname>BoundarySpecifier`? Hopefully someone can come up with something catchier than these. 🙂

It may be a little premature to think about this but..

If each operator implements its own `<operator>BoundaryInterface`, are we not left open to whatever willy-nilly BoundaryInterface at the operator developer's whim?   It may be prudent to roll first a generic `BoundaryDataInterface` base class or so that operator-specific boundary interfaces need to derive from.

majosm:
[Here](https://github.com/majosm/mirgecom/pull/2/files) is a demo of how I'm currently picturing the boundary interfaces working (implemented for AV + diffusion).

To unify this with the fluid operators, we would need to add something like `EulerBoundaryInterface` and `NSBoundaryInterface`:

```python
class EulerBoundaryInterface(metaclass=abc.ABCMeta):
    """
    Interface for Euler boundary information retrieval.

    .. automethod:: get_euler_flux
    """

    @abc.abstractmethod
    def get_euler_flux(self, ...):
        """Compute the numerical boundary flux for the Euler operator."""
        raise NotImplementedError


class NSBoundaryInterface(metaclass=abc.ABCMeta):
    """
    Interface for Navier-Stokes boundary information retrieval.

    .. automethod:: get_ns_gradient_flux
    .. automethod:: get_ns_inviscid_flux
    .. automethod:: get_ns_viscous_flux
    """

    @abc.abstractmethod
    def get_ns_gradient_flux(self, ...):
        r"""Compute the numerical boundary flux for $\nabla \mathbf{Q}$."""
        raise NotImplementedError

    @abc.abstractmethod
    def get_ns_inviscid_flux(self, ...):
        """Compute the numerical boundary flux for the inviscid part of the NS operator."""
        raise NotImplementedError

    @abc.abstractmethod
    def get_ns_viscous_flux(self, ...):
        """Compute the numerical boundary flux for the viscous part of the NS operator."""
        raise NotImplementedError
```

Does this seem reasonable? (Just the general approach I mean; we can work out the details later.) if so, how do we want to go about getting this functionality into main? I don't remember exactly what we agreed on as far as ordering of PR merges. Was it something like: #332, then Euler boundary interface, then AV + diffusion, then NS?

MTCam:
> 
> > Lemme take another crack at this. Delaying further action until #298 lands is unacceptable. It is now my intention to take in your suggestions, roll them into #298, and then amend #332 such that it includes your vision for how to "do" inter-operator boundary treatments. Then we can center our discussion around #332 (or a similar new PR). Does that make sense?
> 
> 👍

Ok, so it was a little hard to figure out exactly what to do without having an example where I have mixed operators - so i started by trying to apply the new vision for boundary interfaces to the AV+NS branch.  The AV+NS PR is #338, and then the boundary stuff is applied on top of that in #341.  If we want to, we can also shift to the diffusion-based AV.

mirgecom/flux.py
MTCam: Overcome by developments

PR #300, opened 2021-04-01
Modernize array operations to be compatible with all array contexts

kaushikcfd:


inducer:
Did you mean this as draft for now? (since CI is unhappy)

kaushikcfd:
> since CI is unhappy

I think it's because the target branch's CI is also failing

majosm:
Closure reason: This was based on `y1_production`, which was out of date. We need to make a new one based off of `y1-production`.

PR #313, opened 2021-04-08
Run flake8 in Github CI on target Py version

inducer:


inducer:
Huh, interesting. It found something!

MTCam:
> Huh, interesting. It found something!

Any ideas on how to fix?  It just says `Invalid syntax`.

inducer:
You can use
```
conda create -n somename python=3.6
```
to make a conda env that has Python 3.6.

This says:
```
  File "<fstring>", line 1
    (step=)
         ^
SyntaxError: invalid syntax
```
for `mirgecom/io.py` and similarly for the `{something=}` f-string in the autoignition example.

The bigger question here is, should we pursue 3.6 compatibility, or should we bump to 3.8, which likes these files just fine?

inducer:
Looks like emirge install is currently broken. @matthiasdiener, could you take a look?

inducer:
@matthiasdiener Thanks for taking a look!

matthiasdiener (approved):

inducer:
No objections so far, so I say let's just make it official. (We can always change back, with, so far, limited effort.)

PR #324, opened 2021-04-27
Add modal filtering capability to Mirgecom

thomasgibson:
This PR revives all the hard work by @MTCam in #92 (and also supersedes #92). Uses some recent functionality in meshmode and grudge.

### DIFF
--- a/mirgecom/filter.py
+++ b/mirgecom/filter.py
@@ -161,8 +161,10 @@ def apply_linear_operator(discr, operator, fields):
     return result
 
 
-def create_group_filter_operator(group, filter_mat):
+def create_group_filter_operator(group, cutoff, response_func):
     """Create spectral filter operator for *group*."""
+    filter_mat = make_spectral_filter(group, cutoff,
+                                      mode_response_function=response_func)
     from modepy import vandermonde
     vander = vandermonde(group.basis(), group.unit_nodes)
     vanderm1 = np.linalg.inv(vander)
@@ -171,7 +173,7 @@ def create_group_filter_operator(group, filter_mat):
 
 
 @obj_array_vectorize
-def filter_modally(discrwb, dd, field):
+def filter_modally(discrwb, dd, field, cutoff, mode_resp_func):
     """Stand-alone procedural interface to spectral filtering."""
     dd = sym.as_dofdesc(dd)
     discr = discrwb.discr_from_dd(dd)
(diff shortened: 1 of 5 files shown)


mirgecom/filter.py
```diff
@@ -0,0 +1,200 @@
+""":mod:`mirgecom.filter` is for filters and filter-related constructs.
```
inducer, line 1: This should have an `.. automodule::` somewhere so that it's included in the docs.

mirgecom/filter.py
```diff
@@ -0,0 +1,200 @@
...
+            mode = sum(mode_id)
+            if mode >= cutoff:
+                filter_mat[mode_index, mode_index] = \
+                    mode_response_function(mode,
+                                           cutoff=cutoff,
+                                           nfilt=nfilt)
+
+        return actx.freeze(actx.from_numpy(filter_mat))
+
+    return _spectral_filter_matrix(group)
```
inducer, line 104: This does way more flops than needed. (i.e. no need for a matrix if you're just wanting to scale stuff)

inducer (commented):
Thanks! Two quick notes here.

mirgecom/filter.py
thomasgibson: @inducer Ah, right. So how about we just build a scaling vector? Then that will simplify the `apply_filter_matrix` method (won't actually do a matvec, just element-wise multiply)

inducer: Yep, exactly.

mirgecom/filter.py
thomasgibson: Done: https://github.com/illinois-ceesd/mirgecom/pull/324/commits/b6002f806d01c7a0df9b2b1775c2d81044005301

mirgecom/filter.py
thomasgibson: Done: https://github.com/illinois-ceesd/mirgecom/pull/324/commits/ae92e70c602bcd084e4b034287883de953be0f85

mirgecom/filter.py
```diff
@@ -0,0 +1,212 @@
+""":mod:`mirgecom.filter` is for filters and filter-related constructs.
+
+Discussion of the spectral filter design can be found in:
+JSH/TW Nodal DG Methods (DOI: 10.1007/978-0-387-72067-8), Section 5.3
```
MTCam, line 4:
```suggestion
[Hesthaven_2009]_, Section. 5.3
```

MTCam:
Thanks a whole bunch for breathing new life into this, btw!

inducer (approved):
Thanks! This looks great.

Issue #139, opened 2020-11-05
Need restart


MTCam:
A restart capability will be required in order to run sufficiently long enough for meaningful flow simulations.  We will need these capabilities:

- [ ] recording the conserved quantities, time, step number, and some user-defined dependent variables (e.g. temperature) for every point on the discretization for the purpose of restart
- [ ] reading the data create in the previous step into simulation data structures
- [ ] restarting the simulation with the data read from previous step
- [ ] verification the same advanced state is simulated at step N + M regardless of intermediate restarts
- [ ] Optionally restarting with a different partitioning (e.g. different number of MPI ranks)

inducer:
> Optionally restarting with a different partitioning (e.g. different number of MPI ranks)

I'd vote that this is out of scope for round one. It would take a while to implement, especially if we're not willing to centralize the mesh and DOF data on a single rank.

>  recording [...] some user-defined dependent variables (e.g. temperature)

Why do we need to save these?

> verification the same advanced state is simulated at step N + M regardless of intermediate restarts

For multi-step time integration (i.e. not us, yet), this would entail also saving a good chunk of time stepper state (vs. just re-bootstrapping the time stepper). How important is this "exact restart"?

inducer:
#140

MTCam:
> > Optionally restarting with a different partitioning (e.g. different number of MPI ranks)
> 
> I'd vote that this is out of scope for round one. It would take a while to implement, especially if we're not willing to centralize the mesh and DOF data on a single rank.

Totally agree.  It would be good to leave this on the radar, however.  We need to handle changing resource availability for production runs.  Even for lead-up science runs; consider the situation where a big resource is used to run several flow-throughs, then a much smaller resource is used to run many "shots" or ignition instances.

> 
> > recording [...] some user-defined dependent variables (e.g. temperature)
> 
> Why do we need to save these?
> 

That's a great question.  We can discuss it with JBF, Anderson, and Esteban  - perhaps we can do this better (or we already have) - but here is the issue stated in a meta sort of way:

Currently the temperature (T) is calculated as a function of state, and the last temperature (i.e. T = temperature(state, Tguess)). 

[...]

PR #326, opened 2021-04-28
Update doc for ConservedVars dataclass

MTCam:
Start on and discuss better documentation for fluid data handling in *MIRGE-Com*.

### DIFF
--- a/mirgecom/fluid.py
+++ b/mirgecom/fluid.py
@@ -128,12 +128,20 @@ class ConservedVars:
             fluid_momentum_density = fluid_cv.momentum  # ndim-vector obj array
             fluid_species_mass_density = fluid_cv.species_mass  # nspecies-vector
 
+        Examples of using `ConservedVars` as in this example in:
+
+        - :mod:`~mirgecom.boundary`
+        - :mod:`~mirgecom.euler`
+        - :mod:`~mirgecom.initializers`
+        - :mod:`~mirgecom.simutil`
+
     :example::
 
         Use `join_conserved` to create an agglomerated $\mathbf{Q}$ array from the
         fluid conserved quantities (CV).
 
-        See the first example for details about CV, and $\mathbf{Q}$.
+        See the first example for the definition of CV, $\mathbf{Q}$, ndim,
+        nspecies, and $N_\mbox{eq}$.
 
         Often, a user starts with the fluid conserved quantities like mass and
         energy densities, and it is desired to glom those quantities together into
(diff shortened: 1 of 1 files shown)


inducer:
It occurred to me that if `ConservedVars` supported just enough arithmetic, we could just use it as a state type and never even use the state-vector business. That might also help ferry temperature for restart (#139, #188). Possibly related: https://github.com/inducer/meshmode/pull/139.

MTCam:
> It occurred to me that if `ConservedVars` supported just enough arithmetic, we could just use it as a state type and never even use the state-vector business. That might also help ferry temperature for restart (#139, #188). Possibly related: [inducer/meshmode#139](https://github.com/inducer/meshmode/pull/139).

In this case, it becomes something else, not a way to represent data that corresponds to the equations we are solving, so we'd need to rename it.

In general, I am not a giant fan of combining the two things - the conserved state, or quantities directly corresponding to the conserved state, vs. dependent quantities that we compute from the conserved state.  These are two separate things, and combining them on equal footing does not seem proper to me. (See e.g. PlasCom2)

Additionally, I think that having the dependent quantities ejected from the main data array has naturally forced some nice code constructions and make it clear *when and how* we compute things like dependent quantities.  If, on the other hand, the dependent quantities are embedded in the state array, it creates a situation where users are likely to do stuff like `project("vol", face, fv.temperature)`, for example, and this is super wrong.  Temperature, and many other dependent quantities *need* to be computed with the projected state, instead.   Having these quantities separated from the state forces you to be observant of where the data comes from.

mirgecom/fluid.py
```diff
@@ -46,35 +46,172 @@
...
+
+        Often, a user starts with the fluid conserved quantities like mass and
+        energy densities, and it is desired to glom those quantities together into
+        a *MIRGE*-compatible $\mathbf{Q}$ data structure.
+
+        For example, a solution initialization routine may set the fluid
+        quantities::
+
+            den = rho  # rho is a DOFArray with fluid density
+            velo = v  # v is an obj array of ndim DOFArrays.
```
thomasgibson, line 154:
```suggestion
            velo = v  # v is an obj array of ndim DOFArrays denoting the components of velocity.
```

thomasgibson (commented):
This is terrific @MTCam! Thank you for doing this. Just a super minor suggestion, but otherwise I'm quite happy with this! Tag me again once it's no longer a draft and I'm happy to give it my approval

mirgecom/fluid.py
MTCam: I also wanted more text here, but ran into the line-length limit.  I will commit this and see if the linter screams at me.

MTCam: Yeah, it complained, and I could not figure out how to make the line continue inside the code block.   See how this kind-of-broken-english description grabs you (72fda48).

MTCam:
> It occurred to me that if `ConservedVars` supported just enough arithmetic, we could just use it as a state type and never even use the state-vector business. That might also help ferry temperature for restart (#139, #188). Possibly related: [inducer/meshmode#139](https://github.com/inducer/meshmode/pull/139).

I'm re-thinking my previous knee-jerk response.  As you note, seems like if the ConservedVars (or StateThing) supported enough operations, then we'd never need to be flipping back-and-forth between "Q" and "CV", giving us a way to have a persistent "CV" object (or StateThing).

We had already been discussing, and toying around with memoizing some DV quantities onto the "StateThing" - and if i understand correctly, having a persistent StateThing would actually be **required** for that type of memoization/caching to be effective.

thomasgibson (approved):

mirgecom/fluid.py
```diff
@@ -46,35 +46,172 @@
...
+        a *MIRGE*-compatible $\mathbf{Q}$ data structure.
+
+        For example, a solution initialization routine may set the fluid
+        quantities::
+
+            den = rho  # rho is a DOFArray with fluid density
+            velo = v  # v is an ndim-vector of DOFArray with components of velocity
+            ener = e  # e is a DOFArray with fluid energy
+            mom_dens = den*velo
+            ener_dens = den*ener
```
majosm, line 157:
```suggestion
             rho = ... # rho is a DOFArray with fluid density
             v = ...  # v is an ndim-vector of DOFArray with components of velocity
             e = ...  # e is a DOFArray with fluid energy
```

mirgecom/fluid.py
```diff
@@ -46,35 +46,172 @@
...
+
+        An agglomerated array of fluid independent variables can then be
+        created with::
+
+            q = join_conserved(ndim, mass=den, energy=ener_dens, momentum=mom_dens)
```
majosm, line 162:
```suggestion
            q = join_conserved(ndim, mass=rho, energy=rho*e, momentum=rho*v)
```

majosm (commented):
I'm too slow, but here are some minor formatting/wording suggestions.

PR #330, opened 2021-04-30
Add array context retriever function

majosm:
Adds a universal-ish array context retrieval function to avoid having to do gymnastics like [this]( https://github.com/illinois-ceesd/mirgecom/blob/ea3dc359abbd2e8f852b4f60ff4a77dbec3e0ac2/mirgecom/artificial_viscosity.py#L126-L130).

### DIFF
--- a/mirgecom/diffusion.py
+++ b/mirgecom/diffusion.py
@@ -41,11 +41,12 @@ from meshmode.dof_array import thaw
 from grudge.dof_desc import DOFDesc, as_dofdesc
 from grudge.eager import interior_trace_pair, cross_rank_trace_pairs
 from grudge.symbolic.primitives import TracePair
+from mirgecom.utils import get_actx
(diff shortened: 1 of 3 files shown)


inducer:
I like this idea. I think it could live in meshmode, perhaps coming in with https://github.com/inducer/meshmode/pull/139/. (If you think of it, all these things are "array containers" in a way.)

majosm:
> I like this idea. I think it could live in meshmode, perhaps coming in with [inducer/meshmode#139](https://github.com/inducer/meshmode/pull/139). (If you think of it, all these things are "array containers" in a way.)

Looks like it [already exists](https://github.com/inducer/meshmode/blob/6226f0ec0f0d3deac24f2c35ebcaf79888fe2abf/meshmode/array_context.py#L120-L129). 🙂

PR #331, opened 2021-04-30
Add alternate outer product function

majosm:
Adds a variation on `np.outer` that handles scalar arguments more gracefully in certain situations. For example:

```python
flux = outer(q, normal)
```

returns a vector if `q` is a scalar, and a matrix if `q` is a vector (whereas `np.outer` always seems to return a matrix).

Originally added in w-hagen/mirgecom#19; requested by @MTCam to be added to main for NS.


MTCam (approved):
Thanks!  lgtm

mirgecom/utils.py
```diff
@@ -113,3 +113,18 @@ def min(self) -> Optional[float]:
...
+
+    Tweaks the behavior of :function:`numpy.outer` to return a lower-dimensional
+    object if either/both of *a* and *b* are scalars (whereas
+    :function:`numpy.outer` always returns a matrix).
+    """
+    import numpy as np
+    if isinstance(a, np.ndarray) and isinstance(b, np.ndarray):
+        return np.outer(a, b)
+    else:
+        return a*b
```
inducer, line 130: I think this'll do the wrong thing when applied to `DOFArray * DOFArray`.

majosm: How come? Isn't it just scalar multiplication when they're both `DOFArray`s?

inducer: Ah, you're right. You wouldn't want a 2D `DOFArray`. There aren't that many groups. :) Still, the `else` seems liable to pick up unwanted cases. I propose we wait until https://github.com/inducer/meshmode/pull/175 is in and make the dispatch to `a*b` specific to those containers.

MTCam: @majosm and @inducer Can we wake up this PR? I am hoping it will help me clean up junk like this:

https://github.com/illinois-ceesd/mirgecom/blob/7759e7d968fb8628e978fbb938cf2d1c356ba1ec/mirgecom/flux.py#L73

and this:

https://github.com/illinois-ceesd/mirgecom/blob/7759e7d968fb8628e978fbb938cf2d1c356ba1ec/mirgecom/viscous.py#L86

inducer: Sure, I'm all for this. The array container code is now in, so that should make things easier. One thing to consider is what should happen if the two arguments are not alike in their container-ness. Imagine `a` is something that *contains* an object array, while `b` is an object array itself, when probably the desired thing would be an outer product of the two object arrays.

majosm: Am I doing this right? @inducer

inducer: Honestly: I have a hard time telling, mostly because I'm not confident I understand all the corner cases of the desired behavior. Could you add some tests that delineate the behavior you'd like to see?

majosm: > Honestly: I have a hard time telling, mostly because I'm not confident I understand all the corner cases of the desired behavior. Could you add some tests that delineate the behavior you'd like to see?

https://github.com/illinois-ceesd/mirgecom/pull/331/commits/c1d69deb0a991e5cfaba72090be3c8991160278f

Any cases I'm missing @MTCam?

MTCam: The ones you have there are a great start, and I think they cover my use-cases. So... remember that when creating CV the case where `species_mass=None` actually creates an empty array to stick there. This empty array has made some of the handling a little cumbersome, but I can't recall the exact case where I struggled with it.   I'll be looking around for that...

inducer:
Likely being superseded by https://github.com/inducer/arraycontext/pull/46.

PR #332, opened 2021-05-03
Rearrange the boundary calls to match upcoming Navier-Stokes boundaries.

MTCam:
This is a picked-apart refactor of boundary treatments to match development coming down from NS. The intent is to unify the fluid-specific boundary treatments between Euler, Navier-Stokes, and artificial viscosity (if used with with a fluid).

### DIFF
--- a/mirgecom/boundary.py
+++ b/mirgecom/boundary.py
@@ -3,7 +3,8 @@
 Boundary Conditions
 ^^^^^^^^^^^^^^^^^^^
 
-.. autoclass:: PrescribedBoundary
+.. autoclass:: FluidBoundary
+.. autoclass:: PrescribedInviscidBoundary
 .. autoclass:: DummyBoundary
 .. autoclass:: AdiabaticSlipBoundary
 """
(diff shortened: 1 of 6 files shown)


mirgecom/boundary.py
```diff
@@ -35,16 +36,72 @@
...
-    """Boundary condition prescribes boundary soln with user-specified function.
+class FluidBoundary:
+    r"""Abstract interface to fluid boundary treatment.
+
+    .. automethod:: inviscid_boundary_flux
+    """
+
+    def inviscid_boundary_flux(self, discr, btag, q, eos, **kwargs):
+        """Get the inviscid flux across the boundary faces."""
+        raise NotImplementedError()
```
thomasgibson, line 52: If we want to use abstract interfaces (which I'm all for!) Then I'd suggest using [metaclasses via abc](https://docs.python.org/3/library/abc.html) So this would become:
```suggestion
from abc import ABCMeta, abstractmethod


class FluidBoundary(metaclass=ABCMeta):
    r"""Abstract interface to fluid boundary treatment.

    .. automethod:: inviscid_boundary_flux
    """

    @abstractmethod
    def inviscid_boundary_flux(self, discr, btag, q, eos, **kwargs):
        """Get the inviscid flux across the boundary faces."""
```

The advantage of this approach is that, when adding addition derived classes, objects will fail to instantiate if they do not implement the necessary abstract methods/properties. You're more or less establishing a contract will all derived classes.

MTCam: Yes, this is great.  Thanks!

PR #336, opened 2021-05-05
Enable flake8-bugbear

inducer:
That's what the world needs! More linters! :P To be fair, this one seems quite useful.

https://github.com/PyCQA/flake8-bugbear

Requesting reviews from @MTCam and @thomasgibson because the linter pointed out loops that didn't touch their loop variables, and I've simply removed those loops. Could you guys check that I haven't maimed any logic that we'd like to keep?

### DIFF
--- a/mirgecom/initializers.py
+++ b/mirgecom/initializers.py
@@ -113,7 +113,7 @@ class Vortex2D:
     """
 
     def __init__(
-        self, *, beta=5, center=[0, 0], velocity=[0, 0],
+        self, *, beta=5, center=(0, 0), velocity=(0, 0),
     ):
         """Initialize vortex parameters.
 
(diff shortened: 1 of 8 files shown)


thomasgibson (approved):
LGTM. Thanks for catching these!

mirgecom/initializers.py
```diff
@@ -147,6 +147,8 @@ def __call__(self, x_vec, *, t=0, eos=IdealSingleGas()):
         eos: mirgecom.eos.IdealSingleGas
             Equation of state class to supply method for gas *gamma*.
         """
+        if eos is None:
+            eos = IdealSingleGas()
```
MTCam, line 151: In what now seems like distant past, someone once insisted that this is a bad idea. He was right.  I think we should eliminate defaulting the EOS - which was done for testing convenience.

MTCam (approved):
Great.  I have a feeling this is going to eviscerate a good deal of NS.

inducer:
> eviscerate a good deal of NS.

Yikes! Hope it doesn't. If it does, sorry in advance!

inducer:
And thanks for taking a look, both of you!

PR #339, opened 2021-05-12
Add forward-looking dependency on arraycontext

inducer:
This isn't technically needed yet (`arraycontext` is nothing but an empty shell for now), but it'll help me with some [upstream PR headaches](https://github.com/inducer/meshmode/pull/181).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,6 +46,7 @@ def main():
               "pytest>=2.3",
               "pytools>=2018.5.2",
               "modepy>=2013.3",
+              "arraycontext=2021.1",
               "meshmode>=2013.3",
               "pyopencl>=2013.1",
               "pymbolic>=2013.2",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -46,6 +46,7 @@ def main():
               "pytest>=2.3",
               "pytools>=2018.5.2",
               "modepy>=2013.3",
+              "arraycontext=2021.1",
```
matthiasdiener, line 49: Should this be >= ?

inducer, line 49:
```suggestion
              "arraycontext>=2021.1",
```

inducer: Thanks, good catch! 5d686b6

matthiasdiener (approved):

matthiasdiener:
Doc build will fail until an arraycontext package is released: https://readthedocs.org/projects/mirgecom/builds/13751500/

inducer:
Yikes, sorry. #340.

PR #342, opened 2021-05-17
Fix passing of time variable to boundary

w-hagen:
This fixes bugs in passing the time dependent boundary conditions to the euler_operator, and similar issues when calling the av_operator in the doublemach case


MTCam (approved):
Thanks Wyatt! 👍

PR #344, opened 2021-05-18
Add moving wall boundary treatment

w-hagen:


mirgecom/boundary.py
```diff
@@ -413,6 +413,57 @@ def exterior_grad_q(self, nodes, nhat, grad_q, **kwargs):
                               species_mass=-gradq_comp.species_mass)
 
 
+
+class AdiabaticNoslipMovingBoundary(PrescribedInviscidBoundary):
+    r"""Boundary condition implementing a noslip moving boundary.
+
+    .. automethod:: boundary_pair
+    """
+    def __init__(self, Vw=0.0):
```
MTCam, line 422: I suggest something like: `wall_velocity=None`  then:

```python
if wall_velocity is None:
    wall_velocity = np.zeros(shape=(dim,))
if len(wall_velocity) != dim:
    raise ValueError(f"Specified wall velocity must be a {dim}-vector.")
```

And then the corresponding change in the meaty parts.

w-hagen: Good idea, and I should have picked a better name to start. That said `dim` isn't defined in init, do we want to do this check on every call to exterior_soln?

MTCam: I see what you mean.  How would you feel about making it an argument to the init, defaulting to 2 or something?

w-hagen: My concern is if it is not zero it is going to result in a normal wall component I believe. Zero would just return a typical noslip boundary which was my first instinct, but then it's not really a moving wall.

MTCam: I guess I don't understand why not:

```
w_pen = 2.0*int.mass*np.dot(wall_velocity, nhat) - int.cv_momentum
```

MTCam (approved):
Thanks, Wyatt! 👍

PR #345, opened 2021-05-18
Create model docs for MIRGE-Com

MTCam:
Update the model documentation to:

- Move it away from PC2 (on which it is currently based)
- Correct a gob mistakes in the model terms
- Align with what we currently do (and plan to do)
- Update formulation references to align with our current approach

PR #348, opened 2021-05-20
Reshuffle fluid model/discr doc sections

inducer:
Targets #345.

This shuffles the doc sectioning a bit. @MTCam Look OK? Merge away if so.


MTCam (approved):
Thanks!  👍

PR #349, opened 2021-05-21
Fix docs for arraycontext move

inducer:


MTCam (approved):
Thanks. Pretty dumb; should have realized array context had just moved.

inducer:
Sorry for the trouble. I should  have anticipated that something like this might break.

PR #352, opened 2021-05-21
Fix loss of bc grad function in inits

w-hagen:


MTCam (approved):
Great catch Wyatt!  Thanks! 👍

PR #356, opened 2021-05-23
Model/discr docs: mathbf -> b in LaTeX

inducer:


MTCam (approved):
Ah geez.

inducer:
I'll merge since it only touches docs, and the doc CI has passed.

PR #360, opened 2021-05-23
Modernize Mirgecom

thomasgibson:
This PR does the following:

1. Removes the use of `grudge.discretization.EagerDiscretization`. Why? Because `grudge` is, for the most part, agnostic to whether arrays are eager, gpu, or lazy. It is the job of `arraycontext` to sort out the behavior of arrays, not `grudge`. Moreover, this uses the more stable (non-deprecated) interface of grudge.
2. Updates deprecated imports from `meshmode` to the new `arraycontext` package. This also includes the swapping of arguments for functions like `thaw`.
3. ~Renames `wave-eager-x.py` -> `wave-x.py`. This is because of item 1. When lazy evaluation comes in, the only thing that should really change (from mirgecom's perspective) is the use of a new lazy array context.~ Apparently CI didn't like this.


examples/mixture-mpi.py
```diff
@@ -97,10 +100,10 @@ def main(ctx_factory=cl.create_some_context):
     local_mesh, global_nelements = generate_and_distribute_mesh(comm, generate_mesh)
     local_nelements = local_mesh.nelements
 
-    discr = EagerDGDiscretization(
+    dcoll = DiscretizationCollection(
         actx, local_mesh, order=order, mpi_communicator=comm
     )
-    nodes = thaw(actx, discr.nodes())
+    nodes = thaw(op.nodes(dcoll), actx)
```
MTCam, line 106: Can't we have a `dcoll.nodes()` interface?

mirgecom/diffusion.py
```diff
@@ -35,46 +35,48 @@
...
+def gradient_flux(dcoll, quad_tag, u_tpair):
     r"""Compute the numerical flux for $\nabla u$."""
     actx = u_tpair.int.array_context
 
     dd = u_tpair.dd
     dd_quad = dd.with_discr_tag(quad_tag)
     dd_allfaces_quad = dd_quad.with_dtag("all_faces")
 
-    normal_quad = thaw(actx, discr.normal(dd_quad))
+    normal_quad = thaw(op.normal(dcoll, dd_quad), actx)
```
MTCam, line 56: To be honest, I am not a fan of the new construction in these calls.  Before this change, it vaguely made sense (wasn't great, but you get used to it), and after these changes it makes almost no intuitive sense; that is, it is much harder for someone to intuit (or even remember) how to get XYZ from the discretization.  

This is something that our physics developers will need to do a *lot*. Is there a way we can massage this interface so that it makes more sense for those types of users?

thomasgibson: So you would prefer something along the lines of `dcoll.normal(dd)`, `dcoll.nodes(dd)`, etc? I think that is a reasonable request. Thing to remember that `dcoll` is _not_ a single discretization. It's a collection of discretizations (volume, face, quadrature, etc).

thomasgibson: >To be honest, I am not a fan of the new construction in these calls. Before this change, it vaguely made sense (wasn't great, but you get used to it), and after these changes it makes almost no intuitive sense; that is, it is much harder for someone to intuit (or even remember) how to get XYZ from the discretization.

@MTCam So here is an explanation for what `op.nodes(dcoll, dd)` _means_. First, what is `dcoll`? `dcoll` is, as the name suggests, a collection of meshmode discretization objects (interior faces, boundaries, volumes, and perhaps quadrature-related ones too). When you ask for `nodes`, you need to qualify this statement with "which nodes?" (similarly for normals, but these are restricted to facet-based discretizations).

So `op.nodes(dcoll, dd)` can be read as: "Here is a collection of discretizations defining my problem (`dcoll`). Reach in and grab the nodes associated with the `dd`-discretization ("vol", "int faces", etc)." Hope that perhaps explains why grudge does it this way, using _functions_ rather than class methods. Functions are generally more composable, and it fits in line with how DG operators are constructed (`op.weak_local_grad(dcoll, vec)`, `op.inverse_mass(dcoll, vec)` etc).

_Having said all that_, I don't necessarily see a huge issue with adding `normal` and `nodes` as methods to `dcoll`. However, that's about a far as I am really comfortable doing. `dcoll` is already a big class. We don't _need_ to make it even larger by adding all these methods to it. And we don't really gain anything from it.

examples/mixture-mpi.py
majosm: `dcoll.discr_from_dd("vol").nodes()`?

mirgecom/diffusion.py
MTCam: > @MTCam So here is an explanation for what `op.nodes(dcoll, dd)` means. First, what is `dcoll`? `dcoll` is, as the name suggests, a collection of meshmode discretization objects (interior faces, boundaries, volumes, and perhaps quadrature-related ones too). When you ask for `nodes`, you need to qualify this statement with "which nodes?" (similarly for normals, but these are restricted to facet-based discretizations).

`dcoll.nodes(dd)` to me does not have less information than `op.nodes(dcoll, dd)`.  To my eyes, the only difference is the latter requires a need-to-know that `nodes` should come from `op` and not from the discretization collection itself.

> 
> So `op.nodes(dcoll, dd)` can be read as: "Here is a collection of discretizations defining my problem (`dcoll`). Reach in and grab the nodes associated with the `dd`-discretization ("vol", "int faces", etc)." Hope that perhaps explains why grudge does it this way, using _functions_ rather than class methods. Functions are generally more composable, and it fits in line with how DG operators are constructed (`op.weak_local_grad(dcoll, vec)`, `op.inverse_mass(dcoll, vec)` etc).

But as I mentioned above, can't `dcoll.nodes(dd)` already indicate to reach in and grab nodes associated with `dd`?

Follow-on question:  Will not all of the discretizations in a collection have stuff like "int_faces"?!

>
> Having said all that, I don't necessarily see a huge issue with adding `normal` and `nodes` as methods to `dcoll`. However, that's about a far as I am really comfortable doing. `dcoll` is already a big class. We don't need to make it even larger by adding all these methods to it. And we don't really gain anything from it.

I see where you are coming from about this, and share your desire to keep the underlying infrastructure clean and lean,  but consider that one of the main and oft touted features of the infrastructure we are trying to build here is that the physics developers can easily intuit how to roll their models into the code. Changes that significantly affect the physics user's ability to read and (~immediately) understand, or (~easily) intuit how to implement a thing frustrate those user's efforts.  I would argue that not losing that quality of our code is important to the extent that it could be marked in the "something to gain" from it category.

So... along these lines: `thaw` is one of the least intuitive thing we have going. Our physics users currently just cut-and-paste this where they need, but they don't understand it, or care. It just gets in the way of their quickly implementing a new thing that needs an array from the discretization.   After the changes proposed in this PR, there is yet another layer: `op`, and at least from my perspective - one needs to guess or comb for examples about when it should be used.  For example `dcoll.mesh` works fine, `dcoll.nodes` does not, and a handful of other unintuitive things.

thomasgibson: >Follow-on question: Will not all of the discretizations in a collection have stuff like "int_faces"?!

No. When you do something like `discr.discr_from_dd(dd)` (as done throughout `main`), you're getting a meshmode discretization object _specifically_ for the entities described by `dd`. Meshmode, for better or for worse, calls a "discretization" a set of nodes defined on mesh entities defined by `dd`. `DiscretizationCollection` as defined in grudge is a _collection_ of meshmode discretizations and provides connections (interpolation/projection) between them.

>So... along these lines: thaw is one of the least intuitive thing we have going.

No arguments there; I completely agree. IMO methods like `thaw` should be more or less hidden from the user (at the driver level at least). In 99% of the cases `nodes`/`normals` are requested, it's wanted in a thawed state. So I think we should do something about that

> After the changes proposed in this PR, there is yet another layer: op, and at least from my perspective - one needs to guess or comb for examples about when it should be used. For example dcoll.mesh works fine, dcoll.nodes does not, and a handful of other unintuitive things.

You view it as "another layer" but this has been in grudge since April: https://github.com/inducer/grudge/pull/67. This is _not_ new. @inducer made this public and even went over it in one of the dev meetings. This PR is merely exposing it.

Now I will say this, I think `grudge.op` is really only necessary for constructing operators: `euler`, `diffusion`, etc. Much of this can be hidden from the user by having mirgecom expose functions which import `grudge.op`. The reason it's imported at the driver level is because of functions like `op.norm`, `op.nodes`, `op.normal` in some cases. So if we address how to handle these without importing `grudge.op`, then a user won't need to see it (if that's the main concern here.)

thomasgibson:
So my plan is to wait until https://github.com/inducer/grudge/pull/74 is merged before this PR is "ready." However, I think we should keep commenting about the user interface until we're all happy with it.

examples/mixture-mpi.py
thomasgibson: > `dcoll.discr_from_dd("vol").nodes()`?

Yeah that's basically what `op.nodes(dcoll, dd=None)` does

thomasgibson: We should also make `nodes` always return a `thaw`ed object

inducer: But then you have to pass in an array context.

mirgecom/diffusion.py
inducer: First of all, I appreciate the discussion about interface ergonomics here!

> IMO methods like thaw should be more or less hidden from the user (at the driver level at least).

I can see where you're coming from about `thaw`. Unfortunately, I can't think of a way to make it entirely transparent to the user, especially in view of lazy eval.

I can also see the argument for `dcoll.nodes()` instead of `op.nodes(dcoll)`. Similarly for `normal`. `norm` should stay `op.norm` IMO.

inducer:
Thanks @thomasgibson for taking this on! (Just did a quick scroll here, no huge red flags spotted.)

inducer:

examples/mixture-mpi.py
thomasgibson: Maybe let's just not freeze instead? https://github.com/inducer/grudge/blob/thg/remove-symbolics/grudge/op.py#L202

thomasgibson: Actually, this would be problematic for lazy, right?

mirgecom/diffusion.py
MTCam: > > Follow-on question: Will not all of the discretizations in a collection have stuff like "int_faces"?!
> 
> No. When you do something like `discr.discr_from_dd(dd)` (as done throughout `main`), you're getting a meshmode discretization object _specifically_ for the entities described by `dd`. Meshmode, for better or for worse, calls a "discretization" a set of nodes defined on mesh entities defined by `dd`. `DiscretizationCollection` as defined in grudge is a _collection_ of meshmode discretizations and provides connections (interpolation/projection) between them.
>

If I have a collection of discretizations in `dcoll` that includes at least the fluid volume discretization and the structures volume discretization and I ask for `dcoll.discr_from_dd("int_faces")`; that seems like trouble to me. Won't they all have internal faces? I have a similar question about a collection that includes a regular and quadrature discretization.  Doesn't the quadrature disc have "int_faces" too?  More succinctly; don't we face a potential issue for dd/name collisions among discretizations under the current construction?

> > After the changes proposed in this PR, there is yet another layer: op, (...)
> 
> You view it as "another layer" but this has been in grudge since April: [inducer/grudge#67](https://github.com/inducer/grudge/pull/67). This is not new. @inducer made this public and even went over it in one of the dev meetings. This PR is merely exposing it. But I think we can perhaps hide it if that's the desire.
> 

I don't mean to say it is unnecessary, or that it isn't a better way to structure the infrastructure.  My only complaint at this point is about the user-facing interface to it, which I can only now (thanks to this PR) get a firm grasp of how its use impacts the code that I (and others) will need to write to implement models and methods, etc.

> `grudge.op` is really only necessary for constructing operators: `euler`, `diffusion`, etc. Much of this can be hidden from the user by having mirgecom expose functions which import `grudge.op`. The reason it's imported at the driver level is because of functions like `op.norm`, `op.nodes`, `op.normal` in some cases. So if we address how to handle these without importing `grudge.op`, then a user won't need to see it (if that's the main concern here.)

Our users should be thought of as user-developers, primarily from the physical science side.  @anderson2981, and @w-hagen are typical example users. They add things to the code deeper than driver level.  They add rhs operators, source terms, boundary treatments, and initializations. Even a "light" user will typically need to add or tweak existing initializations, and boundaries.  They also (and this is very important) dig into the code at a deeper level to understand how it is impacting their running or planned simulation.

I think we should lean on making the interface those users see such that when they write (and read) their expressions - they are quite close to the written out formulation - with as much of the infrastructure as possible staying out of the way.  That said, I also think it is important to expose enough of the infrastructure that help users remain observant (at least tangentially) of data motion, dependencies, and other "features" of the code in the setting of lazy, etc. It's a two-way street where we train the users, but they also train us a bit; both on what tools they need to do to accomplish their science, and especially on the interface to the tools we provide them for the same purpose. 

I'm not saying these things because I think you (or others among us) are opposed to this view. It is only my intent to provide some perspective on why I'm saying them.  While is it not terribly hard to approach/approximate the kind of physics/infrastructure balance that I'm talking about here, it is very easy for it to slip away if we don't keep our eyes on it.

majosm: > If I have a collection of discretizations in `dcoll` that includes at least the fluid volume discretization and the structures volume discretization and I ask for `dcoll.discr_from_dd("int_faces")`; that seems like trouble to me. Won't they all have internal faces? I have a similar question about a collection that includes a regular and quadrature discretization. Doesn't the quadrature disc have "int_faces" too? More succinctly; don't we face a potential issue for dd/name collisions among discretizations under the current construction?

The current API is not multi-volume compatible at the moment. It's likely that `"vol"`, `"int_faces"`, etc. will be going away eventually.

As for quadrature: I think `"int_faces"` is a shorthand for `DOFDesc(FACE_RESTR_INTERIOR, DISCR_TAG_BASE)`, i.e., no quadrature because of `DISCR_TAG_BASE` (formerly `QTAG_NONE`). Do I have this right @thomasgibson ?

inducer: > The current API is not multi-volume compatible at the moment. It's likely that `"vol"`, `"int_faces"`, etc. will be going away eventually.

Personally, I'd prefer if we kept those working in the (common) case of single volumes. I agree that they will not work in the multi-volume case.

thomasgibson: > The current API is not multi-volume compatible at the moment. It's likely that `"vol"`, `"int_faces"`, etc. will be going away eventually.

These definitely will not work in the long term; they're simply not descriptive enough for coupled systems. I think we can probably utilize some of the tech you started in https://github.com/inducer/meshmode/pull/129 to help address this!

> As for quadrature: I think `"int_faces"` is a shorthand for `DOFDesc(FACE_RESTR_INTERIOR, DISCR_TAG_BASE)`, i.e., no quadrature because of `DISCR_TAG_BASE` (formerly `QTAG_NONE`). Do I have this right @thomasgibson ?

Yep, exactly

examples/mixture-mpi.py
thomasgibson: @MTCam I've gone ahead and updated everything to use `dcoll.nodes(...)` and `dcoll.normal(...)`

thomasgibson:
Haven't forgotten about this PR; I would prefer to have https://github.com/inducer/grudge/pull/122 in first

MTCam:
> Haven't forgotten about this PR; I would prefer to have [inducer/grudge#122](https://github.com/inducer/grudge/pull/122) in first

Hey @thomasgibson , I think this addresses some portion of #509 ?

thomasgibson:
> > Haven't forgotten about this PR; I would prefer to have [inducer/grudge#122](https://github.com/inducer/grudge/pull/122) in first
> 
> Hey @thomasgibson , I think this addresses some portion of #509 ?

Yes it indeed does! It will take care of removing references to `discr` (and allow us to remove `EagerDGDiscretization` all together)

PR #364, opened 2021-05-27
Add p refinement functionality

w-hagen:
An initial attempt to upscale a object array of DOFArrays from one discr to another (higher) order one.
@MTCam @anderson2981


inducer:
Just make both discretizations and use https://documen.tician.de/meshmode/connection.html#same-mesh-connections.

w-hagen:
Thanks! Don't know how I missed that.

MTCam:
Yikes, #366 needed this to live. :(

PR #366, opened 2021-05-28
Restart with different order

MTCam:
Was going to use #364, then the direct connection suggested by @inducer.  This was as far as i got just now. 

cc: @w-hagen, @anderson2981


examples/wave-eager-mpi.py
```diff
@@ -139,7 +139,15 @@ def main(snapshot_pattern="wave-eager-{step:04d}-{rank:04d}.pkl", restart_step=N
...
         istep = restart_step
         assert istep == restart_step
-        fields = restart_data["fields"]
+        restart_fields = restart_data["fields"]
+        old_order = restart_data["order"]
+        if old_order != order:
+            old_discr = EagerDGDiscretization(actx, local_mesh, order=old_order,
+                                              mpi_communicator=comm)
+            from mirgecom.simutil import polynomial_refine
+            fields = polynomial_refine(discr, old_discr, restart_fields)
```
w-hagen, line 148:
```suggestion
            from meshmode.discretization.connection import make_same_mesh_connection
            connection = make_same_mesh_connection(actx, discr.discr_from_dd("vol"),
                                                   old_discr.discr_from_dd("vol"))
            fields = connection(restart_fields)
```
I tested the suggestion and I think this will do the trick

MTCam: Thank you, Wyatt! 👍

MTCam:
@anderson2981 is already using this successfully in production.  Note that it merges to #188, not main.

examples/wave-eager-mpi.py
```diff
@@ -139,7 +139,17 @@ def main(snapshot_pattern="wave-eager-{step:04d}-{rank:04d}.pkl", restart_step=N
         t = restart_data["t"]
         istep = restart_step
         assert istep == restart_step
-        fields = restart_data["fields"]
+        restart_fields = restart_data["fields"]
+        old_order = restart_data["order"]
+        if old_order != order:
+            old_discr = EagerDGDiscretization(actx, local_mesh, order=old_order,
+                                              mpi_communicator=comm)
```
inducer, line 146: Not loving that this still uses the old-timey `EagerDGDiscretization`, but OK with me.

inducer (approved):
Thanks, LGTM!

PR #367, opened 2021-05-28
cantera: make sure not to pull in MPI

matthiasdiener:
Fixes the logpyle issues seen on Lassen etc. 🤷 

cantera depends on h5py, which depends on hdf5 (and therefore MPI). Force the non-MPI version of h5py to be installed.


MTCam (approved):
Seems like a great idea if it works :) 👍

PR #368, opened 2021-05-28
Enable CFL and DT 

MTCam:
With this changeset we pick up grudge's new `characteristic_lengthscales` interface to get CFL and maximum stable DT estimates.  This enables the existing infrastructure in *MIRGE-Com* for:

- Constant DT mode:
  - Fixed timestep, variable CFL	
  - CFL tracking and reporting
  - Cell-local CFL visualization
- Constant CFL mode:
  - Variable timestep, fixed CFL
  - DT tracking and reporting
  - Cell-local maximum stable DT visualization


mirgecom/inviscid.py
```diff
@@ -82,21 +82,25 @@ def get_inviscid_timestep(discr, eos, cfl, q):
...
+    max_wavespeed_local = op.nodal_max(discr, "vol", local_wavespeeds)
+
+    mpi_comm = discr.mpi_communicator
+    if mpi_comm is None:
+        return cfl * h_min_local / max_wavespeed_local
+
+    h_min_global = mpi_comm.allreduce(h_min_local, op=MPI.MIN)
+    max_wavespeed_global = mpi_comm.allreduce(max_wavespeed_local, op=MPI.MAX)
+
+    return cfl * h_min_global / max_wavespeed_global
```
inducer, line 103: I think this is kind of dangerous. It's perhaps better than nothing, but it'll fail for elements with large aspect ratio. @MTCam mentioned a more robust procedure given in the JSH/TW book, which I've had good experience with. Question: Should we merge this as a stopgap until we have that, or go for the actual article.

thomasgibson: @inducer Where is the "more robust procedure"? @MTCam and I spend some time trying to find it. This approach here is basically the formula given on page 325, equation 7.32

MTCam: I def remember reading it, but @thomasgibson and I searched together and could not dig it up.

inducer: The 1D approach is towards the end of Section 4.7 for me:

grafik

The generalization to more dimensions is in Section 6.4:

grafik

Also
- https://github.com/tcew/nodal-dg/blob/master/Codes1.1/Codes2D/dtscale2D.m
- https://github.com/tcew/nodal-dg/blob/master/Codes1.1/Codes3D/dtscale3D.m

MTCam: Personally, I like the expression Hesthaven uses here:

Screen Shot 2021-05-28 at 3 59 39 PM

It has the advantage that it comes out in units of time, and is easy to understand.  It doesn't need to be very precise - it's just a guide and the user will need to tune dt or cfl to the particular simulation anyway.

inducer: They're not incompatible though. The approach above just provides more precise C's to the formula you mention.

MTCam: I will implement whatever you guys think is best.  

My main complaints about the other formula is that it is hard to understand, and is not clear to me how to derive it for 3D (or even 2D how it's presented), and it does not have the correct units.  Since it needs to be tuned at runtime no matter what, in my opinion understandability and proper units wins.

If grudge can provide the min del R and min R_d, we can whip it out.

MTCam: But also, ugh, the 7.32 formula at least has some consideration of the viscous terms, and seems specific to viscous flow.  I like that.

thomasgibson: > I think this is kind of dangerous. It's perhaps better than nothing, but it'll fail for elements with large aspect ratio.

@inducer I can see the approach you linked perhaps gives a tighter bound, but how is this any less dangerous? The definition of `r_D` uses the cell area / perimeter. This is just as dangerous for degenerate elements.

> it'll fail for elements with large aspect ratio

Can you clarify here what you mean? In the atmosphere, we deal with "large aspect ratios" all the time. What is "large" here? ~O(100)? ~O(1000)?

thomasgibson: Maybe we should be using the circumradius in our calculations? https://mathworld.wolfram.com/Circumsphere.html

inducer: All I'm saying is that I'd like something that vaguely responds to the element aspect ratio, with a test that it "catches" the growth of the eigenvalues of DG differentiation under modest amounts of h and p refinement. Aspect ratio is easy to test by "squishing' (affine-mapping) a rect mesh.

thomasgibson: This comment is rather concerning (for the 3D case): https://github.com/tcew/nodal-dg/blob/master/Codes1.1/Codes3D/Maxwell3D.m#L14

thomasgibson: Found this paper: https://arxiv.org/pdf/1806.03095.pdf, where they choose a timestep that is more or less what @MTCam posted (equation 7.32 in H&W). 
image

Not necessarily claiming it's better, just that it's at least taking into account the viscous part. One thing I'm concerned about with what @inducer posted is how the viscous timestep should be handled. I've started coding it up, but I'm still not totally convinced it's better.

thomasgibson: Should `2/3 min delta r` replace the `1/N**2` term in the bound and `h_min` is the circumscribed radius?

thomasgibson: Uhg DG + time-stepping is so damn ugly

inducer: > Should `2/3 min delta r` replace the `1/N**2` term in the bound and `h` is the circumscribed radius?

I'd say so. (Are you pursuing the thing with the singular values of the mapping Jacobian we had discussed any further, as opposed to the incircle or circumcircle radius?)

inducer: > Uhg DG + time-stepping is so damn ugly

:shrug: You're guessing eigenvalues. At a cost, you could also use adaptive time stepping and forget about CFL estimation.

thomasgibson: > (Are you pursuing the thing with the singular values of the mapping Jacobian we had discussed any further, as opposed to the incircle or circumcircle radius?)

I going to try that, but I'm struggling to get the right matrix per element. I think I should be using https://github.com/inducer/grudge/blob/main/grudge/geometry/metrics.py#L180, but I need to reshape the components and get it into the form of a matrix. My plan from there was to do a SVD calculation and take the minimum singular value.

inducer: https://github.com/inducer/grudge/pull/107

inducer: (I.e.: prototyped something)

thomasgibson: Took a stab at extending @inducer's prototype and cleaned up the functions: https://github.com/inducer/grudge/pull/108

thomasgibson: OK, I'm happy now with https://github.com/inducer/grudge/pull/108: @MTCam, this should make the `dt_non_geometric_factor`/`dt_geometric_factor` usable for this PR. Don't worry about the viscous case; these two factors are designed to give you an approximation to `h_min / N^2`, which means the only difference in the expression you posted that handles the viscous timestep is that you replace `h / N^2` with the product of the nongeometric and geometric factors.

MTCam: Wow, that's cool!  Thanks a ton @thomasgibson and @inducer!

mirgecom/inviscid.py
```diff
@@ -77,29 +77,20 @@ def inviscid_flux(discr, eos, q):
...
-#    max_v = clmath.max(wavespeeds)
-#    return c*dt_ngf*dt_gf/max_v
+    return (
+        cfl * dt_non_geometric_factor(discr) * dt_geometric_factors(discr)
+        / compute_wavespeed(dim, eos, cv)
+    )
 
 
 def get_inviscid_cfl(discr, eos, dt, q):
     """Calculate and return CFL based on current state and timestep."""
```
thomasgibson, line 95: Maybe clarify in the doc that this returns a node-wise CFL approximation?

mirgecom/simutil.py
```diff
@@ -66,12 +66,17 @@ def inviscid_sim_timestep(discr, state, t, dt, cfl, eos,
...
+    dt_left = t_final - t
+    if dt_left < 0:
+        return 0.0
     if constant_cfl is True:
-        mydt = get_inviscid_timestep(discr=discr, q=state,
-                                     cfl=cfl, eos=eos)
-    if (t + mydt) > t_final:
-        mydt = t_final - t
-    return mydt
+        from grudge.op import nodal_min
```
thomasgibson, line 73: Just a comment that `nodal_min` (and other nodal reductions like max, norm) will become distributive reductions (https://github.com/inducer/grudge/pull/117)

MTCam: This function won't work right until that lands.

mirgecom/inviscid.py
MTCam: see how this grabs ya. (80d8087)

mirgecom/simutil.py
thomasgibson: It landed haha, so `nodal_min` is now MPI-aware

mirgecom/inviscid.py
```diff
@@ -76,30 +76,63 @@ def inviscid_flux(discr, eos, q):
...
 
-#    dt_ngf = dt_non_geometric_factor(discr.mesh)
-#    dt_gf  = dt_geometric_factor(discr.mesh)
-#    wavespeeds = compute_wavespeed(w,eos=eos)
-#    max_v = clmath.max(wavespeeds)
-#    return c*dt_ngf*dt_gf/max_v
+    return (
+        dt_non_geometric_factor(discr) * dt_geometric_factors(discr)
+        / compute_wavespeed(dim, eos, cv)
+    )
```
thomasgibson, line 111: Do you think you could use this utility function `estimate_local_timestep`: https://github.com/inducer/grudge/pull/121/files#diff-59daa571898639daffffa78566f0fe995f723ebe88372e4d1dad5ce27268c0bd?

MTCam: Yes, I think so. (b58594c)

examples/autoignition-mpi.py
```diff
@@ -224,9 +225,13 @@ def my_rhs(t, state):
 
     def my_checkpoint(step, t, dt, state):
         reaction_rates = eos.get_production_rates(state)
-        viz_fields = [("reaction_rates", reaction_rates)]
+        local_cfl = get_inviscid_cfl(discr, eos=eos, dt=current_dt, cv=state)
```
inducer, line 228: Unfortunately, this is potentially tangled up in the driver structure mess that we're currently in. As it stands, it seems that this recomputes the CFL number separate from the value that's actually used for determining the time step, giving rise to potential inconsistency. Is that correct?

mirgecom/simutil.py
```diff
@@ -66,12 +66,16 @@ def inviscid_sim_timestep(discr, state, t, dt, cfl, eos,
                           t_final, constant_cfl=False):
     """Return the maximum stable dt."""
     mydt = dt
+    dt_left = t_final - t
+    if dt_left < 0:
+        return 0.0
```
inducer, line 71:
```suggestion
    if t_remaining < dt:
        return max(0, t_remaining)
```

mirgecom/simutil.py
```diff
@@ -66,12 +66,16 @@ def inviscid_sim_timestep(discr, state, t, dt, cfl, eos,
                           t_final, constant_cfl=False):
     """Return the maximum stable dt."""
     mydt = dt
+    dt_left = t_final - t
```
inducer, line 69:
```suggestion
    t_remaining = t_final - t
```
? I was briefly confused what this had to do with left/right.

inducer (commented):
Thanks! Some thoughts from a quick scroll.

inducer:
> Constant CFL mode (a.k.a. adaptive timestepping)

FWIW, to my mind "adaptive timestepping" is a different concept, e.g. driven by a timestepper that has an error estimator built in, such as an embedded pair.

examples/autoignition-mpi.py
MTCam: *MIRGE-Com* has built-in two modes for timestepping:

Constant CFL:  The user specifies the CFL, and DT is calculated at every step to maintain a constant CFL.  We calculate the domain-global DT as `constant_cfl * stablity_limited_timestep`.

Constant DT: This is the mode we've been operating in since we've been waiting on the grudge back-end for the equivalent of `characteristic_lengthscale`.  In this mode, CFL is not constant; it is a dependent variable that changes with wavespeed (among other things in reactive, viscous flow, but currently just wavespeed). 

Calculating, and visualizing the cell-local CFL in constant DT mode as is done in this example, provides the user a way to examine potential "trouble spots" in the simulation.  The cells with high CFL indicate the DT-limiting areas that require particular care in making a stable sim.

MTCam:
> > Constant CFL mode (a.k.a. adaptive timestepping)
> 
> FWIW, to my mind "adaptive timestepping" is a different concept, e.g. driven by a timestepper that has an error estimator built in, such as an embedded pair.

Ah, makes sense. I should have just said constant cfl mode then.

examples/autoignition-mpi.py
MTCam: > Unfortunately, this is potentially tangled up in the driver structure mess that we're currently in. As it stands, it seems that this recomputes the CFL number separate from the value that's actually used for determining the time step, giving rise to potential inconsistency. Is that correct?

What I mean is that in constant dt mode, which this example is, then CFL needs computing only at viz time based on the user's specified constant dt. We only do it here because the user wants to visualize it.  We don't need or use the actual CFL in the simulation, it's only an example of one of the diagnostic things the user wants to do.  The CFL computed is consistent with the cell-local dts. 

If instead we use constant CFL mode, for which there is no current example, we need to compute dt at every step using the user-specified CFL. In that case the user may like to visualize the cell-local dt's to get an idea of dt-limiting areas. The dt that is computed is consistent with the user's specified CFL, and with that actually used in the simulation.

The driver-adjacent support is here:
https://github.com/illinois-ceesd/mirgecom/blob/6ff984d66909a14d628709f83d6decd85a49846b/mirgecom/simutil.py#L65

I believe this should blend right in with the driver structure development, and is already being used in the y1 drivers.

examples/mixture-fixed-cfl-mpi.py
```diff
@@ -0,0 +1,313 @@
+"""Demonstrate combustive mixture with constant CFL mode."""
```
thomasgibson, line 1: In light of the new example added to `examples`, I was thinking: Do we want to try to cut down the examples and only keep ones that we need for particular functionality? For example, probably don't need `wave-eagar` _and_ `wave-eagar-mpi`? Maybe drop the other `mixture-mpi.py` in favor of this new one? I worry about the steadily increasing CI times, but perhaps It's not that big of a deal?

MTCam: I definitely would like to see the wave stuff go.  It calls very little of the mirgecom infrastructure and I think they are misleading as mirgecom examples. 

Definitely would plan on merging this mixture example to the existing, or just enabling constant CFL mode optionally on all the examples too.

majosm: I would personally rather see wave-eager or wave-eager-mpi get updated with the modern simutils, etc. stuff when that’s done instead of trashing them. I don’t know that we need to keep both of them though.

MTCam: I'd feel a lot better about the wave examples if they used an operator, and boundary data structures from mirgecom. For me, simutils is a little too driver-adjacent or surface to make the wave examples mirgecom-relevant. They just seem like grudge examples copied to the mirgecom examples dir.

MTCam: Let's add a readme for the examples? I hope this can help us organize our thoughts around what can get the axe.  #392

examples/autoignition-mpi.py
inducer: OIC, nvm then.

examples/mixture-fixed-cfl-mpi.py
```diff
@@ -0,0 +1,313 @@
...
+        do_viz = check_step(step, nviz)
+        viz_fields = [("cv", state)]
+        current_dt = dt
+
+        if constant_cfl:  # no matter what
+            local_dt = get_inviscid_timestep(discr, eos=eos, cv=state)
+            from grudge.op import nodal_min
+            current_dt = current_cfl * nodal_min(discr, "vol", local_dt)
+        elif do_viz:   # only if visualizing
+            local_cfl = get_inviscid_cfl(discr, eos=eos, dt=dt, cv=state)
```
inducer, line 233:
```suggestion
        else:
            # constant dt
            if do_viz:   # only if visualizing
                local_cfl = get_inviscid_cfl(discr, eos=eos, dt=dt, cv=state)
```

mirgecom/steppers.py
```diff
@@ -80,12 +81,13 @@ def _advance_state_stepper_func(rhs, timestepper, checkpoint, get_timestep,
...
-        if dt < 0:
-            return istep, t, state
+        if get_timestep:
+            dt = get_timestep(state=state)
+            if dt < 0:
+                return istep, t, state
 
         if checkpoint:
-            checkpoint(state=state, step=istep, t=t, dt=dt)
+            state, dt = checkpoint(state=state, step=istep, t=t, dt=dt)
```
inducer, line 90:
```suggestion
            dt = checkpoint(state=state, step=istep, t=t, dt=dt)
```

examples/mixture-fixed-cfl-mpi.py
```diff
@@ -0,0 +1,313 @@
...
+                            f"---- T({min_temperature}, {max_temperature})\n")
+
+        # this exit is safe, errors(current_dt) is already collective
+        if errors:
+            logger.info("Fatal error: Invalid simulation DT")
+            import sys
+            sys.exit()
+
+        t_remaining = max(0, t_final - t)
+        return state, min(t_remaining, current_dt)
```
inducer, line 278:
```suggestion
        return min(t_remaining, current_dt)
```

MTCam:
@cmikida2 hoping to get your eyes-on to make sure I didn't tear up Leap integration with changes supporting constant cfl.

cmikida2 (approved):
Looks good to me, Mike!

mirgecom/steppers.py
MTCam: Fixed this together with the other places it needed fixin.

examples/mixture-fixed-cfl-mpi.py
MTCam: Ditto; should be all fixed up.

MTCam:
> Looks good to me, Mike!

Thanks!  👍

examples/autoignition-mpi.py
```diff
@@ -358,6 +366,23 @@ def my_health_check(dv):
 
         return health_error
 
+    def my_get_timestep(t, dt, state):
+        #  richer interface to calculate {dt,dfl} returns node-local estimates
```
thomasgibson, line 370:
```suggestion
        #  richer interface to calculate {dt,cfl} returns node-local estimates
```

thomasgibson: Out of curiosity: Do you think it would be worth adding short docstrings to these `my_foo(...)` functions used in the callback routines? Nothing super formal, just a short comment explaining what it does.

thomasgibson (commented):
Just some minor comments. Thanks for seeing this to the end!

examples/autoignition-mpi.py
MTCam: I think that code comments would work fine for those;  I am not sure about adding docstrings for the driver things that aren't in the library. That said, I am not opposed at all if you think they should be there.  Since they are completely user option - I feel a little funny about legitimizing them by adding docs.

What I do really think we need, is documentation that explains the callbacks, including the cautionary statements about state modification - and possibly including examples of callbacks from the example drivers.

MTCam: FYI, I applied your suggestion but unresolved cause we were talking here.  How rude, github.

thomasgibson: > What I do really think we need, is documentation that explains the callbacks, including the cautionary statements about state modification - and possibly including examples of callbacks from the example drivers.

Fair enough. That can be done in a separate PR. I'm totally fine with that as an alternative.

MTCam: #427

.ci-support/production-setup.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 set -x
-PRODUCTION_CHANGE_OWNER="illinois-ceesd"
+PRODUCTION_CHANGE_OWNER=""
```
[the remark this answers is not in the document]
MTCam: Because this branch merges clean with y1-production, we don't need to change y1-production to make it compatible. It really doesn't matter what this is set to, because it doesn't even look at this unless the PRODUCTION_CHANGE_BRANCH is set.

thomasgibson (approved):

PR #369, opened 2021-05-28
Fix wave examples

thomasgibson:
This PR addresses the problems we discussed today at the dev meeting.

The problem related to getting `-inf` in the temperature calculations is fixed in this grudge PR: https://github.com/inducer/grudge/pull/105

A forthcoming PR will add a callback routine to check for NaNs/blowup.

Other than that, the simulations appear to be okay.


inducer (approved):
Thanks!

PR #370, opened 2021-05-28
Add callback hooks to the simulation stepper

thomasgibson:
Adds pre/post-step callbacks to `advance_state`. This also removes the `checkpoint` argument since it is technically a callback. All examples have been updated.

I also threw in a health-check callback that is exercised in the examples. Will write a short unit test for it as well.

cc @MTCam @anderson2981

### DIFF
--- a/examples/autoignition-mpi.py
+++ b/examples/autoignition-mpi.py
@@ -39,13 +39,10 @@ from grudge.shortcuts import make_visualizer
 from mirgecom.euler import euler_operator
 from mirgecom.simutil import (
     inviscid_sim_timestep,
-    sim_visualization,
     sim_checkpoint,
-    sim_healthcheck,
     check_step,
     generate_and_distribute_mesh
 )
-from mirgecom.exceptions import SynchronizedError
 from mirgecom.fluid import split_conserved
 from mirgecom.io import make_init_message
 from mirgecom.mpi import mpi_entry_point
@@ -85,7 +82,6 @@ def main(ctx_factory=cl.create_some_context, use_leap=False):
     constant_cfl = False
     nstatus = 1
     nviz = 5
-    ncheck = 1
     rank = 0
     checkpoint_t = current_t
     current_step = 0
@@ -236,30 +232,12 @@ def main(ctx_factory=cl.create_some_context, use_leap=False):
         cv = split_conserved(dim, state)
         reaction_rates = eos.get_production_rates(cv)
         viz_fields = [("reaction_rates", reaction_rates)]
-        try:
-            # Check the health of the simulation
-            sim_healthcheck(discr, eos, state,
-                            step=step, t=t, freq=ncheck)
-            # Perform checkpointing
-            sim_checkpoint(discr, eos, state,
-                           step=step, t=t, dt=dt, freq=nstatus,
-                           constant_cfl=constant_cfl)
-            # Visualize
-            sim_visualization(discr, eos, state,
-                              visualizer, vizname=casename,
-                              step=step, t=t, freq=nviz,
-                              viz_fields=viz_fields)
-        except SynchronizedError as err:
-            # Log crash error message
-            if rank == 0:
-                logger.info(str(err))
-                logger.info("Visualizing crashed state ...")
-            # Write out crashed field
-            sim_visualization(discr, eos, state,
-                              visualizer, vizname=casename,
-                              step=step, t=t, freq=1,
-                              viz_fields=viz_fields)
-            raise err
+        # Perform status checkpointing
+        sim_checkpoint(discr, visualizer, eos, q=state,
+                       vizname=casename, step=step,
+                       t=t, dt=dt, nstatus=nstatus, nviz=nviz,
+                       constant_cfl=constant_cfl,
+                       viz_fields=viz_fields)
         return state
 
     current_step, current_t, current_state = \
--- a/examples/autoignition-mpi.py
+++ b/examples/autoignition-mpi.py
@@ -41,7 +41,7 @@ from mirgecom.simutil import (
     inviscid_sim_timestep,
     sim_visualization,
     sim_checkpoint,
-    sim_cfd_healthcheck,
+    cfd_healthcheck,
     check_step,
     generate_and_distribute_mesh
 )
@@ -233,29 +233,32 @@ def main(ctx_factory=cl.create_some_context, use_leap=False):
                 + eos.get_species_source_terms(cv))
 
     def my_checkpoint(step, t, dt, state):
+        cv = split_conserved(dim, state)
+        reaction_rates = eos.get_production_rates(cv)
+        viz_fields = [("reaction_rates", reaction_rates)]
         try:
             # Check the health of the simulation
-            sim_cfd_healthcheck(discr, eos, q=state,
-                                ncheck=ncheck, step=step, t=t)
+            cfd_healthcheck(discr, eos, state,
+                            step=step, t=t, freq=ncheck)
             # Perform checkpointing
-            sim_checkpoint(discr, eos, q=state,
-                           exact_soln=initializer,
-                           step=step, t=t, dt=dt,
-                           nstatus=nstatus,
+            sim_checkpoint(discr, eos, state,
+                           step=step, t=t, dt=dt, freq=nstatus,
                            constant_cfl=constant_cfl)
             # Visualize
-            sim_visualization(discr, state, eos,
+            sim_visualization(discr, eos, state,
                               visualizer, vizname=casename,
-                              step=step, t=t, nviz=nviz)
+                              step=step, t=t, freq=nviz,
+                              viz_fields=viz_fields)
         except StepperCrashError as err:
             # Log crash error message
             if rank == 0:
                 logger.info(str(err))
                 logger.info("Visualizing crashed state ...")
             # Write out crashed field
-            sim_visualization(discr, state, eos,
+            sim_visualization(discr, eos, state,
                               visualizer, vizname=casename,
-                              step=step, t=t, nviz=1)
+                              step=step, t=t, freq=1,
+                              viz_fields=viz_fields)
             raise err
         return state
 
--- a/examples/autoignition-mpi.py
+++ b/examples/autoignition-mpi.py
@@ -242,9 +242,9 @@ def main(ctx_factory=cl.create_some_context, use_leap=False):
     try:
         (current_step, current_t, current_state) = \
             advance_state(rhs=my_rhs, timestepper=timestepper,
-                checkpoint=my_checkpoint,
-                get_timestep=get_timestep, state=current_state,
-                t=current_t, t_final=t_final)
+                          pre_step_callback=my_checkpoint,
+                          get_timestep=get_timestep, state=current_state,
+                          t=current_t, t_final=t_final)
     except ExactSolutionMismatch as ex:
         error_state = True
         current_step = ex.step
--- a/examples/autoignition-mpi.py
+++ b/examples/autoignition-mpi.py
@@ -41,11 +41,11 @@ from mirgecom.simutil import (
     inviscid_sim_timestep,
     sim_visualization,
     sim_checkpoint,
-    cfd_healthcheck,
+    sim_healthcheck,
     check_step,
     generate_and_distribute_mesh
 )
-from mirgecom.exceptions import StepperCrashError
+from mirgecom.exceptions import SynchronizedError
 from mirgecom.fluid import split_conserved
 from mirgecom.io import make_init_message
 from mirgecom.mpi import mpi_entry_point
@@ -238,7 +238,7 @@ def main(ctx_factory=cl.create_some_context, use_leap=False):
         viz_fields = [("reaction_rates", reaction_rates)]
         try:
             # Check the health of the simulation
-            cfd_healthcheck(discr, eos, state,
+            sim_healthcheck(discr, eos, state,
                             step=step, t=t, freq=ncheck)
             # Perform checkpointing
             sim_checkpoint(discr, eos, state,
@@ -249,7 +249,7 @@ def main(ctx_factory=cl.create_some_context, use_leap=False):
                               visualizer, vizname=casename,
                               step=step, t=t, freq=nviz,
                               viz_fields=viz_fields)
-        except StepperCrashError as err:
+        except SynchronizedError as err:
             # Log crash error message
             if rank == 0:
                 logger.info(str(err))
--- a/examples/autoignition-mpi.py
+++ b/examples/autoignition-mpi.py
@@ -243,7 +243,9 @@ def main(ctx_factory=cl.create_some_context, use_leap=False):
 
     def my_simhealthcheck(state, step, t, dt):
         cv = split_conserved(discr.dim, state)
-        sim_healthcheck(discr, eos, q=state, conserved_vars=cv, step=step, t=t)
+        return sim_healthcheck(discr, eos, q=state,
+                               conserved_vars=cv,
+                               step=step, t=t)
 
     try:
         (current_step, current_t, current_state) = \
--- a/examples/autoignition-mpi.py
+++ b/examples/autoignition-mpi.py
@@ -39,12 +39,13 @@ from grudge.shortcuts import make_visualizer
 from mirgecom.euler import euler_operator
 from mirgecom.simutil import (
     inviscid_sim_timestep,
+    sim_visualization,
     sim_checkpoint,
-    sim_healthcheck,
+    sim_cfd_healthcheck,
     check_step,
     generate_and_distribute_mesh
 )
-from mirgecom.exceptions import MirgecomException
+from mirgecom.exceptions import StepperCrashError
 from mirgecom.fluid import split_conserved
 from mirgecom.io import make_init_message
 from mirgecom.mpi import mpi_entry_point
@@ -84,6 +85,7 @@ def main(ctx_factory=cl.create_some_context, use_leap=False):
     constant_cfl = False
     nstatus = 1
     nviz = 5
+    ncheck = 1
     rank = 0
     checkpoint_t = current_t
     current_step = 0
@@ -94,7 +96,6 @@ def main(ctx_factory=cl.create_some_context, use_leap=False):
         timestepper = rk4_step
     box_ll = -0.005
     box_ur = 0.005
-    error_state = False
     debug = False
 
     from mpi4py import MPI
@@ -232,33 +233,37 @@ def main(ctx_factory=cl.create_some_context, use_leap=False):
                 + eos.get_species_source_terms(cv))
 
     def my_checkpoint(step, t, dt, state):
-        cv = split_conserved(dim, state)
-        reaction_rates = eos.get_production_rates(cv)
-        viz_fields = [("reaction_rates", reaction_rates)]
-        return sim_checkpoint(discr, visualizer, eos, q=state,
-                              vizname=casename, step=step,
-                              t=t, dt=dt, nstatus=nstatus, nviz=nviz,
-                              constant_cfl=constant_cfl, comm=comm,
-                              viz_fields=viz_fields)
+        try:
+            # Check the health of the simulation
+            sim_cfd_healthcheck(discr, eos, q=state,
+                                ncheck=ncheck, step=step, t=t)
+            # Perform checkpointing
+            sim_checkpoint(discr, eos, q=state,
+                           exact_soln=initializer,
+                           step=step, t=t, dt=dt,
+                           nstatus=nstatus,
+                           constant_cfl=constant_cfl)
+            # Visualize
+            sim_visualization(discr, state, eos,
+                              visualizer, vizname=casename,
+                              step=step, t=t, nviz=nviz)
+        except StepperCrashError as err:
+            # Log crash error message
+            if rank == 0:
+                logger.info(str(err))
+                logger.info("Visualizing crashed state ...")
+            # Write out crashed field
+            sim_visualization(discr, state, eos,
+                              visualizer, vizname=casename,
+                              step=step, t=t, nviz=1)
+            raise err
+        return state
 
-    def my_simhealthcheck(state, step, t, dt):
-        cv = split_conserved(discr.dim, state)
-        return sim_healthcheck(discr, eos, q=state,
-                               conserved_vars=cv,
-                               step=step, t=t)
-
-    try:
-        (current_step, current_t, current_state) = \
-            advance_state(rhs=my_rhs, timestepper=timestepper,
-                          pre_step_callback=my_checkpoint,
-                          post_step_callback=my_simhealthcheck,
-                          get_timestep=get_timestep, state=current_state,
-                          t=current_t, t_final=t_final)
-    except MirgecomException as ex:
-        error_state = True
-        current_step = ex.step
-        current_t = ex.t
-        current_state = ex.state
+    current_step, current_t, current_state = \
+        advance_state(rhs=my_rhs, timestepper=timestepper,
+                      pre_step_callback=my_checkpoint,
+                      get_timestep=get_timestep, state=current_state,
+                      t=current_t, t_final=t_final, eos=eos, dim=dim)
 
     if not check_step(current_step, nviz):  # If final step not an output step
         if rank == 0:
@@ -266,12 +271,6 @@ def main(ctx_factory=cl.create_some_context, use_leap=False):
         my_checkpoint(current_step, t=current_t,
                       dt=(current_t - checkpoint_t),
                       state=current_state)
-
-    if current_t - t_final < 0:
-        error_state = True
-
-    if error_state:
-        raise ValueError("Simulation did not complete successfully.")
 
 
 if __name__ == "__main__":
--- a/examples/autoignition-mpi.py
+++ b/examples/autoignition-mpi.py
@@ -40,10 +40,12 @@ from mirgecom.euler import euler_operator
 from mirgecom.simutil import (
     inviscid_sim_timestep,
     sim_checkpoint,
+    sim_healthcheck,
     check_step,
-    generate_and_distribute_mesh,
-    ExactSolutionMismatch
+    generate_and_distribute_mesh
 )
+from mirgecom.exceptions import MirgecomException
+from mirgecom.fluid import split_conserved
 from mirgecom.io import make_init_message
 from mirgecom.mpi import mpi_entry_point
 
@@ -52,7 +54,7 @@ from mirgecom.steppers import advance_state
 from mirgecom.boundary import AdiabaticSlipBoundary
 from mirgecom.initializers import MixtureInitializer
 from mirgecom.eos import PyrometheusMixture
-from mirgecom.euler import split_conserved
+
 import cantera
 import pyrometheus as pyro
 
@@ -239,13 +241,18 @@ def main(ctx_factory=cl.create_some_context, use_leap=False):
                               constant_cfl=constant_cfl, comm=comm,
                               viz_fields=viz_fields)
 
+    def my_simhealthcheck(state, step, t, dt):
+        cv = split_conserved(discr.dim, state)
+        sim_healthcheck(discr, eos, q=state, conserved_vars=cv, step=step, t=t)
+
     try:
         (current_step, current_t, current_state) = \
             advance_state(rhs=my_rhs, timestepper=timestepper,
                           pre_step_callback=my_checkpoint,
+                          post_step_callback=my_simhealthcheck,
                           get_timestep=get_timestep, state=current_state,
                           t=current_t, t_final=t_final)
-    except ExactSolutionMismatch as ex:
+    except MirgecomException as ex:
         error_state = True
         current_step = ex.step
         current_t = ex.t
--- a/examples/scalar-lump-mpi.py
+++ b/examples/scalar-lump-mpi.py
@@ -149,9 +149,9 @@ def main(ctx_factory=cl.create_some_context, use_leap=False):
     try:
         (current_step, current_t, current_state) = \
             advance_state(rhs=my_rhs, timestepper=timestepper,
-                checkpoint=my_checkpoint,
-                get_timestep=get_timestep, state=current_state,
-                t=current_t, t_final=t_final)
+                          pre_step_callback=my_checkpoint,
+                          get_timestep=get_timestep, state=current_state,
+                          t=current_t, t_final=t_final)
     except ExactSolutionMismatch as ex:
         current_step = ex.step
         current_t = ex.t
--- a/examples/scalar-lump-mpi.py
+++ b/examples/scalar-lump-mpi.py
@@ -42,7 +42,8 @@ from mirgecom.simutil import (
     inviscid_sim_timestep,
     sim_visualization,
     sim_checkpoint,
-    sim_cfd_healthcheck,
+    cfd_healthcheck,
+    compare_with_analytic_solution,
     generate_and_distribute_mesh
 )
 from mirgecom.exceptions import StepperCrashError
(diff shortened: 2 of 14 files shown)


mirgecom/steppers.py
```diff
@@ -59,6 +58,17 @@ def advance_state(rhs, timestepper, checkpoint, get_timestep,
...
+        An optional user-defined function to be called before the timestepper
+        is called for that particular step. A non-zero return code from this
+        function indicates that this function should stop gracefully.
+        Examples of such functions include visualization, io, or restart
+        checkpoints.
+    post_step_callback
+        An optional user-defined function to be called after the timestepper
+        is called for that particular step. A non-zero return code from this
+        function indicates that this function should stop gracefully.
+        Examples of such functions include applying modal filtering or limiters.
```
inducer, line 71: Document signature of the callbacks.

inducer, line 71: Can we get by with one callback?

inducer, line 71: Modifying state would entail returning it. It's also a dangerous thing to do in the setting of multi-step time steppers which maintain their own history.

inducer (commented):
Thanks for working on this! Some thoughts below.

mirgecom/steppers.py
thomasgibson: There are some callback routines that would you definitely want to apply after the timestepper though. Like modal filtering

inducer: Maybe... ultimately the distinction only matters for the first (and possibly last) step though. I could be talked into believing that two is more correct.

MTCam: Having pre and post hooks available have been useful for me in other projects where we were coupling several solvers and other components together. Although this infrastructure is (currently) monolithic, and more flexible than the code in those other projects, my intuition tells me both would be useful.  

That said, at the moment if we just had the pre-hook, it would be fine.  We don't have a real use for the post-hook atm, and it is easy to add.  I'm ok with doing away with the post-hook if we'd rather not have it.

MTCam: Agreed, and we need to take care with that.  But some of the things we've done in the (near and distant) past involve just that: actuators or filters that modify the state between steps.

These user hooks could be a nice way for the users to test different sorts of filters or controls in optimization problems.  As @inducer notes, we'd need to allow the callback to modify the state and return it and that's not in the signature (yet).  And folks would need to be mindful about what sorts of timestepping would make sense.

thomasgibson: @inducer How are we supposed to apply limiters then? These modify the solution. I don't see a way around that. I agree we should be careful though.

inducer: I'm just grumpy that we need a facility like this. I'm not arguing that we shouldn't have one. :) The one substantive bit to my comment is that the callback should return the modified solution, since in-place modification is not a thing.

inducer: Sure, let's keep both.

mirgecom/exceptions.py
```diff
@@ -0,0 +1,78 @@
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+"""
+
+
+class MirgecomException(Exception):
```
majosm, line 33: I dunno about calling this `MirgecomException`, since it's kind of specific to errors that occur inside the timestepper. `StepperError`?

(BTW, in #257, the need to store `step`/`t`/`state` inside an exception goes away.)

mirgecom/simutil.py
```diff
@@ -162,6 +148,40 @@ def sim_checkpoint(discr, visualizer, eos, q, vizname, exact_soln=None,
         raise ExactSolutionMismatch(step, t=t, state=q)
 
 
+def sim_healthcheck(discr, eos, q, conserved_vars, step=0, t=0):
```
majosm, line 151: We'll have to generalize this so it's not hard-coded to fluids (or otherwise make separate versions for different physics?), but maybe we can do that in #257 along with what's already there.

mirgecom/simutil.py
```diff
@@ -162,6 +148,40 @@ def sim_checkpoint(discr, visualizer, eos, q, vizname, exact_soln=None,
...
+    import grudge.op as op
+
+    # NOTE: Derived quantities are functions of the conserved variables.
+    # Therefore is it sufficient to check for unphysical values of
+    # temperature and pressure.
+    dependent_vars = eos.dependent_vars(conserved_vars)
+
+    # Check for NaN
+    if (np.isnan(op.nodal_sum(discr, "vol", dependent_vars.pressure))
+            or np.isnan(op.nodal_sum(discr, "vol", dependent_vars.temperature))):
```
majosm, line 162: Need to synchronize the error condition across all ranks before raising, otherwise the code may hang. [Possibly useful snippet](https://github.com/illinois-ceesd/mirgecom/blob/e47d3d0b39b4782c05ef3d27d8cc63ceb84c47a3/mirgecom/mpi.py#L161-L166)

mirgecom/steppers.py
```diff
@@ -59,6 +58,17 @@ def advance_state(rhs, timestepper, checkpoint, get_timestep,
         Time at which to start
     istep: int
         Step number from which to start
+    pre_step_callback
+        An optional user-defined function to be called before the timestepper
+        is called for that particular step. A non-zero return code from this
+        function indicates that this function should stop gracefully.
+        Examples of such functions include visualization, io, or restart
+        checkpoints.
```
majosm, line 66: Ultimately I think checkpointing will be better off inside `post_step_callback`, but switching that up is probably beyond the scope of this PR.

mirgecom/steppers.py
```diff
@@ -80,11 +90,16 @@ def advance_state(rhs, timestepper, checkpoint, get_timestep,
...
-        checkpoint(state=state, step=istep, t=t, dt=dt)
+        if pre_step_callback is not None:
+            pre_step_callback(state=state, step=istep, t=t, dt=dt)
 
         state = timestepper(state=state, t=t, dt=dt, rhs=rhs)
 
         t += dt
+
+        if post_step_callback is not None:
+            post_step_callback(state=state, step=istep, t=t, dt=dt)
```
majosm, line 101: I would advocate for having `post_step_callback` (and maybe `pre_step_callback` too?) return a boolean indicating whether to terminate timestepping. It gives the user some more flexibility in deciding when the simulation should stop. Then the timestepping loop can be a `while not done:` and the `t_final` stuff can be removed from the interface. I don't know if we want to actually make that change to the stepper in this PR, but in any case, just something to think about.

majosm (commented):
Added a few thoughts based on the work that's been done in #257 (which AFAIK is not dead, just dormant 🙂).

mirgecom/steppers.py
thomasgibson: Done: https://github.com/illinois-ceesd/mirgecom/pull/370/commits/b914d3dd258983b3bbf07a5e315a85579b5565bc

thomasgibson: https://github.com/illinois-ceesd/mirgecom/pull/370/commits/b914d3dd258983b3bbf07a5e315a85579b5565bc

mirgecom/steppers.py
thomasgibson: @majosm I actually think that shouldn't be hard at all; it would just mean you pass your checkpoint function to the `post_step_callback` arg. Or am I missing something?

mirgecom/simutil.py
thomasgibson: @majosm Do you want to maybe talk sometime about combining this PR with yours? I haven't actually looked at #257 in great detail, so maybe there's an opportunity to revive it?

mirgecom/steppers.py
majosm: > @majosm I actually think that shouldn't be hard at all; it would just mean you pass your checkpoint function to the `post_step_callback` arg. Or am I missing something?

IIRC there were some complications with making sure that you get one and only one checkpoint call per step (including step 0), especially when restarts are involved. I don't remember if this was fully resolved or not in #257.

mirgecom/steppers.py
thomasgibson: I like this idea. How does a call signature like this look? `state, stop = callback(...)`, where `stop` is a bool?

thomasgibson: What about callbacks that raise exceptions though? That's my only real concern.

mirgecom/simutil.py
majosm: I suspect there will still be a fair amount of back and forth with that PR when it gets revived, so maybe it would be best to try to get a basic health check functionality into main with this PR and then we can integrate that into #257 later?

mirgecom/simutil.py
thomasgibson: Good point. Question though: Is this okay if `op.nodal_sum` and other reductions are distributed? Once https://github.com/inducer/grudge/pull/117 lands, these (and similar reductions like norms) will do parallel reductions. Or would we rather do things rank-local in these callbacks?

majosm: Oh, neat. Yeah, I think if `op.nodal_sum` is global then it should be fine. It's probably going to be slower than doing a single reduction, but we can always optimize it later if it becomes a problem.

examples/vortex-mpi.py
```diff
@@ -183,14 +185,22 @@ def my_checkpoint(step, t, dt, state):
...
+    def my_simhealthcheck(state, step, t, dt):
+        cv = split_conserved(discr.dim, state)
+        return sim_healthcheck(discr, eos, q=state,
+                               conserved_vars=cv,
+                               step=step, t=t)
+
     try:
         (current_step, current_t, current_state) = \
             advance_state(rhs=my_rhs, timestepper=timestepper,
-                checkpoint=my_checkpoint,
```
MTCam, line 189: These examples need to keep `my_checkpoint` as the call-back.   Can't we just call my_simhealthcheck from my_checkpoint?

majosm: > These examples need to keep `my_checkpoint` as the call-back. Can't we just call my_simhealthcheck from my_checkpoint?

https://github.com/illinois-ceesd/mirgecom/blob/9cbb9883d33069e6ab34778cc90045db651ab700/examples/vortex-mpi.py#L197

mirgecom/steppers.py
MTCam: I would advocate for placing the simulation termination in the hands of the user at the callback level.  Consider this callback:

```python
def my_callback(state, step, t, dt):
   is_healthy = my_healthcheck(state, step, t, dt)
   is_healthy = comm.allreduce(is_healthy, MPI.LOR)
   if not is_healthy:
      write_viz(...)
      write_restart(...)
      do_my_other_biddings(...)
      raise SimError("Simulation health check failed.")
```

This is safe because the callback routine is already collective. If we are concerned with users doing collective operations (I don't see why we should be) - then we can add a wrapper.  I like this a lot because it places healthcheck at the driver level, places full control over how to handle a health issue in the hands of the user (correctly imo), and forces users to opt-in to use it.

mirgecom/steppers.py
MTCam: If you have something separate in-place to make your time0 dumps, then having checkpoint as post-step callback is OK, but I don't see the advantage or the need.   Checkpoint is currently pre-step because users often want to dump viz files before stepping, and simulation health check gets a change to stop the simulation before stepping occurs with a bad state.

examples/vortex-mpi.py
MTCam: Thanks, I missed that!   Typically, users want to perform the health check at checkpoint time (i.e. before stepping occurs).  We don't want to step if we have a bad state. We want to dump, and then stop.  So I would still add health check as a utility that I call from checkpoint.

thomasgibson: @MTCam The only reason I put the health check as a `post_step_callback` was to exercise that branch in the stepper :laughing: But I totally get what you're saying. Would you prefer that I merge these into a single callback?

mirgecom/steppers.py
majosm: > I would advocate for placing the simulation termination in the hands of the user at the callback level. Consider this callback:
> 
> ```python
> def my_callback(state, step, t, dt):
>    is_healthy = my_healthcheck(state, step, t, dt)
>    is_healthy = comm.allreduce(is_healthy, MPI.LOR)
>    if not is_healthy:
>       write_viz(...)
>       write_restart(...)
>       do_my_other_biddings(...)
>       raise SimError("Simulation health check failed.")
> ```
> 
> This is safe because the callback routine is already collective. If we are concerned with users doing collective operations (I don't see why we should be) - then we can add a wrapper. I like this a lot because it places healthcheck at the driver level, places full control over how to handle a health issue in the hands of the user (correctly imo), and forces users to opt-in to use it.

I think I mostly agree with this. My only complaint is that when doing it this way the user won't get any information back from `my_healthcheck` about what exactly failed. My inclination would be to make `my_healthcheck` collective and then either:

1. Print out what failed on rank 0 inside `my_healthcheck`
2. Do something like:

```python
def my_callback(state, step, t, dt):
   try:
      my_healthcheck(state, step, t, dt)
   except HealthCheckError:
      write_viz(...)
      write_restart(...)
      do_my_other_biddings(...)
      raise
```
(i.e., whatever is contained in the HealthCheckError will bubble back up to the user.)

examples/vortex-mpi.py
MTCam: I agree with and completely sympathize with why you put it as a post-call.  In my opinion, we should make the examples in the repo look like what we want the users to do.  Users will almost 100% just copy existing example drivers and tweak them. 

If we craft the examples in a way that makes the most sense for production drivers, then we'll automatically get that practice into the DNA of the simulations we need to support.  

The post hook could be exercised by something that looks more like what we'd expect the users to use it for.  Post-step filtering would probably work perfect here.

mirgecom/steppers.py
MTCam: > My only complaint is that when doing it this way the user won't get any information back from `my_healthcheck` about what exactly failed.

Agreed. If we have a healthcheck utility, say in `simutils`, then it could potentially return a string about what went wrong. `is_healthy, status_message = my_healthcheck(...)`.  Just another alternative.  I like the idea of handing that string to the user and letting them figure out what to do with it at the driver level... but that's just me.

thomasgibson: If we catch the exception using what @majosm proposed (try - except clause), then we can yank out the error message from the exception and print that out on rank 0. How does that sound?

mirgecom/steppers.py
majosm: > If you have something separate in-place to make your time0 dumps, then having checkpoint as post-step callback is OK, but I don't see the advantage or the need. Checkpoint is currently pre-step because users often want to dump viz files before stepping, and simulation health check gets a change to stop the simulation before stepping occurs with a bad state.

IIRC doing it before the step led to duplicated checkpoints on restart, e.g.:

```
....
----- start step N -----
checkpoint
advance
----- end step N -------
final checkpoint
===== end run 1 ========
===== start run 2 ======
----- start step N+1 ---
checkpoint  # oops, same as final checkpoint from run 1
advance
----- end step N+1 -----
```

Easier to handle this with post-step checkpoints + a step 0 checkpoint IMO.

mirgecom/steppers.py
MTCam: > What about callbacks that raise exceptions though? That's my only real concern.

I think exceptions from callbacks are OK if the user handles it right.  Callbacks are collective already.  Clearly, the user can toss an exception from only one rank - but hey: *that's on the user*.

mirgecom/simutil.py
MTCam: > We'll have to generalize this so it's not hard-coded to fluids (or otherwise make separate versions for different physics?), but maybe we can do that in #257 along with what's already there.

I think health checks necessarily need to have knowledge of the operator-specific physics. Maybe not if all it does it check for NAN/INF, but otherwise... yeah.  Checkpoint is similar, as it was initially designed to be the fluid health check.  Since we have viz (awkwardly) all wrapped up in the `simutils::checkpoint`... it's awkward to generalize.  #257 is showing some promise for a path out of this :)

mirgecom/steppers.py
MTCam: > (...)
> IIRC doing it before the step led to duplicated checkpoints on restart, e.g.:
> (...)
> Easier to handle this with post-step checkpoints + a step 0 checkpoint IMO.

Thumbs up, but also further motivation for getting viz and healthcheck detangled from simutils checkpoint (not that healthcheck is tangled [yet!]).  I think simutils checkpoint could easily just be rolled into a `fluids_checkpoint` and then call it from the user's `my_checkpoint`, then we could have a similar thing for each (non-fluid) operator too.  I think we were discussing something along those lines in #257 before.

mirgecom/steppers.py
MTCam: > If we catch the exception using what @majosm proposed (try - except clause), then we can yank out the error message from the exception and print that out on rank 0. How does that sound?

I think that is OK, but imo not as powerful as forming a status string and letting the user deal with the error.  It means that the healthcheck does more than just check the health - it also does I/O on rank 0 - which may or may not be what the user wants to do there.

majosm: > > If we catch the exception using what @majosm proposed (try - except clause), then we can yank out the error message from the exception and print that out on rank 0. How does that sound?
> 
> I think that is OK, but imo not as powerful as forming a status string and letting the user deal with the error. It means that the healthcheck does more than just check the health - it also does I/O on rank 0 - which may or may not be what the user wants to do there.

Sorry, I meant doing one or the other, i.e., _either_ have `my_healthcheck` print, _or_ do the `try`/`except` thing (my preference is the latter). With the latter approach you could extract the error message by doing something like:

```python
try:
    my_healthcheck(state, step, t, dt)
except HealthCheckError as e:
    msg = e.message
    # ... do more processing with msg
```

MTCam: > Sorry, I meant doing one or the other, i.e., _either_ have `my_healthcheck` print, _or_ do the `try`/`except` thing (my preference is the latter). With the latter approach you could extract the error message by doing something like:
> 
> ```python
> try:
>     my_healthcheck(state, step, t, dt)
> except HealthCheckError as e:
>     msg = e.message
>     # ... do more processing with msg
> ```

I like this OK. What I would like to avoid is needing to pass "comm" as an argument to the healthcheck utility.  So if `my_healthcheck` essentially calls a library routine underneath but still does synchronization (i.e. collective error handling) at driver level, then I think that is better than doing collective health check and collective exception in the lower layer (i.e. in the library or simutils).

examples/scalar-lump-mpi.py
```diff
@@ -141,21 +144,37 @@ def my_rhs(t, state):
...
+            # Log crash error message
+            if rank == 0:
+                logger.info(str(err))
+                logger.info("Visualizing crashed state ...")
+            # Write out crashed field
+            sim_visualization(discr, state, eos,
+                              visualizer, vizname=casename,
+                              step=step, t=t, nviz=1)
+            raise err
+        return state
```
thomasgibson, line 171: @MTCam @majosm How does this look to you both? (I split out visualization from checkpoint)

mirgecom/steppers.py
thomasgibson: > Thumbs up, but also further motivation for getting viz and healthcheck detangled from simutils checkpoint

I believe this accomplishes that with minimal pain

mirgecom/simutil.py
thomasgibson: @majosm I've opted to keep it rank-local for now, as exceptions are treated globally right? The error conditions now call out to rank-local grudge routines. Does this sound okay? I just felt doing 3 allreduces in a callback might be a bit much

examples/autoignition-mpi.py
```diff
@@ -233,23 +236,37 @@ def my_checkpoint(step, t, dt, state):
...
+            # Perform checkpointing
+            sim_checkpoint(discr, eos, state,
+                           step=step, t=t, dt=dt, freq=nstatus,
+                           constant_cfl=constant_cfl)
+            # Visualize
+            sim_visualization(discr, eos, state,
+                              visualizer, vizname=casename,
+                              step=step, t=t, freq=nviz,
+                              viz_fields=viz_fields)
+        except StepperCrashError as err:
```
thomasgibson, line 252: @majosm @MTCam here's the updated proposal

MTCam: Cool, thanks @thomasgibson.  I like how you've split these functions out.  This work interacts strongly with that of #257.

 - I dunno about StepperCrashError for these utilities. I do see @majosm's comment about this earlier, but these aren't really stepper errors. I think it is a little confusing to toss stepper errors from these functions that are (on the surface) unrelated to the actual stepping and stepper.
 - I'm not a giant fan of having these utilities check whether they will do something on the inside (via step/freq).  I think it is a fairly good policy to make the routine very dumb; if it is called, it does the thing.  Let the caller worry about if it is time to invoke the function. The upshot/question is: Any objections to putting the `check_step` call on the *outside*? 
 - I would be in favor of renaming these utilities.  I like `cfd_healthcheck` but would maybe change that to `fluid_healthcheck`. The other two are fluid-specific too. As long as we're changing them, we may as well change `sim_visualization` to `fluid_visualization`.  
 - `sim_checkpoint` is basically eliminated by this change; we can go back later and reshape it to be general in #257, or banish it forever.  If we banish, we need to add soln watch via `logpyle` to the examples that lack it.

MTCam: We do need to write some tests that expose some of the stuff that we are concerned about for error handling.  I don't have any specific good ideas about how to test these atm, but some of the stuff we've talked about:

- Asynchronous errors during stepping
- Viz file collisions/overwrites
- Graceful exit after any exceptions
- Handling exact soln mismatch collectively

Some of this may be firmly in #257 territory.

mirgecom/simutil.py
MTCam: It is my understanding that exception handling *is not* global.  I think that's the main crux of our struggle with parallel/asynchronous error handling.  It appears to me, that if your current version of the healthcheck fails, then the simulation will hang or behave pathologically (as an exception gets tossed on one rank only, that rank goes into the handling branch, the rest don't).

I don't think we need a bunch of reductions though, really only one:

```python
def fluid_healthcheck(...):
   nans_present = check_for_local_naninf()
   if nans_present:
      message += "NANs or INFs present in solution"
   health_problem = check_for_whatever_local_problem(...):
   if healt_problem:
      message += "Some other health issue."
   health = nansyes | healthproblem | yadda | yadda | yadda
   health = comm.allreduce(health, MPI.LOR)
   if health:
      log(message)
      toss TheException(message)
```

We could probably do better with the actual message. Consider that we could have Nranks disparate messages, each with valid (and potentially useful) error content. If we use `log` or maybe a similar utility that we ourselves roll to be parallel-smart, then we can capture such messages.

majosm: > It is my understanding that exception handling _is not_ global. I think that's the main crux of our struggle with parallel/asynchronous error handling. It appears to me, that if your current version of the healthcheck fails, then the simulation will hang or behave pathologically (as an exception gets tossed on one rank only, that rank goes into the handling branch, the rest don't).

Yeah. One minor clarification for @thomasgibson: if an exception is raised and doesn't get handled anywhere, it will bubble up to the top and cause mpi4py to call `MPI_Abort` (this is what `python -m mpi4py` does). In that situation it's fine for exceptions to occur locally. It's only when we try to handle them and do I/O, etc. prior to exiting that we need to synchronize globally.

examples/autoignition-mpi.py
majosm: To give some background here @thomasgibson: the main reason `sim_checkpoint` exists is to reduce the amount of duplicated code in the drivers. Ideally we want the code inside `my_checkpoint` to be pretty simple (unless the user wants to do something extra that's specific to their simulation). So, with that motivation in mind, I wonder if there's a way we could move the exception handling part inside `sim_checkpoint`? Maybe something along the lines of:

```python
# driver.py

def healthcheck(step, t, state):
    return cfd_healthcheck(...)

def write_vis(step, t, state):
    return sim_visualization(...)

def checkpoint(step, t, dt, state):
    return sim_checkpoint(..., healthcheck=healthcheck, write_vis=write_vis)
```

```python
# simutil.py

def sim_checkpoint(..., healthcheck=None, write_vis=None):
    try:
        if check_step(...):
            # ... write status ...

        if check_step(...) and healthcheck is not None:
            healthcheck(step, t, state)

        if check_step(...) and write_vis is not None:
            write_vis(step, t, state)

    except StepperCrashError as e:
        # ... log error ...

        if write_vis is not None:
            write_vis(step, t, state)

        raise
```

> * I dunno about StepperCrashError for these utilities. I do see @majosm's comment about this earlier, but these aren't really stepper errors. I think it is a little confusing to toss stepper errors from these functions that are (on the surface) unrelated to the actual stepping and stepper.

IMO the name made sense when the `try`/`except` was wrapped around `advance_state`, but now that it's inside the checkpoint I think it could be named something else. Maybe the base class can just be something low level that indicates that it's raised globally? `SynchronizedError`? 🤷‍♂️ 

> * I'm not a giant fan of having these utilities check whether they will do something on the inside (via step/freq).  I think it is a fairly good policy to make the routine very dumb; if it is called, it does the thing.  Let the caller worry about if it is time to invoke the function. The upshot/question is: Any objections to putting the `check_step` call on the _outside_?

I think this depends on whether the code in question is designed to maximize flexibility or simplicity. `sim_checkpoint` is more intended to do the latter, so I think it makes sense to wrap the `check_step` calls inside it.

> * I would be in favor of renaming these utilities.  I like `cfd_healthcheck` but would maybe change that to `fluid_healthcheck`. The other two are fluid-specific too. As long as we're changing them, we may as well change `sim_visualization` to `fluid_visualization`.

I think I would vote for just calling it `sim_healthcheck` for the time being and fixing it in #257.

MTCam: > I think this depends on whether the code in question is designed to maximize flexibility or simplicity. `sim_checkpoint` is more intended to do the latter, so I think it makes sense to wrap the `check_step` calls inside it.
> 

imo, removing the check both simplifies and increases flexibility (the way it was before). I agree that if the `sim_checkpoint` ends up wrapping all the checks, that `sim_checkpoint` needs to perform the step checks, but I'd still argue that the call to `check_step` should be at the call site to `sim_healthcheck` and not down inside of it. But... this is also a very mild and minor complaint.  It has been down inside the routine all this time.

> > * I would be in favor of renaming these utilities.  I like `cfd_healthcheck` but would maybe change that to `fluid_healthcheck`. The other two are fluid-specific too. As long as we're changing them, we may as well change `sim_visualization` to `fluid_visualization`.
> 
> I think I would vote for just calling it `sim_healthcheck` for the time being and fixing it in #257.

👍

thomasgibson: @majosm @MTCam I've pushed a change that renames `cfd_healthcheck`, and I've renamed the `StepperCrashError` according to @majosm's comment: https://github.com/illinois-ceesd/mirgecom/pull/370/commits/2523fbdfa581c0c86f0e4d21691aa345a64d185b.

Is there anything else you'd like for me to apply before leaving the rest to https://github.com/illinois-ceesd/mirgecom/pull/257?

MTCam: Is there a good reason that `sim_checkpoint` does not make a crash viz dump at exception time for serial runs?

I'm not going to pursue this complaint very far - but I think that we are going backwards a bit here.  This works essentially strips the user of a good portion of the control over what happens when the health check fails.   Health checks are meant to be a user-defined facility.  Imo, *MIRGE-Com* should provide simple utilities to perform health checks, but it should be up to the user to handle what happens if any of those checks are failed.   For me, it is a step backwards to roll a new monolithic "checkpoint driver" that tries to do and automate too much.

majosm: > Is there a good reason that `sim_checkpoint` does not make a crash viz dump at exception time for serial runs?

?

> I'm not going to pursue this complaint very far - but I think that we are going backwards a bit here. This works essentially strips the user of a good portion of the control over what happens when the health check fails. Health checks are meant to be a user-defined facility. Imo, _MIRGE-Com_ should provide simple utilities to perform health checks, but it should be up to the user to handle what happens if any of those checks are failed. For me, it is a step backwards to roll a new monolithic "checkpoint driver" that tries to do and automate too much.

Sounds like we don't have a consensus about what we're actually trying to do here... i.e., are we trying to maximize user control, or minimize code? They conflict with each other to an extent. I was under the impression that we were aiming more for the latter.

I think it would make sense to have the user pass in a health check function to `sim_checkpoint`. I'm less sure I see the value in having them handle errors manually. Can you give an example?

`sim_checkpoint` in its current state here does raise the exceptions again after handling them, so I guess theoretically a user could do:

```python
try:
    sim_checkpoint(...)
except HealthCheckError as e:
    # ... additional handling ...

```

> Edit: actually maybe this adding of callback routines to the stepper has had a bit a scope creep, too. refactoring the checkpoint junk is a good deal of the business in #257.

I'm not sure if the original goal of this PR was to add a basic health check or just the callback functionality and nothing else, but if we want to add a health check here I think we have to deal with at least some of the stuff from #257 to make it work.

majosm: > > Is there a good reason that `sim_checkpoint` does not make a crash viz dump at exception time for serial runs?
> 
> ?

Never mind, I see it now:

```python
    if comm is None:
        if terminate:
            raise exception
        return
```

Yeah, need a vis dump here @thomasgibson. Or maybe you could do something like:

```python
    if comm is not None:
        from mpi4py import MPI
        terminate = comm.allreduce(terminate, MPI.LOR)

    if terminate:
        # ...
```
to avoid multiple returns?

mirgecom/simutil.py
```diff
@@ -74,92 +88,273 @@ def inviscid_sim_timestep(discr, state, t, dt, cfl, eos,
...
+        return
+
+    from mpi4py import MPI
+
+    terminate = comm.allreduce(terminate, MPI.LOR)
+
+    if terminate:
+        # Log crash error message
+        if rank == 0:
+            logger.info(str(exception))
```
majosm, line 243: I think this will need to be tweaked, because `exception` doesn't necessarily exist on rank 0.

mirgecom/simutil.py
```diff
@@ -74,92 +88,273 @@ def inviscid_sim_timestep(discr, state, t, dt, cfl, eos,
...
+            logger.info("Visualizing crashed state ...")
+        # Write out crashed field
+        sim_visualization(discr, eos, q,
+                          visualizer, vizname=vizname,
+                          step=step, t=t,
+                          viz_fields=viz_fields)
+        raise exception
```
majosm, line 250: This too, since `exception` doesn't exist on all ranks.

This one is a little trickier to deal with I think... Seems like we can either:

1. Only re-raise the exception on the rank that originally raised it. I don't know if I like this since it makes any subsequent handling have to redo the MPI reduction.
2. Try to pickle the exception and broadcast it to all ranks. I vaguely remember trying this and running into issues, but unfortunately I can't recall the details.
3. Synchronize the error state before raising and then do the `raise` on all ranks.

Thoughts?

examples/autoignition-mpi.py
MTCam: > Sounds like we don't have a consensus about what we're actually trying to do here... i.e., are we trying to maximize user control, or minimize code? They conflict with each other to an extent. I was under the impression that we were aiming more for the latter.
> 

I agree with that.  My suggestion to add the callback hooks was centered on the need to give the user more freedom to implement what they want in the callbacks - one of those things likely to be health checking.  I wanted to let the user decide what to do once he had evaluated the state.

> I think it would make sense to have the user pass in a health check function to `sim_checkpoint`. I'm less sure I see the value in having them handle errors manually. Can you give an example?
> 

Health check is exactly the example, I think.  The user should be able to compose whatever health check they want.  In fact, lets say the user wants to monitor specific volume (because, for example, it tends to go wild at an injection inlet or something).  Then the user writes that health check function (or uses a built-in utility), and then ... if the check triggers, the user can either give up and quit, or turn on relaxation source terms to combat the issue.

> I'm not sure if the original goal of this PR was to add a basic health check or just the callback functionality and nothing else, but if we want to add a health check here I think we have to deal with at least some of the stuff from #257 to make it work.

It was my understanding that this PR was just trying to add the callbacks, however I see how its scope expanded, and I'm not necessarily opposed. I also want to make clear that my complaint about the structure and user control is not really a major one (from my perspective).  We can make use of it how it is, but it has evolved to the point that it doesn't really accomplish the things I had in mind for introducing the callbacks, and imo tries to do too much with automated health checking.

mirgecom/simutil.py
thomasgibson: @majosm I think option 3 (which I believe the code now sort of does) is probably the easiest for now? I don't want this PR to dig _too_ deeply into this. Are there any major objections to doing this: https://github.com/illinois-ceesd/mirgecom/pull/370/commits/4c938fa2cb2eef71d83aff54792fc14ad20b86f5?

MTCam: > To do option 3 properly I think we would need to synchronize before raising for each check inside `sim_healthcheck` (...)

Why not something along the lines of this that we talked about earlier?

```python
def fluid_healthcheck(...):
   nans_present = check_for_local_naninf()
   if nans_present:
      message += "NANs or INFs present in solution"
   health_problem = check_for_whatever_local_problem(...):
   if healt_problem:
      message += "Some other health issue."
   health = nansyes | healthproblem | yadda | yadda | yadda
   health = comm.allreduce(health, MPI.LOR)
   if health:
      log(message)
      toss TheException(message)
```

majosm: > > To do option 3 properly I think we would need to synchronize before raising for each check inside `sim_healthcheck` (...)
> 
> Why not something along the lines of this that we talked about earlier?
> 
> ```python
> def fluid_healthcheck(...):
>    nans_present = check_for_local_naninf()
>    if nans_present:
>       message += "NANs or INFs present in solution"
>    health_problem = check_for_whatever_local_problem(...):
>    if healt_problem:
>       message += "Some other health issue."
>    health = nansyes | healthproblem | yadda | yadda | yadda
>    health = comm.allreduce(health, MPI.LOR)
>    if health:
>       log(message)
>       toss TheException(message)
> ```

Oops, yeah, I didn't mean to exclude this. I just meant that the error state needs to be synchronized before exceptions are raised.

thomasgibson:
closing in favor of https://github.com/illinois-ceesd/mirgecom/pull/381

PR #376, opened 2021-06-09
Tweaks to error handling in #370

majosm:
This PR tweaks the error handling setup in #370 in the following ways:

1. Converts `SynchronizedError` from a base class to a type that wraps another exception. `SynchronizedException` now indicates to the handling code that the contained exception has been synchronized across MPI ranks.
2. Adds some helper code for synchronizing exceptions (`bcast_from_lowest_rank`).
3. Moves the `try`/`except` from `sim_checkpoint` into `advance_state`.
4. Adds an `exception_callback` to `advance_state`. This function can either re-raise the exception after handling, or it can return a `state, stop` tuple causing stepping to either resume or halt according to what the user decides.

In the process I also added a couple of callbacks to `sim_checkpoint` (`healthcheck_callback` and `vis_callback`).

(Note: I only changed `vortex-mpi` for now, so the CI will probably fail.)

Thoughts @thomasgibson, @MTCam?

### DIFF
--- a/examples/vortex-mpi.py
+++ b/examples/vortex-mpi.py
@@ -40,6 +40,8 @@ from mirgecom.profiling import PyOpenCLProfilingArrayContext
 from mirgecom.euler import euler_operator
 from mirgecom.simutil import (
     inviscid_sim_timestep,
+    sim_healthcheck,
+    sim_visualization,
     sim_checkpoint,
     generate_and_distribute_mesh
 )
@@ -171,22 +173,39 @@ def main(ctx_factory=cl.create_some_context, use_profiling=False, use_logmgr=Fal
                            dt=current_dt, cfl=current_cfl, eos=eos,
                            t_final=t_final, constant_cfl=constant_cfl)
 
+    def healthcheck(step, t, state):
+        return sim_healthcheck(discr, eos, state, step=step, t=t)
+
+    def write_vis(step, t, state):
+        return sim_visualization(discr, eos, state,
+                                 visualizer, vizname=casename,
+                                 step=step, t=t,
+                                 vis_timer=vis_timer)
+
     def my_rhs(t, state):
         return euler_operator(discr, q=state, t=t,
                               boundaries=boundaries, eos=eos)
 
     def my_checkpoint(step, t, dt, state):
         sim_checkpoint(discr, visualizer, eos, q=state,
-                       exact_soln=initializer, vizname=casename, step=step,
-                       t=t, dt=dt, nstatus=nstatus, nviz=nviz,
+                       exact_soln=initializer, step=step,
+                       t=t, dt=dt, nstatus=nstatus, healthcheck_callback=healthcheck,
+                       nviz=nviz, vis_callback=write_vis,
                        exittol=exittol, constant_cfl=constant_cfl)
         return state
+
+    def handle_exception(step, t, state, exception):
+        if rank == 0:
+            logger.info("Visualizing crashed state ...")
+        write_vis(step, t, state)
+        raise
 
     current_step, current_t, current_state = \
         advance_state(rhs=my_rhs, timestepper=timestepper,
                       pre_step_callback=my_checkpoint,
                       get_timestep=get_timestep, state=current_state,
                       t=current_t, t_final=t_final,
+                      exception_callback=handle_exception,
                       logmgr=logmgr, eos=eos, dim=dim)
 
     if rank == 0:
--- a/examples/vortex-mpi.py
+++ b/examples/vortex-mpi.py
@@ -187,12 +187,12 @@ def main(ctx_factory=cl.create_some_context, use_profiling=False, use_logmgr=Fal
                               boundaries=boundaries, eos=eos)
 
     def my_checkpoint(step, t, dt, state):
-        sim_checkpoint(discr, visualizer, eos, q=state,
-                       exact_soln=initializer, step=step,
-                       t=t, dt=dt, nstatus=nstatus, healthcheck_callback=healthcheck,
-                       nviz=nviz, vis_callback=write_vis,
-                       exittol=exittol, constant_cfl=constant_cfl)
-        return state
+        return sim_checkpoint(discr, visualizer, eos, q=state,
+                              exact_soln=initializer, step=step,
+                              t=t, dt=dt, nstatus=nstatus,
+                              healthcheck_callback=healthcheck,
+                              nviz=nviz, vis_callback=write_vis,
+                              exittol=exittol, constant_cfl=constant_cfl)
 
     def handle_exception(step, t, state, exception):
         if rank == 0:
--- a/mirgecom/steppers.py
+++ b/mirgecom/steppers.py
@@ -101,14 +101,18 @@ def _advance_state_stepper_func(rhs, timestepper, get_timestep,
                 raise SynchronizedException(RuntimeError(f"Invalid timestep {dt}."))
 
             if pre_step_callback is not None:
-                state = pre_step_callback(state=state, step=istep, t=t, dt=dt)
+                state, stop = pre_step_callback(state=state, step=istep, t=t, dt=dt)
+                if stop:
+                    break
 
             state = timestepper(state=state, t=t, dt=dt, rhs=rhs)
 
             t += dt
 
             if post_step_callback is not None:
-                state = post_step_callback(state=state, step=istep, t=t, dt=dt)
+                state, stop = post_step_callback(state=state, step=istep, t=t, dt=dt)
+                if stop:
+                    break
 
             istep += 1
 
@@ -212,9 +216,11 @@ def _advance_state_leap(rhs, timestepper, get_timestep,
                 raise SynchronizedException(RuntimeError(f"Invalid timestep {dt}."))
 
             if pre_step_callback is not None:
-                state = pre_step_callback(state=state,
-                                          step=istep,
-                                          t=t, dt=dt)
+                state, stop = pre_step_callback(state=state,
+                                                step=istep,
+                                                t=t, dt=dt)
+                if stop:
+                    break
 
             # Leap interface here is *a bit* different.
             for event in stepper_cls.run(t_end=t+dt):
@@ -223,9 +229,11 @@ def _advance_state_leap(rhs, timestepper, get_timestep,
                     t += dt
 
                     if post_step_callback is not None:
-                        state = post_step_callback(state=state,
-                                                   step=istep,
-                                                   t=t, dt=dt)
+                        state, stop = post_step_callback(state=state,
+                                                         step=istep,
+                                                         t=t, dt=dt)
+                        if stop:
+                            break
 
                     istep += 1
 
--- a/mirgecom/steppers.py
+++ b/mirgecom/steppers.py
@@ -30,6 +30,7 @@ THE SOFTWARE.
 
 from logpyle import set_dt
 from mirgecom.logging_quantities import set_sim_state
+from mirgecom.exceptions import SynchronizedException
 
 
 def _advance_state_stepper_func(rhs, timestepper, get_timestep,
@@ -37,6 +38,7 @@ def _advance_state_stepper_func(rhs, timestepper, get_timestep,
                                 t=0.0, istep=0,
                                 pre_step_callback=None,
                                 post_step_callback=None,
+                                exception_callback=None,
                                 logmgr=None, eos=None, dim=None):
     """Advance state from some time (t) to some time (t_final).
 
@@ -71,6 +73,12 @@ def _advance_state_stepper_func(rhs, timestepper, get_timestep,
         An optional user-defined function, with signature:
         ``state = post_step_callback(state, step, t, dt)``,
         to be called after the timestepper is called for that particular step.
+    exception_callback
+        An optional user-defined function, with signature:
+        ``state, stop = exception_callback(step, t, state, exception)``,
+        to be called when an exception is caught. Note: only exceptions that are
+        guaranteed to be raised on all MPI ranks (i.e., those contained in a
+        :class:`mirgecom.exceptions.SynchronizedException`) will be caught.
 
     Returns
     -------
@@ -84,30 +92,39 @@ def _advance_state_stepper_func(rhs, timestepper, get_timestep,
         return istep, t, state
 
     while t < t_final:
+        try:
+            if logmgr:
+                logmgr.tick_before()
 
-        if logmgr:
-            logmgr.tick_before()
+            dt = get_timestep(state=state)
+            if dt < 0:
+                raise SynchronizedException(RuntimeError(f"Invalid timestep {dt}."))
 
-        dt = get_timestep(state=state)
-        if dt < 0:
-            return istep, t, state
+            if pre_step_callback is not None:
+                state = pre_step_callback(state=state, step=istep, t=t, dt=dt)
 
-        if pre_step_callback is not None:
-            state = pre_step_callback(state=state, step=istep, t=t, dt=dt)
+            state = timestepper(state=state, t=t, dt=dt, rhs=rhs)
 
-        state = timestepper(state=state, t=t, dt=dt, rhs=rhs)
+            t += dt
 
-        t += dt
+            if post_step_callback is not None:
+                state = post_step_callback(state=state, step=istep, t=t, dt=dt)
 
-        if post_step_callback is not None:
-            state = post_step_callback(state=state, step=istep, t=t, dt=dt)
+            istep += 1
 
-        istep += 1
+            if logmgr:
+                set_dt(logmgr, dt)
+                set_sim_state(logmgr, dim, state, eos)
+                logmgr.tick_after()
 
-        if logmgr:
-            set_dt(logmgr, dt)
-            set_sim_state(logmgr, dim, state, eos)
-            logmgr.tick_after()
+        except SynchronizedException as synced_exception:
+            if exception_callback is not None:
+                state, stop = exception_callback(istep, t, state,
+                    synced_exception.exception)
+                if stop:
+                    break
+            else:
+                raise
 
     return istep, t, state
 
@@ -118,6 +135,7 @@ def _advance_state_leap(rhs, timestepper, get_timestep,
                         t=0.0, istep=0,
                         pre_step_callback=None,
                         post_step_callback=None,
+                        exception_callback=None,
                         logmgr=None, eos=None, dim=None):
     """Advance state from some time *t* to some time *t_final* using :mod:`leap`.
 
@@ -152,6 +170,12 @@ def _advance_state_leap(rhs, timestepper, get_timestep,
         An optional user-defined function, with signature:
         ``state = post_step_callback(state, step, t, dt)``,
         to be called after the timestepper is called for that particular step.
+    exception_callback
+        An optional user-defined function, with signature:
+        ``state, stop = exception_callback(step, t, state, exception)``,
+        to be called when an exception is caught. Note: only exceptions that are
+        guaranteed to be raised on all MPI ranks (i.e., those contained in a
+        :class:`mirgecom.exceptions.SynchronizedException`) will be caught.
 
     Returns
     -------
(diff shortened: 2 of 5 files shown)


mirgecom/steppers.py
```diff
@@ -164,41 +188,60 @@ def _advance_state_leap(rhs, timestepper, get_timestep,
...
-        if dt < 0:
-            return istep, t, state
+        stepper_cls = generate_singlerate_leap_advancer(timestepper, component_id,
+                                                        rhs, t, dt, state)
+    except SynchronizedException as synced_exception:
+        if exception_callback is not None:
+            state, stop = exception_callback(istep, t, state,
+                synced_exception.exception)
+            if stop:
+                return istep, t, state
```
thomasgibson, line 201: Why is this not `break` like here: https://github.com/illinois-ceesd/mirgecom/pull/376/files#diff-e13bce10f3dc4e997e0c56c91b3d0309d0ebf5ceaecd41ae16bca424c91ff24fR125?

thomasgibson (commented):
Thanks! I like the error handling a lot. A few initial comments

mirgecom/steppers.py
majosm: The diff makes it hard to see, but this section is actually outside of the timestepping loop.

examples/vortex-mpi.py
```diff
@@ -171,22 +173,39 @@ def main(ctx_factory=cl.create_some_context, use_profiling=False, use_logmgr=Fal
...
                               boundaries=boundaries, eos=eos)
 
     def my_checkpoint(step, t, dt, state):
         sim_checkpoint(discr, visualizer, eos, q=state,
-                       exact_soln=initializer, vizname=casename, step=step,
-                       t=t, dt=dt, nstatus=nstatus, nviz=nviz,
+                       exact_soln=initializer, step=step,
+                       t=t, dt=dt, nstatus=nstatus, healthcheck_callback=healthcheck,
+                       nviz=nviz, vis_callback=write_vis,
                        exittol=exittol, constant_cfl=constant_cfl)
```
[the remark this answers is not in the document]
majosm: Yeah I was wondering about that. Could pass a list of "actions" with a frequency associated with each, or something like that. Might be best to save that for #257 though.

MTCam:
This probably won't come as surprise given my objections to other developments along these lines.. but imo this is becoming far too complicated.

imo, the stepper should not be in the business of handling exceptions, like ever.  Callbacks are *collective* - once the callback is called, all control has passed back the driver.  The user has an *optional* chance to evaluate the state here *however he chooses*, and to exit the simulation if he wants.  The control never needs to be handed back to the stepper if the plan is to exit, but in any case tossing an exception for the stepper to handle is just very backwards to me. 

There is a lot of effort going into this exception tossing and handling by the various components, all of which (imo) are becoming too complicated by these changes that do not seem to support user need.

majosm:
> imo, the stepper should not be in the business of handling exceptions, like ever.

If we want to get back the partially-stepped results after an error occurs, the stepper and error handling become intertwined one way or another. If we don't handle the exceptions inside the stepper, we have to do like we did before and store the last good stepper state inside the exceptions, which in general is pretty awkward.

> Callbacks are _collective_ - once the callback is called, all control has passed back the driver. The user has an _optional_ chance to evaluate the state here _however he chooses_, and to exit the simulation if he wants. The control never needs to be handed back to the stepper if the plan is to exit, but in any case tossing an exception for the stepper to handle is just very backwards to me.

I'm not sure I follow what you're describing here. Elaborate?

> There is a lot of effort going into this exception tossing and handling by the various components, all of which (imo) are becoming too complicated by these changes that do not seem to support user need.

In what sense is this not supporting user need?

MTCam:
> 
> If we want to get back the partially-stepped results after an error occurs, the stepper and error handling become intertwined one way or another. If we don't handle the exceptions inside the stepper, we have to do like we did before and store the last good stepper state inside the exceptions, which in general is pretty awkward.
> 

I don't follow this.  If the stepper calls the callback with the state, which it does, then the user already has the state.  The user can then check the state, and if the state is in some condition that the requires action then the user can take that action; exit, or decide something else to do.  Full stop.  Nothing else is required.  

> I'm not sure I follow what you're describing here. Elaborate?
> 

If the user wants to exit, then just exit. The user already has full control and all ranks present in the callback.  If the user wants to modify the state, then he does, and control (and the state) go back to the stepper. No exception handling is needed there.

> 
> In what sense is this not supporting user need?

The user needs a mechanism for checking the state during stepping. The callback hooks provide that.  The user *might* need a mechanism for *modifying* the state during stepping - the callback that returns the state provides that.  Imo, that ends the user need section.  

If the user wants to do I/O in the callback, he does it.  If he wants to do healthchecking (by whatever the user determines indicates the health of his simulation), then he does it.  If he wants to exit before stepping is done, just exit.  In my opinion we are trying to automate too much, and assume way too much about what the user needs in their checkpointing or health-checking.

MTCam:
I wanted to say just one more thing that might clarify my perspective on this...

If we provided callbacks *only* and no other infrastructure (i.e. no health checking, no checkpointing, no status messaging etc), then someone like @anderson2981 would use the callback to do the things that he needs to do - he'd add a check on specific volume, say, or maybe mass fractions, he'd check to see if he wants to do a viz dump and then do it.

imo, we should let the users define what it is they need to do in their particular simulation, and these things are so wildly varied that rolling some automated system for handling all that is just awkward, not very flexible, and will almost always be less than or grossly more than what the user needs.

We *could* however provide several utilities that users can use to roll their own checkpoint/callback procedures - and this would be very useful to them.  Those utilities may throw exceptions, or whatever - and I think that is fine - but handling of those (imo) is squarely in the domain of user option.

MTCam:
I wanted to say just one more thing that might clarify my perspective on this...

If we provided callbacks *only* and no other infrastructure (i.e. no health checking, no checkpointing, no status messaging etc), then someone like @anderson2981 would use the callback to do the things that he needs to do - he'd add a check on specific volume, say, or maybe mass fractions, he'd check to see if he wants to do a viz dump and then do it.

imo, we should let the users define what it is they need to do in their particular simulation, and these things are so wildly varied that rolling some automated system for handling all that is just awkward, not very flexible, and will almost always be less than or grossly more than what the user needs.

We *could* however provide several utilities that users can use to roll their own checkpoint/callback procedures - and this would be very useful to them.  Those utilities may throw exceptions, or whatever - and I think that is fine - but handling of those (imo) is squarely in the domain of user option.

majosm:
> I don't follow this. If the stepper calls the callback with the state, which it does, then the user already has the state. The user can then check the state, and if the state is in some condition that the requires action then the user can take that action; exit, or decide something else to do. Full stop. Nothing else is required.

To elaborate a little on what I was trying to say previously: I was contrasting the approach laid out here vs. the current approach in main.

Option 1 (here): Handle exception in the stepper via callback, user decides whether to continue or stop.
Option 2 (main): Handle exception in the driver, exception contains the last valid stepper state.

Both involve intertwining the stepping with error handling to some degree.

> If the user wants to exit, then just exit. The user already has full control and all ranks present in the callback. If the user wants to modify the state, then he does, and control (and the state) go back to the stepper. No exception handling is needed there.

Hang on -- which callback are we talking about here? The checkpoint? If I'm understanding correctly and you're saying that we should just handle errors that occur in the checkpoint immediately as they come up; I guess my only dislike about this is that the user would have to add separate error handling for errors that occur in other parts of the stepping (I'm thinking timestep size computation, RHS, maybe elsewhere if we have more sophisticated timestepping through Leap?).

> I wanted to say just one more thing that might clarify my perspective on this...
> 
> If we provided callbacks _only_ and no other infrastructure (i.e. no health checking, no checkpointing, no status messaging etc), then someone like @anderson2981 would use the callback to do the things that he needs to do - he'd add a check on specific volume, say, or maybe mass fractions, he'd check to see if he wants to do a viz dump and then do it.
> 
> imo, we should let the users define what it is they need to do in their particular simulation, and these things are so wildly varied that rolling some automated system for handling all that is just awkward, not very flexible, and will almost always be less than or grossly more than what the user needs.
> 
> We _could_ however provide several utilities that users can use to roll their own checkpoint/callback procedures - and this would be very useful to them. Those utilities may throw exceptions, or whatever - and I think that is fine - but handling of those (imo) is squarely in the domain of user option.

I don't think this precludes any of that. `sim_checkpoint` here provides a callback for the user to define whatever health checks they want. `sim_healthcheck` (or `fluid_healthcheck`, whatever we end up calling it) is just a helper routine that the user can choose to call. Whether `sim_checkpoint` should exist or if the user should just call `sim_healthcheck` directly from `my_checkpoint` is an open question, but not one to be answered in this/Thomas' PR IMO.

MTCam:
> To elaborate a little on what I was trying to say previously: I was contrasting the approach laid out here vs. the current approach in main.
> 
> Option 1 (here): Handle exception in the stepper via callback, user decides whether to continue or stop.
> Option 2 (main): Handle exception in the driver, exception contains the last valid stepper state.
> 
> Both involve intertwining the stepping with error handling to some degree.
> 

Currently in main, there is no exception thrown.  The stepper calls something called `checkpoint`, instead of a more general callback (different in name only). `checkpoint` does not toss, so Option 2 is not like main, but main is a *lot* more like what I am advocating for here.  I think we could replace the name `checkpoint` with `prestep_callback` (or whatever) and be done.  If the user wants to exit the simulation inside the callback, then they are free to do so.  No exceptions are needed in that case.  I am having a lot of trouble figuring out why all of this extra machinery is needed.

Anyway.. this discussion is getting too long to respond to piece-by-piece.  It would be terrific if we could meet to talk about this stuff.

MTCam:
> > If the user wants to exit, then just exit. The user already has full control and all ranks present in the callback. If the user wants to modify the state, then he does, and control (and the state) go back to the stepper. No exception handling is needed there.
> 
> Hang on -- which callback are we talking about here? The checkpoint? If I'm understanding correctly and you're saying that we should just handle errors that occur in the checkpoint immediately as they come up; I guess my only dislike about this is that the user would have to add separate error handling for errors that occur in other parts of the stepping (I'm thinking timestep size computation, RHS, maybe elsewhere if we have more sophisticated timestepping through Leap?).
> 

I'm talking about the pre-step callback called from the stepper.  That's really the only callback that we currently use, and the only one I can really think of that we need.  We already wrap the call to the stepper in a try/catch, so we already catch anything thrown by anything that the stepper calls.


> > We _could_ however provide several utilities that users can use to roll their own checkpoint/callback procedures - and this would be very useful to them. Those utilities may throw exceptions, or whatever - and I think that is fine - but handling of those (imo) is squarely in the domain of user option.
> 
> I don't think this precludes any of that. `sim_checkpoint` here provides a callback for the user to define whatever health checks they want. 

I think that `sim_checkpoint` already precludes it.  `sim_checkpoint` is just a thing that I myself came up with to reduce the amount of duplications in the examples. It is not a needed thing - and imo already does way too much, and is way too complicated.

> `sim_healthcheck` (or `fluid_healthcheck`, whatever we end up calling it) is just a helper routine that the user can choose to call. Whether `sim_checkpoint` should exist or if the user should just call `sim_healthcheck` directly from `my_checkpoint` is an open question, but not one to be answered in this/Thomas' PR IMO.

`sim_healthcheck` is already doing far far too much, too complicated, does stuff the user doesn't want, etc.  Here's an example of what I'm advocating for here:

```python
def my_callback(state, istep, time, dt, whatever):
	if check_step(istep, time, vizrfreq):
       do_viz_dump(...)
    if check_step(istep, time, statusfreq):
       do_my_status(...)
    spec_vol = 1 / state.mass
    if discr.nodal_max(spec_vol) > spec_vol_limit:
       do_viz_dump(...)
       log.info("Quitting due to specific volume violation.")
       exit
```

This is OK, the callback is already collective.  `nodal_max` is collective.  No exceptions needed, no monolithic health-checker or checkpointing crap, just exit. Then `my_callback` is passed to the stepper that calls this just before every step.

majosm:
> Currently in main, there is no exception thrown.

I'm talking about [this](https://github.com/illinois-ceesd/mirgecom/blob/fe13237772e2759d9303c879b5fb00c639147858/mirgecom/simutil.py#L162) and [this](https://github.com/illinois-ceesd/mirgecom/blob/fe13237772e2759d9303c879b5fb00c639147858/examples/vortex-mpi.py#L193-L196).

> The stepper calls something called `checkpoint`, instead of a more general callback (different in name only). `checkpoint` does not toss, so Option 2 is not like main, but main is a _lot_ more like what I am advocating for here. I think we could replace the name `checkpoint` with `prestep_callback` (or whatever) and be done. If the user wants to exit the simulation inside the callback, then they are free to do so. No exceptions are needed in that case. I am having a lot of trouble figuring out why all of this extra machinery is needed.

As I said, this is ok if you only care about errors that occur inside the checkpoint. If errors can occur elsewhere (timstep computation, RHS, timestepper itself?) then this doesn't help.

> We already wrap the call to the stepper in a try/catch, so we already catch anything thrown by anything that the stepper calls.

This requires the user to add separate error handling for checkpointing errors vs. other errors. I don't see why we necessarily need to treat them separately.

> > I don't think this precludes any of that. `sim_checkpoint` here provides a callback for the user to define whatever health checks they want.
> 
> I think that `sim_checkpoint` already precludes it. `sim_checkpoint` is just a thing that I myself came up with to reduce the amount of duplications in the examples. It is not a needed thing - and imo already does way too much, and is way too complicated.
> 
> > `sim_healthcheck` (or `fluid_healthcheck`, whatever we end up calling it) is just a helper routine that the user can choose to call. Whether `sim_checkpoint` should exist or if the user should just call `sim_healthcheck` directly from `my_checkpoint` is an open question, but not one to be answered in this/Thomas' PR IMO.
> 
> `sim_healthcheck` is already doing far far too much, too complicated, does stuff the user doesn't want, etc.

I repeat: whether `sim_checkpoint` should exist or not is orthogonal to this PR. I'm trying to make the minimum amount of changes necessary to it here in order to make it support health checking. The rest can go in #257. The error handling here is independent of this.

> Here's an example of what I'm advocating for here:
> 
> ```python
> def my_callback(state, istep, time, dt, whatever):
> 	if check_step(istep, time, vizrfreq):
>        do_viz_dump(...)
>     if check_step(istep, time, statusfreq):
>        do_my_status(...)
>     spec_vol = 1 / state.mass
>     if discr.nodal_max(spec_vol) > spec_vol_limit:
>        do_viz_dump(...)
>        log.info("Quitting due to specific volume violation.")
>        exit
> ```
> 
> This is OK, the callback is already collective. `nodal_max` is collective. No exceptions needed, no monolithic health-checker or checkpointing crap, just exit. Then `my_callback` is passed to the stepper that calls this just before every step.

Exceptions are the standard way to handle errors in Python. I don't see a compelling reason to avoid them. There really isn't much of a practical difference between:

```python
log.info("Quitting due to specific volume violation.")
exit
```
and
```python
raise RuntimeError("Specific volume violation.")
```
In principle doing the above after the `nodal_max` is fine, and strictly speaking is all we would need to do. The main reasons for the additional machinery entering the picture like `bcast_from_lowest_rank`, `exception_callback`, etc. are: 1) to avoid doing multiple global nodal reductions, 2) to avoid duplicating the error handling code.

MTCam:
> > Currently in main, there is no exception thrown.
> 
> I'm talking about [this](https://github.com/illinois-ceesd/mirgecom/blob/fe13237772e2759d9303c879b5fb00c639147858/mirgecom/simutil.py#L162) and [this](https://github.com/illinois-ceesd/mirgecom/blob/fe13237772e2759d9303c879b5fb00c639147858/examples/vortex-mpi.py#L193-L196).]

Both of these come from an exception tossed by checkpoint routine - which should be eliminated.  I understand you don't want to do it here, or in TG's PR, but my argument is that we shouldn't spend time coding around something that shouldn't even exist.

> 
> > The stepper calls something called `checkpoint`, instead of a more general callback (different in name only). `checkpoint` does not toss, so Option 2 is not like main, but main is a _lot_ more like what I am advocating for here. I think we could replace the name `checkpoint` with `prestep_callback` (or whatever) and be done. If the user wants to exit the simulation inside the callback, then they are free to do so. No exceptions are needed in that case. I am having a lot of trouble figuring out why all of this extra machinery is needed.
> 
> As I said, this is ok if you only care about errors that occur inside the checkpoint. If errors can occur elsewhere (timstep computation, RHS, timestepper itself?) then this doesn't help.

Why is the try/catch around stepping not sufficient to catch any exceptions tossed by anything inside stepping (or that stepping calls).  

I would argue that no exceptions should ever be thrown by any library component unless it is thrown collectively.  This is a good strategy for parallel codes. Library routines should behave the same regardless of domain-decomp.  If one rank tosses in some library routine, then they all should. If we adopt this policy, many of our issues surrounding exception handling can be mitigated.

> 
> > We already wrap the call to the stepper in a try/catch, so we already catch anything thrown by anything that the stepper calls.
> 
> This requires the user to add separate error handling for checkpointing errors vs. other errors. I don't see why we necessarily need to treat them separately.

I disagree. The user often needs to know which components fail.  And if the user does not want to call all the stuff done by (the current checkpoint, or healthcheck) then he doesn't have to.  

   
> > This is OK, the callback is already collective. `nodal_max` is collective. No exceptions needed, no monolithic health-checker or checkpointing crap, just exit. Then `my_callback` is passed to the stepper that calls this just before every step.
> 
> Exceptions are the standard way to handle errors in Python. I don't see a compelling reason to avoid them. There really isn't much of a practical difference between:

I'm not advocating for avoiding all exceptions, I'm only pointing out that no exceptions are needed to do the things we need the simulation to do in this callback example.  No callback should ever need to throw an exception to the stepper - it just is not needed. 

My main point is that all of this work being done to encapsulate health checking and checkpointing, and handling all their errors automatically by the library components is unnecessary and does not address a simulation need; so why do it?

> 
> ```python
> log.info("Quitting due to specific volume violation.")
> exit
> ```
> 
> and
> 
> ```python
> raise RuntimeError("Specific volume violation.")
> ```
> 

I'm fine with this. I don't care how the user exits or causes the simulation to exit, only that doing so is in the hands of the user, not the library. 

> In principle doing the above after the `nodal_max` is fine, and strictly speaking is all we would need to do. The main reasons for the additional machinery entering the picture like `bcast_from_lowest_rank`, `exception_callback`, etc. are: 1) to avoid doing multiple global nodal reductions, 2) to avoid duplicating the error handling code.

I still don't buy it.  The only reason this additional machinery is needed from my perspective is in the processing of errors tossed by routines that should not exist to begin with.  I'm definitely just fine with the prospect of seeing a lot of code duplication in the drivers.  Drivers are the user's business, and almost everything should be done in the driver, including error handling.

majosm:
I guess we're moving over to #381, but just to address a few lingering things here...

> Why is the try/catch around stepping not sufficient to catch any exceptions tossed by anything inside stepping (or that stepping calls).

A try/catch around `advance_state` doesn't allow you to retrieve the partially-stepped state, unless you shoehorn the stepper state into the exception object, which seems awkward. (Not all exceptions can or should be aware that they're part of a time marching loop.)
 
> I would argue that no exceptions should ever be thrown by any library component unless it is thrown collectively. This is a good strategy for parallel codes. Library routines should behave the same regardless of domain-decomp. If one rank tosses in some library routine, then they all should. If we adopt this policy, many of our issues surrounding exception handling can be mitigated.

This isn't really feasible. Any subpackage can raise an exception at any time, and we don't have control over this.

> I disagree. The user often needs to know which components fail. And if the user does not want to call all the stuff done by (the current checkpoint, or healthcheck) then he doesn't have to.

Note that `sim_checkpoint` here takes a healthcheck callback that the user defines. `sim_healthcheck` is/was meant to be a default, or at least a placeholder until we figured out what to do proper in #257.

> I'm fine with this. I don't care how the user exits or causes the simulation to exit, only that doing so is in the hands of the user, not the library.

It is though? The user is choosing to call `sim_healthcheck` here. It's not required.

> I still don't buy it. The only reason this additional machinery is needed from my perspective is in the processing of errors tossed by routines that should not exist to begin with.

All of this extra machinery can be applied to #381, too, and IMO would still be nice to have.

> I'm definitely just fine with the prospect of seeing a lot of code duplication in the drivers.

🤷‍♂️ I don't really share this view.

MTCam:
> I guess we're moving over to #381, but just to address a few lingering things here...
> 
> > Why is the try/catch around stepping not sufficient to catch any exceptions tossed by anything inside stepping (or that stepping calls).
> 
> A try/catch around `advance_state` doesn't allow you to retrieve the partially-stepped state, unless you shoehorn the stepper state into the exception object, which seems awkward. (Not all exceptions can or should be aware that they're part of a time marching loop.)
> 

I didn't recognize the subtle points about the state before you explained this. (at least it's subtle to me). This is a fair point.

As for getting the state back if a sub-package fails or throws an exception:  I think we don't care much about that case. The main failure modes are:

- dt gets a bad value: stepping returns "early", this is handled ok.
- Invalid (or divergent) values in the state (health issue), this is now handled ok.
- Error in subpackage:
    - Most of these occur before the first step, during the startup or compilation, they are rare after things get rolling
    - If this does occur, it kills the simulation without our getting a chance to save the state.

> > I would argue that no exceptions should ever be thrown by any library component unless it is thrown collectively. This is a good strategy for parallel codes. Library routines should behave the same regardless of domain-decomp. If one rank tosses in some library routine, then they all should. If we adopt this policy, many of our issues surrounding exception handling can be mitigated.
> 
> This isn't really feasible. Any subpackage can raise an exception at any time, and we don't have control over this.

I totally agree and recognize this point.  I only mean MIRGE-Com library.  We just have to accept it if a subpackage fails and kills the sim.  This is a rare failure mode beyond initialization and first step.

> 
> > I disagree. The user often needs to know which components fail. And if the user does not want to call all the stuff done by (the current checkpoint, or healthcheck) then he doesn't have to.
> 
> Note that `sim_checkpoint` here takes a healthcheck callback that the user defines. `sim_healthcheck` is/was meant to be a default, or at least a placeholder until we figured out what to do proper in #257.
> 

It turns out that the structure where we use callbacks into the ( i would argue overly-restrictive ) utility drivers (like `sim_checkpoint`) is quite awkward for the user to use to accomplish what they want.  The advantage of reducing code in the driver I think is not worth it due to the impact on utility, and user convenience.

I totally agree that many simple situations can be encapsulated inside some ready-to-use constructs, and these constructs can be arbitrarily complex, accepting callbacks, etc. They should not be required or necessary, however.  I think it is important for us to hang all this out in the driver (for now) for the user (and us), and then have a think about how we can encapsulate these functionalities in a way we can all get on-board with.

The prospect of more callbacks and the circular error handling issues surrounding the retention of stuff like `sim_checkpoint` to me was working against our goals.  I had the feeling (because they told me) that the circuitous internal utility call structure was hiding too much, and not flexible enough. Figuring out how to get your own stuff wedged into the machinery was a slog, resulting in a steeper learning curve for using the code. All I am saying here is that I am glad `sim_checkpoint` is gone, and that we see directly the (pretty minimal at present) stuff that we need in a `checkpoint`-like facility.  

> > I'm definitely just fine with the prospect of seeing a lot of code duplication in the drivers.
> 
> 🤷‍♂️ I don't really share this view.

Let's duke it out in #257?  FWIW, I really like these discussions.

majosm:
> As for getting the state back if a sub-package fails or throws an exception: I think we don't care much about that case. The main failure modes are:
> 
> * dt gets a bad value: stepping returns "early", this is handled ok.
> * Invalid (or divergent) values in the state (health issue), this is now handled ok.
> * Error in subpackage:
>   
>   * Most of these occur before the first step, during the startup or compilation, they are rare after things get rolling
>   * If this does occur, it kills the simulation without our getting a chance to save the state.

> I totally agree and recognize this point. I only mean MIRGE-Com library. We just have to accept it if a subpackage fails and kills the sim. This is a rare failure mode beyond initialization and first step.

I wouldn't say I was thinking so much about subpackages, mainly just other parts of mirgecom. But I'll admit I don't have an entirely clear idea of whether anything outside of timestep computation and checkpointing would ever fail mid-stepping. @inducer Can you think of anything?

> It turns out that the structure where we use callbacks into the ( i would argue overly-restrictive ) utility drivers (like `sim_checkpoint`) is quite awkward for the user to use to accomplish what they want. The advantage of reducing code in the driver I think is not worth it due to the impact on utility, and user convenience.
> 
> I totally agree that many simple situations can be encapsulated inside some ready-to-use constructs, and these constructs can be arbitrarily complex, accepting callbacks, etc. They should not be required or necessary, however. I think it is important for us to hang all this out in the driver (for now) for the user (and us), and then have a think about how we can encapsulate these functionalities in a way we can all get on-board with.
> 
> The prospect of more callbacks and the circular error handling issues surrounding the retention of stuff like `sim_checkpoint` to me was working against our goals. I had the feeling (because they told me) that the circuitous internal utility call structure was hiding too much, and not flexible enough. Figuring out how to get your own stuff wedged into the machinery was a slog, resulting in a steeper learning curve for using the code. All I am saying here is that I am glad `sim_checkpoint` is gone, and that we see directly the (pretty minimal at present) stuff that we need in a `checkpoint`-like facility.

That's fair. A 'reset' could be helpful. Let's continue in #381.

PR #380, opened 2021-06-11
Update and tweak callbacks and health checking

MTCam:
Here is a set of changes that I think accomplishes most the stuff we've been talking about.

- Updates callback changes to work with current main (including CV array containers)
- Eliminates `sim_checkpoint` in favor of fully customizable `my_checkpoint`
- Allows full customization of status messages, viz fields, and health checking
- Implements minimally collective health-check (only one collective comm to sync)
- Optionally dumps viz at health fail (automatically collective)
- Correctly places control over what to do with errors in the hands of the user/driver

Since this also merges current main, the changeset is pretty large.  You can get a good idea of what is done by having a look at the example drivers. #381 shows the diff against main.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -123,6 +123,7 @@ jobs:
             MINIFORGE_INSTALL_DIR=.miniforge3
             . "$MINIFORGE_INSTALL_DIR/bin/activate" testing
 
+            sudo apt-get update
             sudo apt-get install texlive-latex-extra latexmk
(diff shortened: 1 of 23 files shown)


MTCam:
The health check was tweaked to check only pressure since if pressure is bad, temperature will also be bad. Temperature is not used in any of our Euler sims, and even in the NS with Pyro/chemistry, pressure is a function of temperature, and will be nan/inf if any soln var or temperature is nan/inf.

PR #382, opened 2021-06-11
ci: update packages

matthiasdiener:
Should resolve spurious CI failures.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -123,6 +123,7 @@ jobs:
             MINIFORGE_INSTALL_DIR=.miniforge3
             . "$MINIFORGE_INSTALL_DIR/bin/activate" testing
 
+            sudo apt-get update
             sudo apt-get install texlive-latex-extra latexmk
             conda install sphinx graphviz
             pip install sphinx-math-dollar sphinx-copybutton furo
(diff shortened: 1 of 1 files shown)


MTCam (approved):
👍 Thanks Matthias!

.github/workflows/ci.yaml
```diff
@@ -141,7 +142,7 @@ jobs:
         - uses: actions/checkout@v2
         - name: Install emirge
           run: |
-            [[ $(uname) == Linux ]] && sudo apt-get install -y openmpi-bin libopenmpi-dev
+            [[ $(uname) == Linux ]] && sudo apt-get update && sudo apt-get install -y openmpi-bin libopenmpi-dev
```
inducer, line 145: No objection here, though maybe there's a benefit to grabbing this from conda-forge?

matthiasdiener: But this is the emirge case - how would we get emirge to use the conda MPI (from a different env)?

inducer: Fair, I hadn't picked up on that. I was meaning that more for the other CIs, so feel free to disregard this.

MTCam:
Merge: ftw.

PR #381, opened 2021-06-11
Recallbacks

MTCam:
Continued from PR (#370):

This builds on @thomasgibson callbacks and uses the opportunity to clean up some cumbersome junk in the infrastructure that was gumming up the use of the code in production. Instead of building the callback and health checking facilities on old junky code that we didn't like, we refactored it a bit here, to give the user more flexibility, and responsibility, for the behavior of the simulation application.

- Updates the old callback (`checkpoint`) to the explicitly named (`pre_step_callback`)
- Takes apart the clunky `sim_checkpoint`, `sim_vizualization`, for a more direct, much less "fluid-specific" interface
- Allows full customization of status messages, viz fields, and health checking
- Implements minimally collective health-check (only one collective comm to sync)
- Optionally dumps viz at health fail (automatically collective)
- Correctly places control over what to do with errors in the hands of the user/driver

PR #384, opened 2021-06-11
Clean up healthchecking in #381

majosm:
(Only applied to autoignition-mpi for the moment.)

### DIFF
--- a/examples/autoignition-mpi.py
+++ b/examples/autoignition-mpi.py
@@ -243,28 +243,24 @@ def main(ctx_factory=cl.create_some_context, use_leap=False):
             if rank == 0:
                 logger.info(status_msg)
 
-        errored = False
+        exception = None
         if do_health:
             from mirgecom.simutil import check_naninf_local, check_range_local
-            if check_naninf_local(discr, "vol", dv.pressure) \
-               or check_range_local(discr, "vol", dv.pressure):
-                errored = True
-                message = "Invalid pressure data found.\n"
-            comm = discr.mpi_communicator
-            if comm is not None:
-                errored = comm.allreduce(errored, op=MPI.LOR)
-            if errored:
-                if rank == 0:
-                    logger.info("Fluid solution failed health check.")
-                logger.info(message)   # do this on all ranks
+            from mirgecom.exceptions import SimulationHealthError
+            if check_naninf_local(discr, "vol", dv.pressure):
+                exception = SimulationHealthError("NaNs detected in pressure.")
+            elif check_range_local(discr, "vol", dv.pressure):
+                exception = SimulationHealthError("Pressure out of expected range.")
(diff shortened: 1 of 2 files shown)


MTCam (commented):
I love this in terms of what it does to the code, but are you concerned with doing two collectives when one reduce could have done the trick?  Or is it that you want the actual exception with its message on all the ranks? Would it be useful to try to dump the message for all ranks, so that we can get them all? There may be disparate exceptions among ranks, if we care.

inducer:
I think this is a good direction, too... quite possibly, we can have a bunch of rank-local health checks and then keep the communication logic entirely separate. (maybe even in `advance_state`?

majosm:
> I love this in terms of what it does to the code, but are you concerned with doing two collectives when one reduce could have done the trick? Or is it that you want the actual exception with its message on all the ranks? Would it be useful to try to dump the message for all ranks, so that we can get them all? There may be disparate exceptions among ranks, if we care.

Synchronizing just seems like the least surprising behavior IMO. Different ranks raising different exceptions could get nasty since they may not be handled the same way. It's probably best to choose one consistently.

The extra broadcast is only incurred if there actually is an exception, so I don't think it should present a performance issue.

MTCam:
> > I love this in terms of what it does to the code, but are you concerned with doing two collectives when one reduce could have done the trick? Or is it that you want the actual exception with its message on all the ranks? Would it be useful to try to dump the message for all ranks, so that we can get them all? There may be disparate exceptions among ranks, if we care.
> 
> Synchronizing just seems like the least surprising behavior IMO. Different ranks raising different exceptions could get nasty since they may not be handled the same way. It's probably best to choose one consistently.
> 

I agree completely about synchronizing, but I am still not convinced the exceptions themselves should be exchanged or synchronized. I think the specificity of the exception is too high.

I think we should synchronize that a fatal error occurred, but let each rank log its own error, and then toss the same exception on all ranks.  Even if it is just ("Some rank had a fatal sim error").  If we let each rank report its errors then we get more information about what went wrong, and it eliminates the need to broadcast or exchange the exceptions as a whole, reducing the complexity of this error handling business.

majosm:
> I agree completely about synchronizing, but I am still not convinced the exceptions themselves should be exchanged or synchronized. I think the specificity of the exception is too high.
> 
> I think we should synchronize that a fatal error occurred, but let each rank log its own error, and then toss the same exception on all ranks. Even if it is just ("Some rank had a fatal sim error"). If we let each rank report its errors then we get more information about what went wrong, and it eliminates the need to broadcast or exchange the exceptions as a whole, reducing the complexity of this error handling business.

I guess I'm ok with this. Generally I think exception specificity is a good thing, since it allows the caller to pick and choose what errors to handle and how. But since these errors aren't likely to be handled except by aborting, it probably doesn't matter that much.

PR #385, opened 2021-06-13
Array docs

inducer:


MTCam:
Sorry @inducer, just now seeing this!  Thanks for kicking this off!  Would it be useful to provide a couple real code examples from *MIRGE-Com* inside the doc section?  Or is that just a distraction?  I can add that if you think it is useful.

MTCam (commented):
I like this, and certainly it could be pushed right away as a vast improvement to the docs.  The thing it doesn't (directly) do that I was hoping for is improve user understanding of when/how to use `freeze/thaw` in *MIRGE-Com*.

A few practical examples *might* help, but I'm not sure that even that will do it.  How are the users to know what quantities are frozen to begin with?

inducer:
I think examples would be immensely helpful. They should be doctests so that we can catch when they fall out of date. I'll add an example example. ;)

> How are the users to know what quantities are frozen to begin with?

https://documen.tician.de/arraycontext/array_context.html#usage-guidelines provides some hints:

* If you didn't provide an array context (explicitly or implicitly via an input array), you're going to get frozen data.
* Any data that is cached/long-lived/"at rest' is going to be frozen.

I think it'd be useful to spell this out.

inducer:
Added a demo doctest in a37b678.

doc/support/arrays.rst
```diff
@@ -0,0 +1,92 @@
...
+a fair bit of functionality predates these concepts, we are still catching up
+in terms of updating the documentation. (Help welcome!)
+
+In the meantime, these rules of thumb should cover most cases:
+
+* If you did not provide an array context to the function
+  (explicitly or implicitly via an input array), you will receive frozen data.
+* Any data that is cached/long-lived/"at rest" is going to be frozen.
+
+To demonstrate the effect of this, first we need som setup:
```
inducer, line 39:
```suggestion
To demonstrate the effect of this, first we need some setup:
```

inducer:
(@MTCam, please do add further explanation as you see fit.)

MTCam:
Nice! Loving this doctest thing.

doc/support/arrays.rst
```diff
@@ -0,0 +1,106 @@
...
+  (explicitly or implicitly via an input array), you will receive frozen data.
+* Any data that is cached/long-lived/"at rest" is going to be frozen.
+* If the data is the result of a `memoized` function, then it will likely
+  be frozen.
+
+To demonstrate the effect of this, first we need some setup:
+
+.. doctest::
+
+   >>> # setup
```
inducer, line 45:
```suggestion
```

doc/support/arrays.rst
```diff
@@ -0,0 +1,106 @@
...
+Most quantities that are maintained by the discretization will be frozen. For example,
+if one wanted to grab the nodes of the mesh or normals of a named surface in the mesh:
+
+.. doctest::
+
+   >>> from grudge.dof_desc import DTAG_BOUNDARY
+   >>> nodes = thaw(dcoll.nodes(), actx)
+   >>> dd = DTAG_BOUNDARY("all_faces")
+   >>> nhat = thaw(dcoll.normal(dd), actx)
+
```
inducer, line 65:
```suggestio
```

inducer:
@MTCam Thanks! Approved from your end?

MTCam (approved):
Thanks for doing this! It is tremendously helpful.

PR #388, opened 2021-06-16
User friendlier restart

MTCam:
This adds a restart to a relevant example, similar to how it is used in production, and adds a couple of utilities to make restarting more convenient and user-friendly.  Note that this PR is against the restart capability in #188 .

### DIFF
--- a/examples/autoignition-mpi.py
+++ b/examples/autoignition-mpi.py
@@ -111,7 +111,7 @@ def main(ctx_factory=cl.create_some_context, casename="autoignition", use_leap=F
             + restart_file_pattern.format(casename=restart_name,
                                           step=restart_step, rank=rank)
         )
-        from mirgecom.simutil import read_restart_data
+        from mirgecom.restart import read_restart_data
         restart_data = read_restart_data(rst_filename)
         local_mesh = restart_data["local_mesh"]
         local_nelements = local_mesh.nelements
@@ -203,8 +203,8 @@ def main(ctx_factory=cl.create_some_context, casename="autoignition", use_leap=F
         current_t = restart_data["t"]
         current_step = restart_step
         restart_discr = discr.discr_from_dd("vol")
-        from mirgecom.simutil import make_fluid_restart_state
-        current_state = make_fluid_restart_state(actx, restart_discr,
+        from mirgecom.restart import make_fluid_state
+        current_state = make_fluid_state(actx, restart_discr,
                                                  restart_data["state"])
     else:
         # Set the current state from time 0
@@ -267,7 +267,7 @@ def main(ctx_factory=cl.create_some_context, casename="autoignition", use_leap=F
                 "global_nelements": global_nelements,
                 "num_parts": nproc
             }
-            from mirgecom.simutil import write_restart_file
+            from mirgecom.restart import write_restart_file
             write_restart_file(actx, rst_data, rst_filename, comm)
 
         # awful - computes potentially expensive viz quantities
--- a/examples/autoignition-mpi.py
+++ b/examples/autoignition-mpi.py
@@ -59,7 +59,8 @@ logger = logging.getLogger(__name__)
 
 
 @mpi_entry_point
-def main(ctx_factory=cl.create_some_context, use_leap=False):
+def main(ctx_factory=cl.create_some_context, casename="autoignition", use_leap=False,
+         restart_step=None, restart_name=None):
     """Drive example."""
     cl_ctx = ctx_factory()
     queue = cl.CommandQueue(cl_ctx)
@@ -73,7 +74,7 @@ def main(ctx_factory=cl.create_some_context, use_leap=False):
     # This example runs only 3 steps by default (to keep CI ~short)
     # With the mixture defined below, equilibrium is achieved at ~40ms
     # To run to equlibrium, set t_final >= 40ms.
-    t_final = 3e-9
+    t_final = 1e-8
     current_cfl = 1.0
     velocity = np.zeros(shape=(dim,))
     current_dt = 1e-9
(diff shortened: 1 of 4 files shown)


MTCam:
I just realized I was using `make_conserved` changes from #298 for this, but we'll get it fixed up.  I mostly want us to converge on an interface.

examples/autoignition-mpi.py
```diff
@@ -97,12 +99,31 @@ def main(ctx_factory=cl.create_some_context, use_leap=False):
...
-
-    from meshmode.mesh.generation import generate_regular_rect_mesh
-    generate_mesh = partial(generate_regular_rect_mesh, a=(box_ll,) * dim,
-                            b=(box_ur,) * dim, nelements_per_axis=(nel_1d,) * dim)
-    local_mesh, global_nelements = generate_and_distribute_mesh(comm, generate_mesh)
-    local_nelements = local_mesh.nelements
+    nproc = comm.Get_size()
+
+    restart_file_pattern = "{casename}-{step:04d}-{rank:04d}.pkl"
+    restart_path = "restart_data/"
```
MTCam, line 105: The user (@anderson2981) wants this interface because he may start/restart from datasets in different locations.

mirgecom/restart.py
```diff
@@ -0,0 +1,68 @@
...
+def write_restart_file(actx, restart_data, filename, comm=None):
+    """Pickle the simulation data into a file for use in restarting."""
+    rank = 0
+    if comm:
+        rank = comm.Get_rank()
+    if rank == 0:
+        import os
+        rst_dir = os.path.dirname(filename)
+        if not os.path.exists(rst_dir):
+            os.mkdir(rst_dir)
```
inducer, line 61: `makedirs`? https://docs.python.org/3/library/os.html#os.makedirs

mirgecom/restart.py
```diff
@@ -0,0 +1,68 @@
...
+    comm.barrier()
+    from pytools.obj_array import obj_array_vectorize
+    state = restart_data["state"].join()
+    restart_data["state"] = obj_array_vectorize(actx.to_numpy,
+                                                      flatten(state))
```
inducer, line 66: IIRC the flattening/unflattening shouldn't be necessary, cf. #188.

mirgecom/restart.py
```diff
@@ -0,0 +1,68 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+"""
+
+import pickle
+from meshmode.dof_array import unflatten, flatten
+from mirgecom.fluid import make_conserved
+
+
+def read_restart_data(filename):
```
inducer, line 38: Maybe give this an `actx` so that you can use `with array_context_for_pickling(actx):` in here?

mirgecom/restart.py
```diff
@@ -0,0 +1,68 @@
...
+    """Pickle the simulation data into a file for use in restarting."""
+    rank = 0
+    if comm:
+        rank = comm.Get_rank()
+    if rank == 0:
+        import os
+        rst_dir = os.path.dirname(filename)
+        if not os.path.exists(rst_dir):
+            os.mkdir(rst_dir)
+    comm.barrier()
```
inducer, line 62: Doesn't handle `comm is None`?

mirgecom/restart.py
```diff
@@ -0,0 +1,68 @@
...
+    with open(filename, "rb") as f:
+        restart_data = pickle.load(f)
+    return restart_data
+
+
+def make_fluid_state(actx, discr, restart_q):
+    """Make a :class:`~mirgecom.fluid.ConservedVars` from pickled restart data."""
+    from pytools.obj_array import obj_array_vectorize
+    q = unflatten(actx, discr, obj_array_vectorize(actx.from_numpy, restart_q))
+    return make_conserved(discr.dim, scalar_quantities=q)
```
inducer, line 49: As below: Not sure this is needed. Using [this context manager](https://documen.tician.de/meshmode/array.html?highlight=array_context_for#meshmode.dof_array.array_context_for_pickling) (https://github.com/inducer/meshmode/pull/229 has some doc improvements), `ConservedVars` itself should be picklable, cf. #188.

inducer:

mirgecom/restart.py
MTCam: 🤦

mirgecom/restart.py
```diff
@@ -0,0 +1,59 @@
...
+    rank = 0
+    if comm:
+        rank = comm.Get_rank()
+    if rank == 0:
+        import os
+        rst_dir = os.path.dirname(filename)
+        if not os.path.exists(rst_dir):
+            os.makedirs(rst_dir)
+    if comm:
+        comm.barrier()
```
MTCam, line 56: I'm tempted to say this should be the responsibility of the user setting up the driver/run. I like it when stuff just does one thing. (these are not the `sim_checkpoint` you are looking for :hand:).  On the other hand, it is a luxury interface item...  it's doing no harm here. I dunno.

MTCam:
It's a *lot* thinner after switching to `array_context_for_pickling`, but I think it is still nice to hide some of the guts, and centralize into a module. I like how it is now not fluid-specific at all.

mirgecom/restart.py
```diff
@@ -0,0 +1,59 @@
...
+import pickle
+from meshmode.dof_array import array_context_for_pickling
+
+
+def read_restart_data(actx, filename):
+    """Read the raw restart data dictionary from the given pickle restart file."""
+    with array_context_for_pickling(actx):
+        with open(filename, "rb") as f:
+            restart_data = pickle.load(f)
+    return restart_data
```
inducer, line 42:
```suggestion
            return pickle.load(f)
```

mirgecom/restart.py
inducer: :shrug: I think it's alright.

inducer (approved):
LGTM

PR #390, opened 2021-06-16
generalize state logging and add cfl logging

matthiasdiener:


mirgecom/logging_quantities.py
```diff
@@ -286,6 +291,25 @@ def __call__(self):
...
+        self.discr = discr
+        self.eos = eos
+
+    def __call__(self) -> float:
+        """Return the value of cfl."""
+        if self.state is None:
+            return None
+
+        from mirgecom.inviscid import get_inviscid_cfl
+        return get_inviscid_cfl(self.discr, eos=self.eos, dt=self.dt, cv=self.state)
```
MTCam, line 310: What if `get_inviscid_cfl` is expensive, and we want not only to log it, but also have the results of the computation?  

That is... if I want the results of `get_inviscid_cfl` outside of logging, then I need to call it yet again for my own purposes.  Is there any way the interface could just update the "result" of this function on the outside and then pass it in?

matthiasdiener: Could we cache the result of `get_inviscid_cfl` somewhere inside that function? In that case, both logging and non-logging could benefit.

MTCam: That's a great idea, and you are definitely right about CFL, Temperature, Pressure, etc. 

The main issue I have is that the general case does not seem to be supported; we will need a way for users to compute things of their own device and add that to logging.  (Regardless of whether those things depend on the fluid state or not).  CFL is both a good and a bad example.

The only reason to LOG CFL is because it changes. It only changes when the user is running in *constant DT* mode.  In that mode, CFL is a nodal field, and the user is interested in *occasionally* seeing the min and max value of it in the domain. It is not *used* ever in the simulation - the only reason it is ever even computed is to report the values to the user. Outside of that, it should not even be computed.

So, in constant DT mode, when the user is ready to see CFL, he computes the CFL nodal field and adds that to the visualization dumps, then he takes the min and max of it across the domain and reports that to the stdio for the user's simulation monitoring.

In this way the CFL field is a good example of the sorts of things users will do.  It is a field that the user wants to compute and use in their own code, and they want to log the min/max values of it (domain-wide), and they don't want to compute it all the time, only when they have to in order to make a diagnostic, vizfile, or whatever.

But it is a bad example in that it is a very typical quantity to compute and use, encouraging a treat-it-special kind of approach (e.g. memoize it, treat it like a permanent fixture, etc).   In general, the users will compute all sorts of stuff that is not the CFL, but that is similar in the nature above where it is computed, but not all the time, etc. 
If we intend to use this logging is the sole facility for reporting computed quantities to the user, then we should have a way for users to add those quantities to logging, and review them at some interval.

MTCam:
Beware about the CFL and DT functions that are used here are the ones that are in main, and they are broken.  Real, meaningful, CFL/DT calculations are added in #368.

anderson2981:
I like this, I like #391 even better, which builds on it. I propose closing this and working to merge #391.

PR #395, opened 2021-06-22
Memoize DV on CV

MTCam:
Cache the results of the following to CV:

These had a significant impact on performance of eager mode mixtures:
- Pyrometheus pressure and temperature - significant performance improvement
- Speed of sound for Pyrometheus EOS only

These were removed because they had very little benefit:
- Wavespeeds (|v| + c) - no apparent impact
- Velocity = (cv.momentum / cv.mass) - no apparent impact
- Pressure and temperature for IdealSingleGas - no apparent impact
- Speed of sound for non-mixture EOS

The performance impacts were evaluated on Lassen/GPU using production drivers.


MTCam:
@thomasgibson you mentioned that we should memoize frozen data, but I had a little trouble making that work right. I tried freeze from both dof_array and array_context but I get similar errors. Could you have a quick look?

mirgecom/eos.py
```diff
@@ -398,9 +408,19 @@ def pressure(self, cv: ConservedVars):
...
-        y = self.species_fractions(cv)
-        return self._pyrometheus_mech.get_pressure(cv.mass, temperature, y)
+        @memoize_in(cv, (PyrometheusMixture.pressure,
+                         type(self._pyrometheus_mech)))
+        def get():
+            temperature = self.temperature(cv)
+            y = self.species_fractions(cv)
+            press = self._pyrometheus_mech.get_pressure(cv.mass, temperature, y)
+            # from meshmode.dof_array import freeze
+            # return freeze(cv.array_context, press)
```
inducer, line 418:
```suggestion
```

mirgecom/eos.py
```diff
@@ -398,9 +408,19 @@ def pressure(self, cv: ConservedVars):
...
+            return press
+
+        # from meshmode.dof_array import thaw
+        # return thaw(cv.array_context, get())
```
inducer, line 422:
```suggestion
```

mirgecom/fluid.py
```diff
@@ -248,6 +257,11 @@ def join(self):
             momentum=self.momentum,
             species_mass=self.species_mass)
 
+    def __reduce__(self):
+        """Return a tuple reproduction of self for pickling."""
+        return(ConservedVars, tuple(getattr(self, f.name)
+                                    for f in fields(ConservedVars)))
```
inducer, line 263:
```suggestion
        return (ConservedVars, tuple(getattr(self, f.name)
                                    for f in fields(ConservedVars)))
```

MTCam (commented):
Remove ineffective memoizations.

MTCam:
This is merge-ready pending review @majosm @thomasgibson @inducer.  :)

MTCam:
> @thomasgibson you mentioned that we should memoize frozen data, but I had a little trouble making that work right. I tried freeze from both dof_array and array_context but I get similar errors. Could you have a quick look?

@inducer and I discussed this together, @thomasgibson and I think the conclusion was that normally memoized data would be frozen - but it is OK in this case to memoize the thawed data because it is very transient - changes from step-to-step and any time CV changes.

mirgecom/eos.py
```diff
@@ -398,9 +396,13 @@ def pressure(self, cv: ConservedVars):
 
             p = (\gamma_{\mathtt{mix}} - 1)e
         """
-        temperature = self.temperature(cv)
-        y = self.species_fractions(cv)
-        return self._pyrometheus_mech.get_pressure(cv.mass, temperature, y)
+        @memoize_in(cv, (PyrometheusMixture.pressure,
+                         type(self._pyrometheus_mech)))
+        def get():
```
thomasgibson, line 401: This is _purely_ personal opinion. But I think it looks nicer if we name all the getters something like `get_pressure()`, `get_temperature()`, etc

thomasgibson (approved):
Thanks! This is very clean. I only have a minor comment

mirgecom/eos.py
MTCam: Please see how this grabs ya (4005f6f)

thomasgibson: Great!

MTCam: Well, your suggestion broke Y1 compatibility. THere, i've said it.

PR #396, opened 2021-06-23
Add quick test of reading CV with restart API

MTCam:
Adds a test that demonstrates reading and writing a restart file with the CV.


MTCam:
This could surely be reproduced much more easily, but here it is @inducer. Thanks so much for your help!

inducer:
Should be fixed by https://github.com/inducer/meshmode/pull/234 (once it goes in). Thanks for the squeaky clean reproducer! That made the job a lot easier.

MTCam:
> Should be fixed by [inducer/meshmode#234](https://github.com/inducer/meshmode/pull/234) (once it goes in). Thanks for the squeaky clean reproducer! That made the job a lot easier.

Thanks for the lightning-quick fix!

test/test_restart.py
```diff
@@ -0,0 +1,86 @@
...
+    mesh = generate_regular_rect_mesh(
+        a=(-0.5,) * dim, b=(0.5,) * dim, nelements_per_axis=(nel_1d,) * dim
+    )
+    order = 3
+    discr = EagerDGDiscretization(actx, mesh, order=order)
+
+    def rand():
+        ary = discr.zeros(actx)
+        for grp_ary in ary:
+            grp_ary.set(np.random.rand(*grp_ary.shape))
```
inducer, line 61: * Should use `actx.from_numpy`.
* Shouldn't update in place.

test/test_restart.py
```diff
@@ -0,0 +1,86 @@
...
+
+    rst_data = {"state": test_state}
+    from mirgecom.restart import write_restart_file
+    write_restart_file(actx, rst_data, rst_filename)
+
+    from mirgecom.restart import read_restart_data
+    restart_data = read_restart_data(actx, rst_filename)
+
+    resid = test_state - restart_data["state"]
+    assert discr.norm(resid.join(), np.inf) < 1e-14
```
inducer, line 86:
```suggestion
    assert discr.norm(resid.join(), np.inf) == 0
```

test/test_restart.py
MTCam: Bah, I couldn't figure this out and it was unneeded so I just removed it.  This does appear elsewhere in some test I copied, however.

inducer: #401.

inducer (approved):

PR #397, opened 2021-06-23
Clean up flux functions in NS using #331

majosm:


MTCam:
Wow, super cool Matt. Thanks!

PR #399, opened 2021-06-23
Extend make conserved to handle joined quantities, sharpen the docs.

MTCam:
Now that CV is the main object in the fluid components of *MIRGE-Com*, we can reshape the CV creation tools to be a bit more convenient and transparent at the call site. 

Users typically make CV from these data:

* Split arrays for scalar quantities (e.g. dof_array for each of mass, energy, momentum, etc)
* Split arrays for vector quantities (e.g. array of dof_array for mass, etc)
* Joined arrays for scalar components of Q[Neq] (i.e. with Q[0]=mass, Q[1]=energy, etc)
* Joined arrays for vector components of flux or gradients

Right now, *MIRGE-Com* handles the first two cases (split quantities) properly with `make_conserved`. This change set extends `make_conserved` to the other two cases.  Currently the other two cases are handled by:

```python
# Joined scalar array q
cv = split_conserved(dim, q)
# Joined flux arrays f
flux = split_conserved(dim, q=f)
```

The proposed change (which is currently used by Navier-Stokes, and y1-production) would replace the above with:

```python
# Joined scalar array q
cv = make_conserved(dim, scalar_quantities=q)
# Joined flux arrays f
flux = make_conserved(dim, vector_quantities=f)
```

Although this does not reduce the amount of user-facing code, I believe it makes it more clear that this returns a CV object, and that it is taking a scalar quantity, or vector quantity as an argument.  Internally, our infrastructure is smart enough to figure it out - so the back-end handling is identical, but as a user and reader of the code, I like this better.

### DIFF
--- a/mirgecom/fluid.py
+++ b/mirgecom/fluid.py
@@ -62,9 +62,9 @@ class ConservedVars:
     for the canonical conserved quantities (mass, energy, momentum,
     and species masses) per unit volume: $(\rho,\rho{E},\rho\vec{V},
     \rho{Y_s})$ from an agglomerated object array.  This data structure is intimately
-    related to helper functions :func:`join_conserved` and :func:`split_conserved`,
-    which pack and unpack (respectively) quantities to-and-from flat object
-    array representations of the data.
+    related to helper functions :func:`join_conserved`, :func:`split_conserved`,
+    and :func:`make_conserved`, which pack and unpack quantities to-and-from flat
+    object array representations of the data.
 
     .. note::
 
@@ -117,19 +117,20 @@ class ConservedVars:
 
         with the `ndim`-vector components of fluid velocity ($v_i$), and the
         `nspecies`-vector of species mass fractions ($Y_\alpha$). In total, the
-        fluid system has $N_{\text{eq}}$ = (`ndim + 2 + nspecies`) equations.
+        fluid system has $N_{\text{eq}}$ = (`ndim + 2 + nspecies`) equations, and
+        thus has $N_{\text{eq}}$ scalar components.
 
         Internally to `MIRGE-Com`, $\mathbf{Q}$ is stored as an object array
         (:class:`numpy.ndarray`) of :class:`~meshmode.dof_array.DOFArray`, one for
-        each component of the fluid $\mathbf{Q}$, i.e. a flat object array of
+        each scalar component of the fluid $\mathbf{Q}$, i.e. a flat object array of
         $N_{\text{eq}}$ :class:`~meshmode.dof_array.DOFArray`.
 
         To use this dataclass for a fluid CV-specific view on the content of
-        $\mathbf{Q}$, one can call :func:`split_conserved` to get a `ConservedVars`
+        $\mathbf{Q}$, one can call :func:`make_conserved` to get a `ConservedVars`
         dataclass object that resolves the fluid CV associated with each conservation
         equation::
 
-            fluid_cv = split_conserved(ndim, Q),
+            fluid_cv = make_conserved(ndim, scalar_quantities=Q),
 
         after which::
 
@@ -147,7 +148,8 @@ class ConservedVars:
     :example::
 
         Use `join_conserved` to create an agglomerated $\mathbf{Q}$ array from the
-        fluid conserved quantities (CV).
+        fluid conserved quantities (CV).  This may be useful for users who need to
+        interact with non-fluid simulation components.
 
         See the first example for the definition of CV, $\mathbf{Q}$, `ndim`,
         `nspecies`, and $N_{\text{eq}}$.
@@ -204,11 +206,11 @@ class ConservedVars:
 
         Presuming that `grad_q` is the agglomerated *MIRGE* data structure with the
         gradient data, this dataclass can be used to get a fluid CV-specific view on
-        the content of $\nabla\mathbf{Q}$. One can call :func:`split_conserved` to
+        the content of $\nabla\mathbf{Q}$. One can call :func:`make_conserved` to
         get a `ConservedVars` dataclass object that resolves the vector quantity
         associated with each conservation equation::
 
-            grad_cv = split_conserved(ndim, grad_q),
+            grad_cv = make_conserved(ndim, vector_quantities=grad_q),
 
         after which::
 
@@ -320,8 +322,13 @@ def join_conserved(dim, mass, energy, momentum, species_mass=None):
                            momentum=momentum, species_mass=species_mass)
 
 
-def make_conserved(dim, mass, energy, momentum, species_mass=None):
-    """Create :class:`ConservedVars` from separated conserved quantities."""
+def make_conserved(dim, mass=None, energy=None, momentum=None, species_mass=None,
+                   scalar_quantities=None, vector_quantities=None):
+    """Create :class:`ConservedVars` from separated or joined quantities."""
+    if scalar_quantities:
+        return split_conserved(dim, q=scalar_quantities)
+    if vector_quantities:
+        return split_conserved(dim, q=vector_quantities)
     return split_conserved(
         dim, _join_conserved(dim, mass=mass, energy=energy,
                              momentum=momentum, species_mass=species_mass)
--- a/mirgecom/euler.py
+++ b/mirgecom/euler.py
@@ -61,7 +61,7 @@ from grudge.eager import (
 )
 from mirgecom.fluid import (
     compute_wavespeed,
-    split_conserved,
+    make_conserved
 )
 
 from mirgecom.inviscid import (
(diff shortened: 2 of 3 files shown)


mirgecom/fluid.py
```diff
@@ -320,8 +322,13 @@ def join_conserved(dim, mass, energy, momentum, species_mass=None):
...
-    """Create :class:`ConservedVars` from separated conserved quantities."""
+def make_conserved(dim, mass=None, energy=None, momentum=None, species_mass=None,
+                   scalar_quantities=None, vector_quantities=None):
+    """Create :class:`ConservedVars` from separated or joined quantities."""
+    if scalar_quantities is not None:
+        return split_conserved(dim, q=scalar_quantities)
+    if vector_quantities is not None:
+        return split_conserved(dim, q=vector_quantities)
     return split_conserved(
         dim, _join_conserved(dim, mass=mass, energy=energy,
```
inducer, line 333:
```suggestion
        # Going through _join_conserved here to inherit its input
        # validation.
        dim, _join_conserved(dim, mass=mass, energy=energy,
```

mirgecom/fluid.py
```diff
@@ -320,8 +322,13 @@ def join_conserved(dim, mass, energy, momentum, species_mass=None):
...
 
-def make_conserved(dim, mass, energy, momentum, species_mass=None):
-    """Create :class:`ConservedVars` from separated conserved quantities."""
+def make_conserved(dim, mass=None, energy=None, momentum=None, species_mass=None,
+                   scalar_quantities=None, vector_quantities=None):
+    """Create :class:`ConservedVars` from separated or joined quantities."""
+    if scalar_quantities is not None:
+        return split_conserved(dim, q=scalar_quantities)
+    if vector_quantities is not None:
+        return split_conserved(dim, q=vector_quantities)
```
inducer, line 331: Not totally loving this: It's two functions operating on disjoint arguments jammed into the same function body. Separate one out, name it better? `conserved_from_array`? `conserved_from_join{t,ed}`?

MTCam: imo, the naming of the functions that create CV from different sorts of arguments should make it clear that a CV dataclass is returned, and ideally should be a clear read *at the call site* what type of data are being used to create the CV.  The `split` and `join` naming for the old routines is just too cryptic, and not intuitive enough, and offers no way to indicate the type of argument data.  Users are consistently confused about when, if, and which routines to use to create CV objects. 

After the CV array container stuff landed, the situation was much improved because we replaced `join_conserved` with `make_conserved` which handles the case where the components are already split-out:

`make_conserved(dim, mass=ary, energy=ary, momentum=ary[ary], ...)`

Candidates for replacing `split_conserved`

- `make_conserved(dim, scalar_quantities=arg)`, `make_conserved(dim, vector_quantities=arg)`, the current proposal in this PR
- `make_conserved_from_joined`  implies the existing `make_conserved` we already like should be changed to `make_conserved_from_split`.  Not my favorite solution.
- `create_conserved_from...` , blah

I am having a really hard time thinking of a different naming convention that makes sense from the users' perspective. It is no secret that I am a proponent of the approach where we offer a uniform interface to the user: `make_conserved` that has dispatch on the inside to sort out and deal with the arguments the user passed in.  Imo, the unified interface is more readable and accurately indicates at the call site what the user can expect to get back.

majosm: > imo, the naming of the functions that create CV from different sorts of arguments should make it clear that a CV dataclass is returned, and ideally should be a clear read _at the call site_ what type of data are being used to create the CV.

I think the latter can be accomplished using appropriate variable names rather than separate arguments:

```python
scalar_cv = make_conserved(dim, q=scalar_q)
```

The separate `scalar_quantities` and `vector_quantities` arguments are jarring to me.

> The `split` and `join` naming for the old routines is just too cryptic, and not intuitive enough, and offers no way to indicate the type of argument data. Users are consistently confused about when, if, and which routines to use to create CV objects.
>
> After the CV array container stuff landed, the situation was much improved because we replaced `join_conserved` with `make_conserved` which handles the case where the components are already split-out:
> 
> `make_conserved(dim, mass=ary, energy=ary, momentum=ary[ary], ...)`
> 
> Candidates for replacing `split_conserved`
> 
> * `make_conserved(dim, scalar_quantities=arg)`, `make_conserved(dim, vector_quantities=arg)`, the current proposal in this PR
> * `make_conserved_from_joined`  implies the existing `make_conserved` we already like should be changed to `make_conserved_from_split`.  Not my favorite solution.
> * `create_conserved_from...` , blah
> * `make_conserved_from_components` (for join) and `make_conserved_from_array` (for split). This is a distant 2nd for me.
> 
> I am having a really hard time thinking of a different naming convention that makes sense from the users' perspective. It is no secret that I am a proponent of the approach where we offer a uniform interface to the user: `make_conserved` that has dispatch on the inside to sort out and deal with the arguments the user passed in. Imo, the unified interface is more readable and accurately indicates at the call site what the user can expect to get back.

I agree that `split` and `join` are confusing (I can never remember which is `cv -> array` and which is `array -> cv`). How about something like `flat_from_conserved` and `conserved_from_flat`? Now that we have `make_conserved`, `flat_from_conserved` could take a cv directly instead of separate arguments.

MTCam: > I think the latter can be accomplished using appropriate variable names rather than separate arguments:
> 
> ```python
> scalar_cv = make_conserved(dim, q=scalar_q)
> ```
> 

Agreed, I like this (it is what is currently done in NS, actually) just fine and it will work for me.

The trouble I had with this (your suggested approach) appearing all over the code is that the naming of it, `q`, is very overloaded already in our markup and formulation. That's what led me to separate them out explicitly with more distinct language.

> I agree that `split` and `join` are confusing (I can never remember which is `cv -> array` and which is `array -> cv`). How about something like `flat_from_conserved` and `conserved_from_flat`? Now that we have `make_conserved`, `flat_from_conserved` could take a cv directly instead of separate arguments.

I just want to be sure we are talking about the same thing.  There are a couple of different situations we were trying to capture.  The original code had us doing this:

```python
q = join_conserved(dim, mass=<ary>, ...)  # join components into a CV
cv = split_conserved(dim, q) # split a flat, agglomerated object array into a CV
q = cv.join()  # create a flat agglomerated object array of cv components
```

This is how the code has always been (i.e. before CV array container support in the back-end). 

`join_conserved` was taking the conservation quantity components and creating an agglomerated array, and `split_conserved` would create a CV from the agglomerated array.  The nice thing about `join_conserved` is all the data and shape validations and massaging that is done on the inside to ensure that the arguments make sense, regardless of whether they are vector or scalar quantities.  The real ugly thing about it is (the language of it, which we've noted) and the user *must* follow those steps to get a CV from the components of a CV. (i.e. join, followed by split); two confusing things.

Now enter the CV array container stuff that @incuder, et al added to grudge, meshmode, etc.  This allowed us to all but drop `split_conserved`, because the fluid code never really needs to create a CV from an agglomerated `Q`.  We removed the external API for `join_conserved` in favor of `make_conserved` which has plainer language, and encapsulates the old two-step (join, then split) validation function and procedure of creating a CV array container dataclass from the components - which the user is usually creating, or manipulating.  Now (in main) we have this:

```python
cv = split_conserved(dim, q)  # in just a few places, maybe one test (rare)
cv = make_conserved(dim, mass=<ary>, ...)  # create CV from components (vector or scalar)
q = cv.join()   # convert a CV to a flat object array when you have to (rare)
```

Now the fact is that old `join_conserved` (now `make_conserved`), and `split_conserved` can already handle (thanks to the internal data handling and validation) vector and scalar quantities.  

If we want to separate them, then some symmetry can be achieved with:
`make_conserved_from_components(dim, mass=<ary>, ...)` and
`make_conserved_from_arrays(dim, q=scalar_quantities)` and
`make_conserved_from_arrays(dim, q=vector_quantities)`

where it is on the user to make sure to be clear about what they are passing as Q as @majosm suggests above.  Would this be acceptable?

MTCam: Just want to bump this convo back up @majosm, in case it has slipped through the cracks.

majosm: Two thoughts:

1. I wonder if `make_conserved` might be better off as an `__init__` method in `ConservedVars`, since that's basically what it's doing.
2. I'm finding the implementation kind of confusing with the lingering `split_conserved`/`join_conserved` usage and having `make_conserved` built on top of those (seems like it ought to be the other way around). It should be possible move the validation stuff from `join_conserved` into `make_conserved`/`__init__`.

I took a stab at doing something like this here: https://github.com/majosm/mirgecom/commit/997c0ccc17f7f2494b9f6b1e55817ce3d84d3116

MTCam: I like where you are going with that, @majosm.  I just want it to continue to work for other fluid quantities like fluxes, and gradients.  The viscous codes make much more use of it in that context, especially obviously for gradients.  The inviscid components exercise it in the context of fluxes.

MTCam: @majosm - do you want to finish out your attempt to move it into the constructor, or were you just providing the example that I should try to emulate?

majosm: Sure I can finish it. How should we go about testing it against NS?

MTCam: We can pilot it against y1-production.  Once it goes into this PR, we can test it out there.  This is another one of those PR's that is just to reduce the size of NS/Y1 against main - so whatever happens here needs to get rolled up into Y1/NS anyway.

majosm: MTCam/mirgecom#1

majosm:
OK to merge main into this @MTCam?

MTCam:
> OK to merge main into this @MTCam?

Yes, btw, merge at will.   If there is stuff in here that you are improving or working on - feel free to push/merge etc.

PR #411, opened 2021-06-29
Add constant CFL mode to examples

jlevine18:


examples/poiseuille-mpi.py
```diff
@@ -164,26 +166,39 @@ def poiseuille_soln(nodes, eos, cv=None, **kwargs):
...
+            from mirgecom.viscous import get_viscous_cfl
+            cfl_field = get_viscous_cfl(discr, eos, dt, state)
+            viz_fields.append(("cfl", cfl_field))
+            from grudge.op import nodal_max
+            checkpoint_cfl = nodal_max(discr, "vol", cfl_field)
+
+        dt = min(dt, t_remaining)
         return sim_checkpoint(discr, visualizer, eos, cv=state,
-                              vizname=casename, step=step,
+                              vizname=casename, step=step, cfl=checkpoint_cfl,
```
MTCam, line 192: `checkpoint_cfl` needs to get defined even for constant CFL mode. Something like...
`checkpoint_cfl = current_cfl` somewhere near the top of the routine or just before the block where CFL (or DT) is calculated.

jlevine18: AFAIK, it is. It's first defined on line 174 as current CFL, which should be correct for constant CFL mode.

examples/poiseuille-mpi.py
```diff
@@ -164,26 +166,39 @@ def poiseuille_soln(nodes, eos, cv=None, **kwargs):
...
-
     def my_rhs(t, state):
         return ns_operator(discr, eos=eos, boundaries=boundaries, cv=state, t=t)
 
     def my_checkpoint(step, t, dt, state):
+        t_remaining = max(0, t_final - t)
+        checkpoint_cfl = current_cfl
+        viz_fields = []
+
+        if constant_cfl is True:
```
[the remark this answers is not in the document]
jlevine18: Added as a visualization field if constant CFL mode is enabled (as otherwise, it's dt everywhere)

examples/wave-eager-mpi.py
```diff
@@ -112,15 +116,21 @@ def main(snapshot_pattern="wave-eager-{step:04d}-{rank:04d}.pkl", restart_step=N
...
-        dt = 0.4 / (nel_1d*order**2)
+    if not constant_cfl:
+        if local_mesh.dim == 2:
+            # no deep meaning here, just a fudge factor
+            dt = 0.7 / (nel_1d*order**2)
+        elif dim == 3:
+            # no deep meaning here, just a fudge factor
+            dt = 0.4 / (nel_1d*order**2)
+        else:
+            raise ValueError("don't have a stable time step guesstimate")
```
MTCam, line 127: I was actually going to say we should replace this hokey junk with just constant DT mode, but as I am writing now I think I agree with your notion to leave it alone.  It sort of auto-tunes the constant DT (hokey or not), obviating the user's configuration-specific choice for DT.

Let's see what shakes out if we ask the author.

MTCam: Ok, it looks like we want to go ahead and rip out this stuff.   Can you please replace the kludge here with a single call to the timestep estimator?   Actually - I've done a dumb thing to you.  The wavespeed and grid never change in this simulation (unlike a fluid sim) - so a single call to the timestep estimator is sufficient for the wave examples.  They do not need a constant CFL mode.  Sigh :(

jlevine18: I just removed it. I also realized that dt was constant after I pushed up- I'm not exactly sure why I kept the old dt code in there in the first place.

MTCam (approved):
Thanks for your work on this! 👍

PR #420, opened 2021-07-10
Add local maximum species diffusivity calculation

jlevine18:


mirgecom/viscous.py
```diff
@@ -275,3 +280,37 @@ def get_viscous_cfl(discr, eos, dt, cv):
...
+    cv: :class:`~mirgecom.fluid.ConservedVars`
+        Fluid solution
+    """
+    actx = cv.array_context
+
+    if(transport is None or transport._d_alpha.size == 0):
+        return 0 * cv.mass / cv.mass
+
+    species_diffusivity = transport.species_diffusivity(eos, cv)
+    stacked_diffusivity = cl.array.stack([x[0] for x in species_diffusivity])
```
thomasgibson, line 303: We probably want to avoid the openCL specificness here. Instead, I would just use `stack` from the array context like this:
```suggestion
    actx = cv.array_context
    species_diffusivity = transport.species_diffusivity(eos, cv)
    stacked_diffusivity = actx.np.stack([x[0] for x in species_diffusivity])
```

MTCam:
This is nice work @jlevine18, let's just get @thomasgibson and @majosm to have a quick look to make sure we aren't screwing the pooch, and we can merge it.   Just keep in mind - this merges to y1/ns, not main.

mirgecom/viscous.py
jlevine18: Switched to the actx version. I didn't realize you could do that; thank you so much for the fix.

test/test_viscous.py
```diff
@@ -240,6 +240,45 @@ def test_diffusive_heat_flux(actx_factory):
...
 
+@pytest.mark.parametrize("dim", [1, 2, 3])
+def test_local_max_species_diffusivity(actx_factory, dim):
+    actx = actx_factory()
+    nel_1d = 5
+
+    from meshmode.mesh.generation import generate_regular_rect_mesh
+
+    mesh = generate_regular_rect_mesh(
+        a=(1.0,) * dim, b=(2.0,) * dim, n=(nel_1d,) * dim
```
majosm, line 251:
```suggestion
        a=(1.0,) * dim, b=(2.0,) * dim, nelements_per_axis=(nel_1d,) * dim
```
(`n` is deprecated.)

test/test_viscous.py
```diff
@@ -240,6 +240,45 @@ def test_diffusive_heat_flux(actx_factory):
         assert discr.norm(j[ispec+1] - exact_j, np.inf) < tol
 
 
+@pytest.mark.parametrize("dim", [1, 2, 3])
+def test_local_max_species_diffusivity(actx_factory, dim):
+    actx = actx_factory()
+    nel_1d = 5
```
majosm, line 246:
```suggestion
    nel_1d = 4
```
(The `n` parameter to `generate_regular_rect_mesh` was actually a number of points per dimension; this change makes my suggested `nelements_per_axis` version equivalent to your original code.)

mirgecom/viscous.py
```diff
@@ -275,3 +279,37 @@ def get_viscous_cfl(discr, eos, dt, cv):
...
+    cv: :class:`~mirgecom.fluid.ConservedVars`
+        Fluid solution
+    """
+    actx = cv.array_context
+
+    if(transport is None or transport._d_alpha.size == 0):
+        return 0 * cv.mass / cv.mass
+
+    species_diffusivity = transport.species_diffusivity(eos, cv)
+    stacked_diffusivity = actx.np.stack([x[0] for x in species_diffusivity])
```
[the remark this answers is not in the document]
thomasgibson: Good catch. And if `x` is a DOFArray, you shouldn't need to do `x[0]` in the list comprehension I think.

mirgecom/viscous.py
```diff
@@ -275,3 +279,37 @@ def get_viscous_cfl(discr, eos, dt, cv):
...
+    n_species, ni1, ni0 = stacked_diffusivity.shape
+
+    # fun fact: arraycontext needs these exact loop names to work (even though a
+    # loopy kernel can have whatever iterator names the user wants)
+    # TODO: see if the opposite order [i0, i1, i2] is faster due to higher
+    # spatial locality, causing fewer cache misses
+    knl = arraycontext.make_loopy_program(
+        "{ [i1,i0,i2]: 0<=i1<ni1 and 0<=i0<ni0 and 0<=i2<n_species}",
+        "out[i1,i0] = max(i2, a[i2,i1,i0])"
+    )
```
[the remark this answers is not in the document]
jlevine18: There isn't with loopy, but arraycontext for some reason designates i0 as an "outer" loop and i1 as an "inner" loop and won't work without there being an iterator designated i0 and i1

mirgecom/viscous.py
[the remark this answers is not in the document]
jlevine18: `transport.species_diffusivity` should return a tuple of `DOFArray`s. I was doing it this way because `DOFArray`s don't have some features that their underlying arrays do (like shape) but I can change it if you want.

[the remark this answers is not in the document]
majosm: Yeah, `DOFArray`s won't have a single shape, because the per-mesh-group arrays inside them may have different shapes (corresponding to different element types/counts). We will want to make sure we're handling the multiple mesh group case here; this will probably mean calling the loopy kernel once per group.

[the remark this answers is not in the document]
jlevine18: Ok, I looped over all possible mesh groups, calling the kernel once for each one. Unfortunately, if size can change, I have to make a new kernel each time as far as I know, but I don't think there's much overhead there.

MTCam (changes requested):
After discussions with @majosm and @thomasgibson , we plan to memoize the loopy kernels onto the discretization collection with `number of species` as the unique key.  Please redirect me if I got this wrong guys :)

mirgecom/viscous.py
```diff
@@ -275,3 +284,37 @@ def get_viscous_cfl(discr, eos, dt, cv):
...
+    ----------
+    actx: :class:`arraycontext.ArrayContext`
+        Array context to use
+    discr: :class:`grudge.eager.EagerDGDiscretization`
+        the discretization to use
+    d_alpha: np.ndarray
+        Species diffusivities
+    """
+    return_dof = []
+    for i in range(len(d_alpha[0])):
```
thomasgibson, line 302: Based on a discussion in our slack: should we explicitly check for `d_alpha` being `np.empty`?

MTCam: (4f95721)

jlevine18: It's not np.empty that's the problem, it's of size 0. See my last change

majosm: `d_alpha is not np.empty` isn't going to do what you want (it will check whether `d_alpha` is the function `np.empty`). `d_alpha.size > 0` or `len(d_alpha) > 0` should work.

MTCam: Perfect, thanks @majosm.  Do you know if there is a performance difference?   Does len(d_alpha) have constant time? I guess I suspect not if it is allowed to be non-contiguous.

majosm: For 1D arrays I would think they'd be the same (`len` just calls `ndarray.__len__()`, which I assume just returns `shape[0]`). For higher dimensional arrays `size` returns a different result (it's the product of `shape`).

PR #424, opened 2021-07-12
Update Y1 CI testing so that it also working in main.

MTCam:
Sorry - addition of Y1 tests could not have been tested in main until merge.  This should fix them.  When there are no changes to apply - skip trying to apply changes.


matthiasdiener (approved):

MTCam:
Sorry had to make a quick tweak for forks of *MIRGE-Com*.  I had not tested that properly before.  Apparently there is nothing in the CI environment that indicates that the change set is coming from a fork.  I mistakenly used REPOSITORY_OWNER before - but that appears to be "illinois-ceesd" in all cases, ha.

matthiasdiener (approved):

PR #426, opened 2021-07-12
Run CI on MTCam#1

majosm:
MTCam/mirgecom#1


MTCam:
I don't think we want/need this any more, but we can open it back up if needed.

PR #428, opened 2021-07-12
Restore visualization timers in examples

majosm:


majosm:
I think at the moment the logic is a little too awkward to be put directly in the driver, but it seems fixable (see illinois-ceesd/logpyle#71). In some cases we may still want to pass a timer into a function so that it can time only a piece of the call (like if we want lower-level I/O timings, say), but in this case I think it's something that could be done in the driver. Let's try that in a followup PR.

MTCam (approved):
lgtm! We'll need to remember to update the new viscous examples and stuff so that they are consistent, but can't do that here!

matthiasdiener (approved):

majosm:
> lgtm! We'll need to remember to update the new viscous examples and stuff so that they are consistent, but can't do that here!

Should I make a PR into NS with the rest?

PR #432, opened 2021-07-13
Add driver/callback docs

### LINKED ISSUE #427: Missing documentation on callbacks
I tried to read up on callbacks so that I can use them in my drivers, but I couldn't find much of any documentation on them.  What are they, what do they do, and why should I use them?   Can I just modify the state willy-nilly? (#415)

We should add some documentation on callbacks including cautionary statements about state modification, and maybe include some code snippet examples.

MTCam:
fixes #427


doc/support/drivers.rst
```diff
@@ -0,0 +1,60 @@
...
+   |     k1 = rhs(t, state)
+   |     k2 = rhs(t+dt/2, state + dt/2*k1)
+   |     k3 = rhs(t+dt/2, state + dt/2*k2)
+   |     k4 = rhs(t+dt, state + dt*k3)
+   |
+   |     return state + dt/6*(k1 + 2*k2 + 2*k3 + k4)
+
+All user callbacks *must* follow the signature provided above and in the documentation
+of :mod:`~mirgecom.steppers`. That is, user callbacks must return the *state*, and
+*dt* to the stepper, even if it just returns the values passed in by the stepper.
```
MTCam, line 59: Let's move this above the 'important' note.

doc/support/drivers.rst
```diff
@@ -0,0 +1,60 @@
...
+   enough to ensure realizability, as observed from the snippet below:
+
+   | def rk4_step(state, t, dt, rhs):
+   |     """Take one step using the fourth-order Classical Runge-Kutta method."""
+   |     k1 = rhs(t, state)
+   |     k2 = rhs(t+dt/2, state + dt/2*k1)
+   |     k3 = rhs(t+dt/2, state + dt/2*k2)
+   |     k4 = rhs(t+dt, state + dt*k3)
+   |
+   |     return state + dt/6*(k1 + 2*k2 + 2*k3 + k4)
```
MTCam, line 55: I'd maybe move this above the 'important' note warning users about modification to the state.  Actually, we don't use or have any timesteppers that keep a history like that, so the note may be causing unneeded concern.

```suggestion
.. important::
   Operations that modify the state like solution limiting and filtering must
   be applied in the *pre_step* or *post_step* callback *and* inside the *RHS*
   to ensure that a limited or filtered state is both passed to the time integrator
   and used in the operators that make up the *RHS*.

   Consider this snippet from the *RK4* time integrator:

   | def rk4_step(state, t, dt, rhs):
   |     """Take one step using the fourth-order Classical Runge-Kutta method."""
   |     k1 = rhs(t, state)
   |     k2 = rhs(t+dt/2, state + dt/2*k1)
   |     k3 = rhs(t+dt/2, state + dt/2*k2)
   |     k4 = rhs(t+dt, state + dt*k3)
   |
   |     return state + dt/6*(k1 + 2*k2 + 2*k3 + k4)
   
   Applying the state-modification inside the *RHS* function ensures that each
   stage update is using a modified state, and applying the state-modification 
   in the *pre_step* or *post_step* callback ensures that the incoming *state*
   or outgoing updated *state* are also properly modified.
```

doc/support/drivers.rst
```diff
@@ -0,0 +1,68 @@
+Writing drivers for :mod:`mirgecom`
+==============================================
```
tulioricci, line 2: Trim this a bit to align with the text

doc/support/drivers.rst
```diff
@@ -0,0 +1,68 @@
...
+
+:mod:`mirgecom` provides a great deal of flexibility in what users
+are able to do with their drivers. This section of the documentation
+is designed to introduce the tools, and recommended practices for
+designing drivers for :mod:`mirgecom` simulations.
+
```
tulioricci, line 8: I was wondering if we should add a section about the RHS too, particular because of make_operator_fluid_states (i.e., we gotta pass the limiter). These covers the most important pitfalls concerning limiters.

MTCam: I agree that we need better documentation and developer guidance for drivers. It would be my preference to document this and your suggestion in a sort of detailed driver walk-through.   For now, maybe it would suffice to something like this:

```
  Consider these snippets from the *RK4* time integrator, and example *rhs* function:

   | def rk4_step(state, t, dt, rhs):
   |     """Take one step using the fourth-order Classical Runge-Kutta method."""
   |     k1 = rhs(t, state)
   |     k2 = rhs(t+dt/2, state + dt/2*k1)
   |     k3 = rhs(t+dt/2, state + dt/2*k2)
   |     k4 = rhs(t+dt, state + dt*k3)
   |
   |     return state + dt/6*(k1 + 2*k2 + 2*k3 + k4)
   
   | def rhs(t, state):
   |     """Return the RHS of the conservation system."""
   |     filtered_state = apply_filter(state)
   |     return rhs_operator(t, filtered_state)
```

What I really want in the end is documentation that covers an integrated driver with all of its components and constructs.   Would be nice if those docs could be (largely) generated from the specific examples and their inline documentation, but I think we need to restructure the example drivers somewhat in order to easily wedge their docs into the automatic online documentation.

tulioricci: Capisce. There are a few PRs concerning driver updates, so we can put more doc details in those then.

tulioricci:
:+1:

tulioricci (approved):
:+1:

PR #435, opened 2021-07-14
Vanquish timestep estimation kludge in wave examples. 

MTCam:
Switches it up so that wave examples use grudge's characteristic lengthscales for a real dt estimate.  This change was implemented upstream by UG researcher @jlevine18.


thomasgibson (commented):
LGTM! Thanks for de-kludging the code :) Would it be possible while we're at it to do the same for the heat equation/diffusion example? The calculation will give you something that will look similar to what you need to do in the viscous Navier-Stokes case anyways

MTCam:
> LGTM! Thanks for de-kludging the code :) Would it be possible while we're at it to do the same for the heat equation/diffusion example? The calculation will give you something that will look similar to what you need to do in the viscous Navier-Stokes case anyways

Yes!  We have the diffusive DT estimates in NS/Y1 right now - but those can't be rolled into this atm because they depend on the viscous junk.

jlevine18:
> > LGTM! Thanks for de-kludging the code :) Would it be possible while we're at it to do the same for the heat equation/diffusion example? The calculation will give you something that will look similar to what you need to do in the viscous Navier-Stokes case anyways
> 
> Yes! We have the diffusive DT estimates in NS/Y1 right now - but those can't be rolled into this atm because they depend on the viscous junk.

By themselves, I don't think the diffusive elements depend on viscous, but they do depend on a transport model

thomasgibson (approved):

MTCam:
> > > LGTM! Thanks for de-kludging the code :) Would it be possible while we're at it to do the same for the heat equation/diffusion example? The calculation will give you something that will look similar to what you need to do in the viscous Navier-Stokes case anyways
> > 
> > 
> > Yes! We have the diffusive DT estimates in NS/Y1 right now - but those can't be rolled into this atm because they depend on the viscous junk.
> 
> By themselves, I don't think the diffusive elements depend on viscous, but they do depend on a transport model

You are right @jlevine18; they depend on the transport model. I was just mentioning it because the transport model is used only in viscous flows, and is not available in our main branch.  The transport model is introduced in changes coming through the addition of viscous terms in navier-stokes.

PR #437, opened 2021-07-14
Robustify production testing a little

MTCam:
- Replaces the old scripts with `production-install.sh` `production-driver-install.sh`, and `production-testing-env.sh`
- Developers only ever need to set up `production-testing-env.sh` which only sets variables
- Main never uses customizations - it only ever tests y1-production and nozzle driver without changes
- Supports specifying production branch outright
- Supports specifying production driver branch outright
- Add a gob of documentation to explain the settings/vars
- Makes explicit the defaulting of the owners to the local repo


MTCam (commented):
Heh, this is a dumb change.  Not sure what I thought this would do.

MTCam:
Ok - this now fixes the Y1 testing issues and cannot be messed up from merges.

.ci-support/install.sh
```diff
@@ -1,7 +1,7 @@
 if [ "$(uname)" = "Darwin" ]; then
 PLATFORM=MacOSX
 brew update
-brew upgrade
+# brew upgrade
```
thomasgibson, line 4: Any reason to leave this here? We could also leave a short note explaining why this is commented out

thomasgibson (approved):
LGTM! Just a minor comment/question.

.ci-support/install.sh
MTCam: I overlooked/forgot it, I say we should eliminate with prejudice.

PR #438, opened 2021-07-15
add mirgecompare script

matthiasdiener:
Implemented by @annabellej

### DIFF
--- a/.ci-support/production-testing-env.sh
+++ b/.ci-support/production-testing-env.sh
@@ -15,7 +15,7 @@ set -x
 # patched by the incoming development. The following vars control the
 # production environment:
 #
-export PRODUCTION_BRANCH="y1-production-testing"   # The base production branch to be installed by emirge
+# export PRODUCTION_BRANCH=""   # The base production branch to be installed by emirge
 # export PRODUCTION_CHANGE_FORK=""  # The fork/home of production changes (if any)
 # export PRODUCTION_CHANGE_BRANCH=""  # Branch from which to pull prod changes (if any)
 #
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -97,6 +97,13 @@ jobs:
             . "$MINIFORGE_INSTALL_DIR/bin/activate" testing
             export XDG_CACHE_HOME=/tmp
             examples/run_examples.sh ./examples
+ 
+        - name: Test lazy accuracy
+          run: |
+            MINIFORGE_INSTALL_DIR=.miniforge3
+            . "$MINIFORGE_INSTALL_DIR/bin/activate" testing
+            export XDG_CACHE_HOME=/tmp
+            cd examples && ./test_loopy_accuracy.sh
 
     doc:
         name: Documentation
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -91,6 +91,13 @@ jobs:
           run: |
             . .ci-support/install.sh
 
+        - name: Test lazy accuracy
+          run: |
+            MINIFORGE_INSTALL_DIR=.miniforge3
+            . "$MINIFORGE_INSTALL_DIR/bin/activate" testing
+            export XDG_CACHE_HOME=/tmp
+            cd examples && ./test_lazy_accuracy.sh
+
         - name: Run examples
           run: |
             MINIFORGE_INSTALL_DIR=.miniforge3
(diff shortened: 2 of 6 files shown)


MTCam (commented):
COoL, want to add some tests or examples?

matthiasdiener:
> COoL, want to add some tests or examples?

The main repo already has some tests (https://github.com/annabellej/mirgecompare) 
I think it's hard to add tests to mirgecom at the moment, but we could add some as soon as we have (e.g.) some lazy-eval examples in main.

MTCam:
> > COoL, want to add some tests or examples?
> 
> The main repo already has some tests ([annabellej/mirgecompare](https://github.com/annabellej/mirgecompare))
> I think it's hard to add tests to mirgecom at the moment, but we could add some as soon as we have (e.g.) some lazy-eval examples in main.

Interesting - is this capability depending on lazy?  I would like to talk about why it is hard to add tests for this in mirgecom.  Can we fix that?

matthiasdiener:
> > > COoL, want to add some tests or examples?
> > 
> > 
> > The main repo already has some tests ([annabellej/mirgecompare](https://github.com/annabellej/mirgecompare))
> > I think it's hard to add tests to mirgecom at the moment, but we could add some as soon as we have (e.g.) some lazy-eval examples in main.
> 
> Interesting - is this capability depending on lazy? I would like to talk about why it is hard to add tests for this in mirgecom. Can we fix that?

We could do several things:
1. compare to a "known-good" result that is part of the repo
2. compare different MPI configs (e.g. single rank vs. two ranks)
3. compare different OpenCL devices (can't do that in CI I think)
4. compare lazy vs eager execution (imho, this would be the most interesting)

I think 2.) above might the easiest to implement at the moment - what do you think?

MTCam:
> We could do several things:
> 
> 1. compare to a "known-good" result that is part of the repo
> 2. compare different MPI configs (e.g. single rank vs. two ranks)
> 3. compare different OpenCL devices (can't do that in CI I think)
> 4. compare lazy vs eager execution (imho, this would be the most interesting)
> 
> I think 2.) above might the easiest to implement at the moment - what do you think?

If we can *generate* correct results on the fly (which we can for some cases), (1) would be useful.  It would be quite interesting to be able to compare using option (2) as well.  Both should be easy to implement.

inducer:
> The main repo already has some tests (annabellej/mirgecompare)

Which repo is going to be the "authoritative" one for this code? IMO, that's where tests and code should live. Having code copied in multiple places is not a good idea (again IMO).

matthiasdiener:
> > We could do several things:
> > 
> > 1. compare to a "known-good" result that is part of the repo
> > 2. compare different MPI configs (e.g. single rank vs. two ranks)
> > 3. compare different OpenCL devices (can't do that in CI I think)
> > 4. compare lazy vs eager execution (imho, this would be the most interesting)
> > 
> > I think 2.) above might the easiest to implement at the moment - what do you think?
> 
> If we can _generate_ correct results on the fly (which we can for some cases), (1) would be useful. It would be quite interesting to be able to compare using option (2) as well. Both should be easy to implement.

I just realized: option (2) isn't really possible at the moment I think, because the structure of the output files differs with different rank numbers, right? This isn't supported yet AFAIK. In that case, we could just run the same experiment twice and check that the output matches.

> Which repo is going to be the "authoritative" one for this code? IMO, that's where tests and code should live. Having code copied in multiple places is not a good idea (again IMO).

I think this repo should become the authoritative one (IMO).

MTCam:
> > > 1. compare to a "known-good" result that is part of the repo
> > > 2. compare different MPI configs (e.g. single rank vs. two ranks)
> > > 3. compare different OpenCL devices (can't do that in CI I think)
> > > 4. compare lazy vs eager execution (imho, this would be the most interesting)

> I just realized: option (2) isn't really possible at the moment I think, because the structure of the output files differs with different rank numbers, right? This isn't supported yet AFAIK. In that case, we could just run the same experiment twice and check that the output matches.

Technically, the structure is the same I think. You get one file per MPI rank, the comparison utility doesn't actually need to assemble these together, but to do an effective solution comparison, it will need to conjure the element correspondence between the two decomps.  I do not know what that looks like in the context of mirgecom.   

In the short term, we could postpone the notion of doing an element-by-element comparison and just look at field limits. That will already indicate differences between runs.

> 
> > Which repo is going to be the "authoritative" one for this code? IMO, that's where tests and code should live. Having code copied in multiple places is not a good idea (again IMO).
> 
> I think this repo should become the authoritative one (IMO).

Agreed.   However there are already tests and examples existing in the other repo, and porting them to this repo seems non-trivial, so does that make it the de-facto main one already?

MTCam (approved):
I filed a collection of issues in the main repo, changes that I would like to have. However, I see no reason to delay this.  We are now using it to check that lazy run results match eager to within 1e-15, and that is working well.   It would be terrific to have this facility available in main so we can use it to build up tests of lazy on the production problems.

PR #439, opened 2021-07-15
Revert production-testing-env.sh after inadvertently changing it.

MTCam:
A merge twisted this file - but it's ok for main. It would be nice if the default behavior (even for branches) was to test against the current y1-production.  This change makes it so.

### DIFF
--- a/.ci-support/production-testing-env.sh
+++ b/.ci-support/production-testing-env.sh
@@ -15,7 +15,7 @@ set -x
 # patched by the incoming development. The following vars control the
 # production environment:
 #
-export PRODUCTION_BRANCH="y1-production-testing"   # The base production branch to be installed by emirge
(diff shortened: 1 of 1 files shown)


.ci-support/production-testing-env.sh
```diff
@@ -15,7 +15,7 @@ set -x
 # patched by the incoming development. The following vars control the
 # production environment:
 #
-export PRODUCTION_BRANCH="y1-production-testing"   # The base production branch to be installed by emirge
+# export PRODUCTION_BRANCH=""   # The base production branch to be installed by emirge
```
thomasgibson, line 18:
```suggestion
export PRODUCTION_BRANCH=""   # The base production branch to be installed by emirge
```
?

thomasgibson (approved):

PR #441, opened 2021-07-17
Laxify the examples in production

MTCam:
This unifies the argparse and main signature across fluidy examples, and does the easy part of enabling lazy execution.


MTCam:
Be aware - this is just merging to y1-production, not main.   Just would like to get your eyes-on to make sure this is how we want it. I skipped over the wave examples atm.

PR #449, opened 2021-07-19
Lift Euler module to production

MTCam:
This change set makes euler@main look like the code used in production:

- Makes Euler better, more succinct, cleaner arrangement
- Updates the boundary treatments to be closer to production and final vision
- Unifies all the CV/make_conserved constructions, eliminating public split/join facilities
- Greatly reduces the size of NS and y1-production against main

Things to keep in mind:
- Boundary treatments do not need to be perfect, we have a long-standing effort to refactor them and get them perfect in #341 , and others.   
- CV utilities (e.g. make_conserved) do not need to be perfect.  We have an ongoing effort to refactor CV creation to our liking in #399

In the interest of making main like production as fast as possible, I would prefer if we do not postpone these changes waiting for resolution on the mentioned refactoring efforts. The natural follow-ons from this PR are navier-stokes, then artificial viscosity.  At that point y1-production will be down to driver stuff only.


mirgecom/boundary.py
```diff
@@ -57,32 +147,33 @@ def __init__(self, userfunc):
...
-        actx = cv.array_context
-
-        boundary_discr = discr.discr_from_dd(btag)
-        nodes = thaw(actx, boundary_discr.nodes())
-        ext_soln = self._userfunc(nodes, **kwargs)
-        int_soln = discr.project("vol", btag, cv)
-        return TracePair(btag, interior=int_soln, exterior=ext_soln)
+        from warnings import warn
+        warn("Do not use PrescribedBoundary; use PrescribedInvscidBoundary. This"
+             "boundary type  will disappear soon.", DeprecationWarning, stacklevel=2)
```
inducer, line 152: Put a deadline on it.

inducer (approved):
Did a quick scroll, didn't see anything I'd object to.

mirgecom/boundary.py
MTCam: The reason I made it so tight is we will likely re-use this boundary type in our boundary refactor effort.  We've already gone around and updated all the examples and production stuff.  Nobody out there should be still using this boundary.

majosm:
> In the interest of making main like production as fast as possible, I would prefer if we do not postpone these changes waiting for resolution on the mentioned refactoring efforts. The natural follow-ons from this PR are navier-stokes, then artificial viscosity. At that point y1-production will be down to driver stuff only.

Are we going to try to do this all before the review? I kind of wonder if this is a good use of time. I have a handful of things I'd like to review here (and likely in NS/AV too), but it's going to take me away from other things I should be working on right now. Can we take care of the big merges post-review, when things are a little more relaxed? Or do we need it in `main` now?

MTCam:
> > In the interest of making main like production as fast as possible, I would prefer if we do not postpone these changes waiting for resolution on the mentioned refactoring efforts. The natural follow-ons from this PR are navier-stokes, then artificial viscosity. At that point y1-production will be down to driver stuff only.
> 
> Are we going to try to do this all before the review? I kind of wonder if this is a good use of time. I have a handful of things I'd like to review here (and likely in NS/AV too), but it's going to take me away from other things I should be working on right now. Can we take care of the big merges post-review, when things are a little more relaxed? Or do we need it in `main` now?

I strongly prefer to get y1-production stuff into main asap.  imo, we can work on refactoring as time permits.  Our focus should be on production - and since everything we are trying to merge is exercised and working in production, there doesn't seem to be much harm in moving mirgecom as close to production as we can as quickly as possible.

mirgecom/flux.py
```diff
@@ -72,7 +74,4 @@ def lfr_flux(cv_tpair, flux_func, normal, lam):
         object array of :class:`meshmode.dof_array.DOFArray` with the
         Lax-Friedrichs/Rusanov flux.
     """
-    flux_avg = 0.5*(flux_func(cv_tpair.int)
-                    + flux_func(cv_tpair.ext))
-    return flux_avg @ normal - 0.5*lam*(cv_tpair.ext
-                                        - cv_tpair.int)
+    return f_tpair.avg@normal - lam*jump(cv_tpair)/2
```
thomasgibson, line 77: `cv_tpair.diff` is a thing now, in case you want to use that

mirgecom/operators.py
```diff
@@ -0,0 +1,71 @@
...
+    Parameters
+    ----------
+    trace_pair: :class:`grudge.sym.TracePair`
+    Represents the quantity for which the jump is to be calculated.
+
+    Returns
+    -------
+    like(trace_pair.int)
+    """
+    return trace_pair.ext - trace_pair.int
```
thomasgibson, line 50: `TracePair` defines a `diff` property as exactly this, so we definitely don't need this function I think.

mirgecom/operators.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def dg_div(discr, vol_flux, bnd_flux):
```
thomasgibson, line 53: Eh, to be honest, I'm not sure this really helps much. I sort of like seeing all the DG-related constructions self-contained in the Euler routines.

MTCam: I do see where you are coming from - especially as a DG person.  I like this wrapper better though because it allows me to write what I see as an uncluttered expression that looks very much like the equation I am trying to implement, and without having to worry about dealing with the sign conventions.

mirgecom/flux.py
MTCam: (56e30d3)

mirgecom/operators.py
MTCam: (56e30d3)

mirgecom/operators.py
thomasgibson: Sure, I can live with this if you feel it's very helpful

thomasgibson (approved):

mirgecom/boundary.py
```diff
@@ -35,15 +41,99 @@
...
+    .. automethod:: boundary_pair
+    """
+
+    def inviscid_boundary_flux(self, discr, btag, cv, eos, **kwargs):
+        """Get the inviscid part of the physical flux across the boundary *btag*."""
+        raise NotImplementedError()
+
+    def boundary_pair(self, discr, btag, cv, eos, **kwargs):
+        """Get the interior and exterior solution (*u*) on the boundary."""
+        raise NotImplementedError()
```
thomasgibson, line 75: Minor suggestion, but since the BC class is already an abstract class, might as well use the abstract methods? Anyways, consider this an optional request if you don't care for it :P
```suggestion
    @abstractmethod
    def inviscid_boundary_flux(self, discr, btag, cv, eos, **kwargs):
        """Get the inviscid part of the physical flux across the boundary *btag*."""

    @abstractmethod
    def boundary_pair(self, discr, btag, cv, eos, **kwargs):
        """Get the interior and exterior solution (*u*) on the boundary."""
```
?

MTCam: I'm game - bout to push the button.  I fear chaos breaking out.

mirgecom/operators.py
```diff
@@ -0,0 +1,49 @@
...
+    r"""Compute a DG divergence for the flux vectors given in *vol_flux* and *bnd_flux*.
+
+    Parameters
+    ----------
+    discr: grudge.eager.EagerDGDiscretization
+        the discretization to use
+    vol_flux: np.ndarray
+        the volume flux term in the element
```
majosm, line 39: I think we could just call this `u` and describe it as the quantity the div is being applied to. Doesn't need to be a "flux", per se.

mirgecom/operators.py
```diff
@@ -0,0 +1,49 @@
...
+    bnd_flux: np.ndarray
+        the boundary fluxes across the faces of the element
```
majosm, line 41: I'd maybe call this just `flux` or `facial_flux`. `bnd_flux` is easy to misinterpret here.

test/test_fluid.py
```diff
@@ -76,11 +76,8 @@ def test_velocity_gradient_sanity(actx_factory, dim, mass_exp, vel_fac):
     velocity = vel_fac * nodes
     mom = mass * velocity
 
-    q = join_conserved(dim, mass=mass, energy=energy, momentum=mom)
-    cv = split_conserved(dim, q)
-
-    grad_q = obj_array_vectorize(discr.grad, q)
-    grad_cv = split_conserved(dim, grad_q)
+    cv = make_conserved(dim, mass=mass, energy=energy, momentum=mom)
+    grad_cv = make_conserved(dim, q=obj_array_vectorize(discr.grad, cv.join()))
```
majosm, line 80: Is it ok to just use `discr.grad` here, with no flux terms?

mirgecom/boundary.py
```diff
@@ -35,15 +41,99 @@
...
+    @abstractmethod
+    def inviscid_boundary_flux(self, discr, btag, cv, eos, **kwargs):
+        """Get the inviscid part of the physical flux across the boundary *btag*."""
+
+    @abstractmethod
+    def boundary_pair(self, discr, btag, cv, eos, **kwargs):
+        """Get the interior and exterior solution (*u*) on the boundary."""
+
+
+class PrescribedInviscidBoundary(FluidBC):
```
majosm, line 78: I find this base class really hard to decipher (and the [viscous version](https://github.com/illinois-ceesd/mirgecom/blob/3586065551999bd8fc6da8511f9e889f363dbdcd/mirgecom/boundary.py#L136-L159) even more so) due to all of the optional callbacks. I'd like to see this get refactored at some point (maybe into a set of helper functions?). But I guess I can tolerate it going into main as is for now.

majosm (commented):
LGTM, apart from the boundary base class and a couple of minor things.

test/test_fluid.py
MTCam: It is safe here where I am just testing the form and function of the product rule grad functions.  We just want the grad of a continuous function in space for that test.

mirgecom/boundary.py
```diff
@@ -35,15 +41,99 @@
...
+        if not self._inviscid_facial_flux_func:
+            self._inviscid_facial_flux_func = inviscid_facial_flux
+        self._fluid_soln_func = fluid_solution_func
+        self._fluid_soln_flux_func = fluid_solution_flux_func
+
+    def _boundary_quantity(self, discr, btag, quantity, **kwargs):
+        """Get a boundary quantity on local boundary, or projected to "all_faces"."""
+        if "local" in kwargs:
+            if kwargs["local"]:
+                return quantity
```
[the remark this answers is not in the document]
MTCam: Actually - let's defer this.  You are right, and clearly this was a pain point for me or I would not have delved into kwargs to try to deal with it.  Apparently it is used only in the viscous boundary junk so I've just removed it from this change set.

MTCam:
> LGTM, apart from the boundary base class and a couple of minor things.

Thanks, @majosm!  I've addressed your comments. It'll take a few goes to get the change in divergence operator propagated back up to y1, but I'll plan on merging once I've got the cascade of green checks again.

PR #452, opened 2021-07-20
Add tests for lazily-compiled operators

majosm:
Only diffusion/Euler for now. Production version with NS/AV [here](https://github.com/majosm/mirgecom/blob/lazy-operator-tests-ns-av/test/test_lazy.py).

### DIFF
--- a/test/test_lazy.py
+++ b/test/test_lazy.py
@@ -73,14 +73,15 @@ def op_test_data(ctx_factory):
     return eager_actx, lazy_actx, get_discr
 
 
-def _rel_linf_error(discr, x, y):
+def _rel_linf_error(discr, actual, expected):
     from mirgecom.fluid import ConservedVars
-    if isinstance(x, ConservedVars):
-        return _rel_linf_error(discr, x.join(), y)
-    if isinstance(y, ConservedVars):
-        return _rel_linf_error(discr, x, y.join())
+    if isinstance(actual, ConservedVars):
+        return _rel_linf_error(discr, actual.join(), expected)
+    if isinstance(expected, ConservedVars):
+        return _rel_linf_error(discr, actual, expected.join())
     return np.max(obj_array_vectorize_n_args(
-        lambda a, b: discr.norm(a - b, np.inf) / discr.norm(b, np.inf), x, y))
+        lambda a, b: discr.norm(a - b, np.inf) / discr.norm(b, np.inf),
+        actual, expected))
 
 
 # FIXME: Re-enable and fix up if/when standalone gradient operator exists
--- a/test/test_lazy.py
+++ b/test/test_lazy.py
@@ -30,7 +30,7 @@ from meshmode.array_context import (  # noqa
     PyOpenCLArrayContext,
     PytatoPyOpenCLArrayContext
 )
-from meshmode.dof_array import thaw
+from arraycontext.container.traversal import freeze, thaw
 
 from meshmode.array_context import (  # noqa
     pytest_generate_tests_for_pyopencl_array_context
@@ -44,7 +44,10 @@ logger = logging.getLogger(__name__)
 
 def _op_test_fixture(cl_ctx):
     queue = cl.CommandQueue(cl_ctx)
-    actx = PytatoPyOpenCLArrayContext(
+    eager_actx = PyOpenCLArrayContext(
+        queue,
+        allocator=cl_tools.MemoryPool(cl_tools.ImmediateAllocator(queue)))
+    lazy_actx = PytatoPyOpenCLArrayContext(
         queue,
         allocator=cl_tools.MemoryPool(cl_tools.ImmediateAllocator(queue)))
 
@@ -61,9 +64,9 @@ def _op_test_fixture(cl_ctx):
         })
 
     from grudge.eager import EagerDGDiscretization
-    discr = EagerDGDiscretization(actx, mesh, order=2)
+    discr = EagerDGDiscretization(eager_actx, mesh, order=2)
 
-    return actx, discr
+    return eager_actx, lazy_actx, discr
 
 
 def _rel_linf_error(discr, x, y):
@@ -129,8 +132,7 @@ def _rel_linf_error(discr, x, y):
 
 def test_lazy_op_diffusion(ctx_factory):
     cl_ctx = ctx_factory()
-    actx, discr = _op_test_fixture(cl_ctx)
-    nodes = thaw(actx, discr.nodes())
+    eager_actx, lazy_actx, discr = _op_test_fixture(cl_ctx)
 
     from grudge.dof_desc import DTAG_BOUNDARY, DISCR_TAG_BASE
     from mirgecom.diffusion import (
@@ -147,22 +149,27 @@ def test_lazy_op_diffusion(ctx_factory):
         return diffusion_operator(
             discr, DISCR_TAG_BASE, alpha, boundaries, u)
 
-    compiled_op = actx.compile(op)
+    lazy_op = lazy_actx.compile(op)
 
-    alpha = discr.zeros(actx) + 1
-    u = actx.np.cos(np.pi*nodes[0])
+    def get_inputs(actx):
+        nodes = thaw(discr.nodes(), actx)
+        alpha = discr.zeros(actx) + 1
+        u = actx.np.cos(np.pi*nodes[0])
+        return alpha, u
 
     rel_linf_error = partial(_rel_linf_error, discr)
 
-    eager_result = op(alpha, u)
-    lazy_result = compiled_op(alpha, u)
+    def lazy_to_eager(u):
+        return thaw(freeze(u, lazy_actx), eager_actx)
+
+    eager_result = op(*get_inputs(eager_actx))
+    lazy_result = lazy_to_eager(lazy_op(*get_inputs(lazy_actx)))
     assert rel_linf_error(lazy_result, eager_result) < 1e-12
 
 
 def test_lazy_op_euler(ctx_factory):
     cl_ctx = ctx_factory()
-    actx, discr = _op_test_fixture(cl_ctx)
-    nodes = thaw(actx, discr.nodes())
+    eager_actx, lazy_actx, discr = _op_test_fixture(cl_ctx)
 
     from grudge.dof_desc import DTAG_BOUNDARY
     from mirgecom.eos import IdealSingleGas
@@ -176,20 +183,27 @@ def test_lazy_op_euler(ctx_factory):
         DTAG_BOUNDARY("y"): AdiabaticSlipBoundary()
     }
 
-    def op(cv):
-        return euler_operator(discr, eos, boundaries, cv)
+    def op(state):
+        return euler_operator(discr, eos, boundaries, state)
 
-    compiled_op = actx.compile(op)
+    lazy_op = lazy_actx.compile(op)
 
-    from mirgecom.initializers import MulticomponentLump
-    init = MulticomponentLump(
-        dim=2, velocity=np.ones(2), spec_y0s=np.ones(3), spec_amplitudes=np.ones(3))
-    state = init(nodes)
+    def get_inputs(actx):
+        nodes = thaw(discr.nodes(), actx)
+        from mirgecom.initializers import MulticomponentLump
+        init = MulticomponentLump(
+            dim=2, velocity=np.ones(2), spec_y0s=np.ones(3),
+            spec_amplitudes=np.ones(3))
+        state = init(nodes)
+        return state,
 
     rel_linf_error = partial(_rel_linf_error, discr)
 
-    eager_result = op(state)
-    lazy_result = compiled_op(state)
+    def lazy_to_eager(u):
+        return thaw(freeze(u, lazy_actx), eager_actx)
+
+    eager_result = op(*get_inputs(eager_actx))
+    lazy_result = lazy_to_eager(lazy_op(*get_inputs(lazy_actx)))
     assert rel_linf_error(lazy_result, eager_result) < 1e-12
 
 
--- a/test/test_lazy.py
+++ b/test/test_lazy.py
@@ -209,7 +209,7 @@ def test_lazy_op_euler(op_test_data, order):
         nodes = thaw(discr.nodes(), actx)
         from mirgecom.initializers import MulticomponentLump
         init = MulticomponentLump(
-            dim=2, velocity=np.ones(2), spec_y0s=np.ones(3),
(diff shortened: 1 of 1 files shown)


test/test_lazy.py
```diff
@@ -0,0 +1,188 @@
...
+        return euler_operator(discr, eos, boundaries, cv)
+
+    compiled_op = actx.compile(op)
+
+    cv = make_conserved(
+        dim=2,
+        mass=discr.zeros(actx) + 1,
+        energy=discr.zeros(actx) + 1,
+        momentum=discr.zeros(actx) + np.array([1, 2]),
+        species_mass=discr.zeros(actx) + np.array([0.5, 0.25, 0.25]))
```
majosm, line 175: @MTCam Is there an easy way to make this physical? I just need to change the energy, right? Or does `euler_operator` not care?

MTCam: Euler op won't care, but if you set `energy = 2.5 + .5*np.dot(momentum, momentum)/mass` it'll get P=1.0.

test/test_lazy.py
```diff
@@ -0,0 +1,202 @@
...
+#         return _gradient_operator(
+#             discr, DISCR_TAG_BASE, alpha, boundaries, u)
+
+#     compiled_op = actx.compile(op)
+#     alpha = discr.zeros(actx) + 1
+#     u = discr.zeros(actx)
+#     compiled_op(alpha, u)
+
+
+# FIXME: Re-enable and fix up if/when standalone divergence operator exists
```
MTCam, line 104: Would this one work?
https://github.com/illinois-ceesd/mirgecom/blob/aad1a100f4cf836c3ffce2a9c3c964e9844b7d61/mirgecom/operators.py#L31

MTCam (approved):
I love this.  Would there be any value in using logger or so to time these, and report the timings on testing output?

test/test_lazy.py
```diff
@@ -0,0 +1,202 @@
...
+    compiled_op = actx.compile(op)
+
+    from mirgecom.initializers import MulticomponentLump
+    init = MulticomponentLump(
+        dim=2, velocity=np.ones(2), spec_y0s=np.ones(3), spec_amplitudes=np.ones(3))
+    state = init(nodes)
+
+    rel_linf_error = partial(_rel_linf_error, discr)
+
+    eager_result = op(state)
```
kaushikcfd, line 191: I think using the `PyOpenCLArrayContext` for the `eager_result` would probably be better. Right now, this seems to be computing the difference between an operator and its compiled counterpart. Although checking them is a good idea, but the non-compiled operator isn't really eagerly computed.

test/test_lazy.py
majosm: 👍 Added.

test/test_lazy.py
majosm: How does it look now? (Is it allowed to share a `discr` between two array contexts? What about my lazy-to-eager result conversion, does that look ok?)

majosm:
> I love this. Would there be any value in using logger or so to time these, and report the timings on testing output?

I think that would be good to have. One easy way we could do it is to add the `--durations=0` flag to the pytest command in CI, which will make it write out the timings of every test.

MTCam:
> > I love this. Would there be any value in using logger or so to time these, and report the timings on testing output?
> 
> I think that would be good to have. One easy way we could do it is to add the `--durations=0` flag to the pytest command in CI, which will make it write out the timings of every test.

Cool, let's see how it looks? #459

test/test_lazy.py
```diff
@@ -0,0 +1,231 @@
...
+    return eager_actx, lazy_actx, get_discr
+
+
+def _rel_linf_error(discr, x, y):
+    from mirgecom.fluid import ConservedVars
+    if isinstance(x, ConservedVars):
+        return _rel_linf_error(discr, x.join(), y)
+    if isinstance(y, ConservedVars):
+        return _rel_linf_error(discr, x, y.join())
+    return discr.norm(x - y, np.inf) / discr.norm(y, np.inf)
```
MTCam, line 82: @majosm do you trust this comparison for CV (or other multi-component solution vectors)?  It seems to compare them as a whole, but each component has its own equation and expected result. I'm thinking about a situation similar to this:

```
x = < .01, 20, 30, 40, 50>
y = < .06, 20, 30, 40, 50>
```

In this case, the first equation just failed to get the same result between the two methods (~83% error), but the relative error comes out to only 1e-3.

Another concern about this comparison is that it is comparing the maximum difference to the maximum value over the whole domain - meaning the error at any given point could be quite large - but if it is small compared to the domain-wide max value of the fields, it is lost.

How would you feel about replacing this with something that evaluates the maximum (relative) error in each component across the whole domain?

```
perr = x - y / y   # relative error in each field at each point
max_err = [np.max(perr[i]) for i in range(len(perr))]  # max relerr in each field
```
Of course, there would be trouble if your expected field was anywhere zero (and with your init below, y will have zeros).

Are these concerns misguided?

test/test_lazy.py
```diff
@@ -0,0 +1,231 @@
...
+        return euler_operator(discr, eos, boundaries, state)
+
+    lazy_op = lazy_actx.compile(op)
+
+    def get_inputs(actx):
+        nodes = thaw(discr.nodes(), actx)
+        from mirgecom.initializers import MulticomponentLump
+        init = MulticomponentLump(
+            dim=2, velocity=np.ones(2), spec_y0s=np.ones(3),
+            spec_amplitudes=np.ones(3))
```
MTCam, line 211: This should produce mostly zeros in the RHS.  Only the passive advected-scalar components will get non-zero RHS.  Given how the relative error function above works, I'm wondering whether it would be prudent to add tests that test equation-at-a-time, for each equation.

test/test_lazy.py
majosm: > @majosm do you trust this comparison for CV (or other multi-component solution vectors)? It seems to compare them as a whole, but each component has its own equation and expected result. I'm thinking about a situation similar to this:
> 
> ```
> x = < .01, 20, 30, 40, 50>
> y = < .06, 20, 30, 40, 50>
> ```
> 
> In this case, the first equation just failed to get the same result between the two methods (~83% error), but the relative error comes out to only 1e-3.

Agreed, and fixed (5c7ac75).

> Another concern about this comparison is that it is comparing the maximum difference to the maximum value over the whole domain - meaning the error at any given point could be quite large - but if it is small compared to the domain-wide max value of the fields, it is lost.
> 
> How would you feel about replacing this with something that evaluates the maximum (relative) error in each component across the whole domain?
> 
> ```
> perr = (x - y) / y   # relative error in each field at each point
> max_err = [np.max(perr[i]) for i in range(len(perr))]  # max relerr in each field
> ```
> 
> Of course, there would be trouble if your expected field was anywhere zero (and with your init below, y will have zeros).

Yeah, the zero/near-zero edge cases make this difficult. I don't think it's too unreasonable to just use a single global scaling factor per component here. For the most part we can pick and choose which cases to use in tests, so we can generally avoid those that have too large a range of scales.

test/test_lazy.py
majosm: Is this case OK with the per-component error? Or is there maybe a better one I should use?

MTCam: The expected RHS for this case goes something like [0, 0 , 0 , 0, a, b, c].  

I'm not saying that just because the expected result is 0, that the equation is not exercised, but I am concerned that when the expected values are zero/small that we run right up against situation that can allow errors to escape undetected; when a potentially large point-local error is diminished by the max value of the field across the entire domain to which it is compared when calculating the relative error as it is written above. Dealing with this type of issue was important for fluid solution comparisons we did in pc2compare, for example.

I can't think of a (physical) case that obviously avoids this issue for all the equations - but I think we could come up with contrived cases for each equation that would give us more confidence in the results. 

FWIW, I think we should write a centralized solution comparison utility in the *MIRGE-Com* library. It should report errors to the user on a component-by-component basis, and revert to absolute differences when the relative comparisons will yield anomalous results.  I know that may seem like overkill for this particular application - but such a utility would be pretty useful in this and many other cases where our current error evaluations are, imo, falling way short.

majosm: > The expected RHS for this case goes something like [0, 0 , 0 , 0, a, b, c].

Hm, I'm getting a nonzero RHS for all of the components for some reason:

```python
eager_components = eager_result.join()
print([discr.norm(comp) for comp in eager_components])
```

```
[9.797958971132713, 44.090815370097204, 6.9282032302755026, 6.928203230275506, 16.98354190507502, 16.98354190507502, 16.98354190507502]
```

...though I just noticed by BCs are inconsistent with me setting `velocity=np.ones(2)`, so I might have to tweak things.
 
> I can't think of a (physical) case that obviously avoids this issue for all the equations - but I think we could come up with contrived cases for each equation that would give us more confidence in the results.

We can always do multiple tests. I think we need that anyway, since we probably want to check that all of the boundary conditions work with lazy?

> FWIW, I think we should write a centralized solution comparison utility in the _MIRGE-Com_ library. It should report errors to the user on a component-by-component basis, and revert to absolute differences when the relative comparisons will yield anomalous results. I know that may seem like overkill for this particular application - but such a utility would be pretty useful in this and many other cases where our current error evaluations are, imo, falling way short.

👍

test/test_lazy.py
```diff
@@ -0,0 +1,233 @@
...
+
+def _rel_linf_error(discr, actual, expected):
+    from mirgecom.fluid import ConservedVars
+    if isinstance(actual, ConservedVars):
+        return _rel_linf_error(discr, actual.join(), expected)
+    if isinstance(expected, ConservedVars):
+        return _rel_linf_error(discr, actual, expected.join())
+    return np.max(obj_array_vectorize_n_args(
+        lambda a, b: discr.norm(a - b, np.inf) / discr.norm(b, np.inf),
+        actual, expected))
```
MTCam, line 84: Lambdas and `obj_array_vectorize_n_args` are tough for me.  I've read up on these - but please have patience if I mess up what this is doing.   What I _think_ it is doing is:

```
max_component( max_domain(local_component_error) / max_domain(local_component_value) )
```

If that's true, one of my remaining concerns is that the numerator and denominator are not spatially correlated.  Why not something like this:

```
max_component( max_domain (local_component_error / local_component_value) )
```

.. or, in words, the maximum relative error of the component?  This is a real question.  Then the only thing left to worry about is whether `local_component_value` is dinky, in which case we could address with something along the lines of:

```
max_component( max_domain(compare_values( value1, value2, tol)
```

.. with `compare_values` handling the floating point value comparison by some method we can look up - returning relative errors where appropriate and possible, otherwise a different, but representative comparison.

majosm: > Lambdas and `obj_array_vectorize_n_args` are tough for me. I've read up on these - but please have patience if I mess up what this is doing.

Lambdas are just a shorthand for functions. The above is equivalent to:

```python
    def component_error(a, b):
        discr.norm(a - b, np.inf) / discr.norm(b, np.inf)

    return np.max(obj_array_vectorize_n_args(component_error, actual, expected))
```

`obj_array_vectorize` and `obj_array_vectorize_n_args` just take a function, apply it to each component of an object array (or multiple object arrays for the `_n_args` version) and spit back the results in an object array.

So it's equivalent to something like:

```python
    def component_error(a, b):
        discr.norm(a - b, np.inf) / discr.norm(b, np.inf)

    ncomponents = len(actual)
    comp_errors = np.empty(ncomponents, dtype=object)
    for i in range(ncomponents):
        comp_errors[i] = component_error(actual[i], expected[i])

    return np.max(comp_errors)
```

> What I _think_ it is doing is:
> 
> ```
> max_component( max_domain(local_component_error) / max_domain(local_component_value) )
> ```

Yup.

> If that's true, one of my remaining concerns is that the numerator and denominator are not spatially correlated. Why not something like this:
> 
> ```
> max_component( max_domain (local_component_error / local_component_value) )
> ```
> 
> .. or, in words, the maximum relative error of the component? This is a real question.

Here I'm kind of thinking about the denominator as if it was a nondimensionalization factor, i.e., it's just something that's representative of the overall scale of that component.

> Then the only thing left to worry about is whether `local_component_value` is dinky, in which case we could address with something along the lines of:
> 
> ```
> max_component( max_domain(compare_values( value1, value2, tol)
> ```
> 
> .. with `compare_values` handling the floating point value comparison by some method we can look up - returning relative errors where appropriate and possible, otherwise a different, but representative comparison.

This sounds fine to me, I just don't know offhand how to deal with the divide-by-zero etc. issues gracefully.

MTCam: > Lambdas are just a shorthand for functions. The above is equivalent to:
> 
> `obj_array_vectorize` and `obj_array_vectorize_n_args` just take a function, apply it to each component of an object array (or multiple object arrays for the `_n_args` version) and spit back the results in an object array.
> 

Thanks for that @majosm.  Very clear. I'll get used to it :)

> Here I'm kind of thinking about the denominator as if it was a nondimensionalization factor, i.e., it's just something that's representative of the overall scale of that component.
>

I do think that makes sense, but is likely to not yield what you want when the field values vary a great deal across the domain - which is not atypical of our simulations.
   
> > .. with `compare_values` handling the floating point value comparison by some method we can look up - returning relative errors where appropriate and possible, otherwise a different, but representative comparison.
> 
> This sounds fine to me, I just don't know offhand how to deal with the divide-by-zero etc. issues gracefully.

I like this:

https://floating-point-gui.de/errors/comparison/

test/test_lazy.py
MTCam: > ```
> [9.797958971132713, 44.090815370097204, 6.9282032302755026, 6.928203230275506, 16.98354190507502, 16.98354190507502, 16.98354190507502]
> ```

Ho! 😕 😲 :concerned:

MTCam:
Btw @majosm, just in case this was not clear.  It is not my intention to hold up merging this work with my questions - these tests are terrific and I see no harm in putting them in when you are satisfied with how they are working.  It is probably appropriate for us to discuss error metrics in a more broadly scoped effort, as whatever we do there is likely to affect a lot of our existing tests and junk.

majosm:
> Btw @majosm, just in case this was not clear. It is not my intention to hold up merging this work with my questions - these tests are terrific and I see no harm in putting them in when you are satisfied with how they are working. It is probably appropriate for us to discuss error metrics in a more broadly scoped effort, as whatever we do there is likely to affect a lot of our existing tests and junk.

No worries. 🙂  I'm just waiting for @kaushikcfd to take a look when he has a chance.

kaushikcfd (approved):
LGTM, thanks!

PR #455, opened 2021-07-22
Correct the leap stepper for dt modification

MTCam:
This was some oversight in our changes to stepper to support lazy - leap needs to know when dt changes.

### DIFF
--- a/mirgecom/steppers.py
+++ b/mirgecom/steppers.py
@@ -214,7 +214,6 @@ def _advance_state_leap(rhs, timestepper, state,
         if isinstance(state, np.ndarray):
             state = thaw(freeze(state, actx), actx)
 
-
         if pre_step_callback is not None:
             state, dt = pre_step_callback(state=state,
                                           step=istep,
--- a/mirgecom/steppers.py
+++ b/mirgecom/steppers.py
@@ -218,7 +218,7 @@ def _advance_state_leap(rhs, timestepper, state,
             state, dt = pre_step_callback(state=state,
                                           step=istep,
                                           t=t, dt=dt)
-        stepper_cls.dt = dt
+            stepper_cls.dt = dt
 
         # Leap interface here is *a bit* different.
         for event in stepper_cls.run(t_end=t+dt):
@@ -230,7 +230,7 @@ def _advance_state_leap(rhs, timestepper, state,
                     state, dt = post_step_callback(state=state,
                                                    step=istep,
                                                    t=t, dt=dt)
-                stepper_cls.dt = dt
+                    stepper_cls.dt = dt
 
                 istep += 1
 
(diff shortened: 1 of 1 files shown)


MTCam:
btw, I do not understand the leap stepper very well. From this it looks like the pre-step-callback gets called every major step, but the post-step-callback is called every leap mini-step.  I'm not sure that makes sense.  Any ideas @cmikida2?

mirgecom/steppers.py
```diff
@@ -214,13 +214,11 @@ def _advance_state_leap(rhs, timestepper, state,
...
             state = thaw(freeze(state, actx), actx)
 
-        if dt < 0:
-            return istep, t, state
-
         if pre_step_callback is not None:
             state, dt = pre_step_callback(state=state,
                                           step=istep,
                                           t=t, dt=dt)
+        stepper_cls.dt = dt
```
cmikida2, line 221: I think this can/should be moved into the if-block directly above it. The timestep the stepper class has only needs to be updated if dt is changed.

mirgecom/steppers.py
```diff
@@ -232,6 +230,7 @@ def _advance_state_leap(rhs, timestepper, state,
                     state, dt = post_step_callback(state=state,
                                                    step=istep,
                                                    t=t, dt=dt)
+                stepper_cls.dt = dt
```
cmikida2, line 233: I think this can/should be moved into the if-block directly above it. The timestep the stepper class has only needs to be updated if dt is changed.

cmikida2:
The interface is a little different, but right now, the pre-callback is called once before a single timestep, and (while it may be hard to tell) the post-callback is also called once at the end of a single timestep. If you can clarify what you mean by major and mini-step, I may be able to explain further, but right now I see nothing wrong with what's here, other than the comments I added

cmikida2 (approved):

MTCam:
> The interface is a little different, but right now, the pre-callback is called once before a single timestep, and (while it may be hard to tell) the post-callback is also called once at the end of a single timestep. If you can clarify what you mean by major and mini-step, I may be able to explain further, but right now I see nothing wrong with what's here, other than the comments I added

Thanks Cory.  It just looks like the `post_step_callback` is down inside an inner `for` loop such that there would be calls to it without matching calls to the `pre_step_callback`.  That would break *MIRGE-Com* (I think).  If that's not actually the case, then no foul!

mirgecom/steppers.py
MTCam: (68e1e60)

mirgecom/steppers.py
MTCam: (68e1e60)

matthiasdiener (approved):

PR #458, opened 2021-07-23
Adds flame1d production testing to CI

MTCam:


thomasgibson (approved):
Based on the chat from today's dev meeting, I think we all agreed this is good to have! Thanks for adding the flame case.

PR #460, opened 2021-07-27
Check that species inputs in MulticomponentLump are consistent even when nspecies == 0

majosm:
Catches mistakes like the one I made here: https://github.com/illinois-ceesd/mirgecom/pull/452/commits/f91e99c4ae48e7a8a8eef0c50f08b67629647e4a

### DIFF
--- a/mirgecom/initializers.py
+++ b/mirgecom/initializers.py
@@ -479,22 +479,22 @@ class MulticomponentLump:
         if center.shape != (dim,) or velocity.shape != (dim,):
             raise ValueError(f"Expected {dim}-dimensional vector inputs.")
 
-        if nspecies > 0:
-            if spec_y0s is None:
-                spec_y0s = np.ones(shape=(nspecies,))
-            if spec_centers is None:
-                spec_centers = make_obj_array([np.zeros(shape=dim,)
-                                               for i in range(nspecies)])
-            if spec_amplitudes is None:
-                spec_amplitudes = np.ones(shape=(nspecies,))
-            if len(spec_y0s) != nspecies or\
-               len(spec_amplitudes) != nspecies or\
-                   len(spec_centers) != nspecies:
-                raise ValueError(f"Expected nspecies={nspecies} inputs.")
-            for i in range(nspecies):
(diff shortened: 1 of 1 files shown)


mirgecom/initializers.py
```diff
@@ -292,7 +292,7 @@ class Lump:
     """
 
     def __init__(
-            self, *, dim=1, nspecies=0,
+            self, *, dim=1,
```
MTCam, line 295: I don't understand this part of the diff.  When I look at your branch, the `nspecies` is there, but this diff seems to suggest it was removed?

MTCam (approved):
I believe this is better.  I just am having trouble piecing together what the behavior difference is?

mirgecom/initializers.py
majosm: Sorry, this part is unrelated to the PR title. I saw that `Lump` had an `nspecies` parameter that wasn't used or documented, so I got rid of it (https://github.com/illinois-ceesd/mirgecom/pull/460/commits/b50fc0e6faf1de3ba120c94d752353b60dad28e9).

MTCam: Laugh, I just didn't read the diff very well.  The fact that this was a whole different initialization just escaped my attention altogether.

majosm:
> I believe this is better. I just am having trouble piecing together what the behavior difference is?

It just removes the `if nspecies > 0:` around the species array initialization/checks, so that if you do like I did and pass in the species arrays but forget to also set `nspecies` (in which case `nspecies` defaults to 0), it will error.

MTCam:
> > I believe this is better. I just am having trouble piecing together what the behavior difference is?
> 
> It just removes the `if nspecies > 0:` around the species array initialization/checks, so that if you do like I did and pass in the species arrays but forget to also set `nspecies` (in which case `nspecies` defaults to 0), it will error.

I see it now. It really helped to realize the thing above and then also click the thing that hides whitespace changes.  I think my original thought was that one would be able to implicitly set nspecies by passing in an appropriately sized species array - but that would not have passed review by you I'm almost certain.  I agree with your change.

PR #462, opened 2021-07-28
Dumb down the mixture case to make it run much faster.

MTCam:
This case was taking up a giant portion of the examples section of CI.  This just makes it 2D, so that it can do more steps much faster.


MTCam:
Anybody :)

matthiasdiener (approved):

PR #465, opened 2021-07-30
Tweak lazy operator tests

majosm:
I fIxed the inconsistent BC in the multicomponent lump case (which I forgot about in #452). This caused the error computation to fail, because the RHS became all 0s in some components. So I tweaked the implementation to mimic `math.isclose`. I also added a second Euler test case that exercises the components that became 0 in the original case.

### DIFF
--- a/test/test_lazy.py
+++ b/test/test_lazy.py
@@ -190,22 +190,56 @@ def test_lazy_op_diffusion(op_test_data, order):
     assert isclose(lazy_result, eager_result)
 
 
+def _get_pulse():
+    from mirgecom.eos import IdealSingleGas
+    eos = IdealSingleGas()
+
+    from mirgecom.initializers import Uniform, AcousticPulse
+    uniform_init = Uniform(dim=2)
+    pulse_init = AcousticPulse(dim=2, center=np.zeros(2), amplitude=1.0, width=.1)
+
+    def init(nodes):
+        return pulse_init(x_vec=nodes, cv=uniform_init(nodes), eos=eos)
+
+    from meshmode.mesh import BTAG_ALL
+    from mirgecom.boundary import AdiabaticSlipBoundary
+    boundaries = {
+        BTAG_ALL: AdiabaticSlipBoundary()
+    }
+
+    return eos, init, boundaries, 1e-12
+
+
+def _get_scalar_lump():
+    from mirgecom.eos import IdealSingleGas
+    eos = IdealSingleGas()
+
+    from mirgecom.initializers import MulticomponentLump
+    init = MulticomponentLump(
+        dim=2, nspecies=3, velocity=np.ones(2), spec_y0s=np.ones(3),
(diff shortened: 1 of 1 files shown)


majosm:
I just noticed that this doesn't print out the difference when a test fails. Let me see if I can fix that.

test/test_lazy.py
```diff
@@ -73,15 +73,20 @@ def get_discr(order):
...
-    from mirgecom.fluid import ConservedVars
-    if isinstance(actual, ConservedVars):
-        return _rel_linf_error(discr, actual.join(), expected)
-    if isinstance(expected, ConservedVars):
-        return _rel_linf_error(discr, actual, expected.join())
-    return np.max(obj_array_vectorize_n_args(
-        lambda a, b: discr.norm(a - b, np.inf) / discr.norm(b, np.inf),
-        actual, expected))
+# Mimics math.isclose for state arrays
+def _isclose(discr, x, y, rel_tol=1e-9, abs_tol=0):
```
MTCam, line 77: Cool 👍

MTCam (approved):
I like this a lot.  I'm happy with it when you are.  The only thing... is 1e-12 how close we get?  Or is it closer?  I found that after bunch of steps in the integrated tests that lazy was within 5e-15 of eager - so it seems that this could be tightened up a bunch - but ... unsure.

majosm:
Alright, should be good now. It's a little clunkier, but on the other hand now we get component-by-component output when a test fails, which is kind of nice:

```python
        is_close, lhs, rhs = isclose(lazy_result, eager_result)
>       assert is_close, f"{lhs} not <= {rhs}"
E       AssertionError: [8.526512829121202e-14 5.7252918850281276e-08 8.957385944086127e-10
E          8.729542075557275e-10] not <= [1e-12 8.281250000025936e-09 4.0000000000211496e-10 1e-12]
E       assert False
```

> The only thing... is 1e-12 how close we get? Or is it closer? I found that after bunch of steps in the integrated tests that lazy was within 5e-15 of eager - so it seems that this could be tightened up a bunch - but ... unsure.

1e-12 is my default "good enough" tolerance. 🙂 It looks like for most of the tests the (absolute) errors are in the 1e-15 to 1e-14 range, but diffusion creeps up a little higher (it does hit 1e-12, for order 3).

MTCam:
> Alright, should be good now. It's a little clunkier, but on the other hand now we get component-by-component output when a test fails, which is kind of nice:

I agree. Super cool feature. Well worth a bit of clunk, imo.

> 1e-12 is my default "good enough" tolerance. 🙂 It looks like for most of the tests the (absolute) errors are in the 1e-15 to 1e-14 range, but diffusion creeps up a little higher (it does hit 1e-12, for order 3).

Mine too! I only tested the examples as-is, which I think only run at order 1. I wonder why NS and diffusion seem to have higher errors.

MTCam (approved):

PR #466, opened 2021-08-01
Add viscous module

MTCam:
This change set adds the viscous module implementing viscous fluxes and utils (such as we use them in Y1). Testing the viscous fluxes in a relevant setting also required adding the `grad_operator` utility to the `operators` module, which also appears in this PR.


mirgecom/viscous.py
```diff
@@ -0,0 +1,377 @@
...
+# enabling support for other models and simultaneous eqns.
+def viscous_flux(discr, eos, cv, grad_cv, grad_t):
+    r"""Compute the viscous flux vectors.
+
+    The viscous fluxes are:
+
+    .. math::
+
+        \mathbf{F}_V = [0,\tau\cdot\mathbf{v} - \mathbf{q},
+        \tau_{:i},-\mathbf{J}_\alpha],
```
majosm, line 220: What is `\tau_{:i}`?

mirgecom/viscous.py
```diff
@@ -0,0 +1,377 @@
...
+    grad_v = velocity_gradient(discr, cv, grad_cv)
+    div_v = np.trace(grad_v)
+
+    return mu*(grad_v + grad_v.T) + (mu_b - 2*mu/3)*div_v*np.eye(dim)
+
+
+def diffusive_flux(discr, eos, cv, grad_cv):
+    r"""Compute the species diffusive flux vector, ($\mathbf{J}_{\alpha}$).
+
+    The species diffussive flux is defined by:
```
majosm, line 95:
```suggestion
    The species diffusive flux is defined by:
```

mirgecom/simutil.py
```diff
@@ -108,23 +103,25 @@ def get_sim_timestep(discr, state, t, dt, cfl, eos,
     cfl: float
         The current CFL number
     eos: :class:`~mirgecom.eos.GasEOS`
-        Gas equation-of-state supporting speed_of_sound
+        Gas equation-of-state supporting speed_of_sound, and, optionally for viscous
+        fluids, a non-empty :class:`~mirgecom.transport.TransportModel` for viscous
+        transport properties.
```
majosm, line 108: Deriving from `GasEOS` implies that it supports `speed_of_sound` (or `sound_speed`, whatever), so it shouldn't be necessary to say this explicitly.

mirgecom/viscous.py
```diff
@@ -0,0 +1,377 @@
...
+
+def get_viscous_timestep(discr, eos, cv):
+    """Routine returns the the node-local maximum stable viscous timestep.
+
+    Parameters
+    ----------
+    discr: grudge.eager.EagerDGDiscretization
+        the discretization to use
+    eos: :class:`~mirgecom.eos.GasEOS`
+        An equation of state implementing the speed_of_sound method
```
majosm, line 333: Same here.

mirgecom/viscous.py
```diff
@@ -0,0 +1,377 @@
...
+def get_viscous_cfl(discr, eos, dt, cv):
+    """Calculate and return node-local CFL based on current state and timestep.
+
+    Parameters
+    ----------
+    discr: :class:`grudge.eager.EagerDGDiscretization`
+        the discretization to use
+    eos: :class:`~mirgecom.eos.GasEOS`
+        Implementing the pressure and temperature functions for
+        returning pressure and temperature as a function of the state q.
```
majosm, line 366: Similar here.

majosm (commented):
OK, I think this is enough for a first pass. 🙂 I'll take another look once these are out of the way (though I think I got most of it.)

mirgecom/simutil.py
```diff
@@ -202,7 +206,11 @@ def check_naninf_local(discr, dd, field):
 
 
 def compare_fluid_solutions(discr, red_state, blue_state):
```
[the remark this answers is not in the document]
MTCam: Correct.

mirgecom/viscous.py
```diff
@@ -0,0 +1,377 @@
...
+    Returns
+    -------
+    numpy.ndarray
+        The total diffusive heath flux vector
+    """
+    nspec = cv.nspecies
+    if nspec > 0:
+        try:
+            h_alpha = eos.get_species_enthalpies(cv)
+            return sum(h_alpha[i]*j[i] for i in range(nspec))
```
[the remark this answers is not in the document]
majosm: Whoops, I missed the `_alpha` somehow. Sorry about that.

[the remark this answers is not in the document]
MTCam: Heh, no worries @majosm!

mirgecom/viscous.py
MTCam: F_V = [ 0, \tau.dot.v - q, row 1 of \tau, .. row (dim) of \tau, -J]

I see this \tau{:i} notation used in the literature, but I'll see if I can make this doc more clear.

test/test_viscous.py
```diff
@@ -0,0 +1,520 @@
...
+    mesh = generate_regular_rect_mesh(
+        a=(1.0,) * dim, b=(2.0,) * dim, n=(nel_1d,) * dim
+    )
+    order = 1
+    discr = EagerDGDiscretization(actx, mesh, order=order)
+    nodes = thaw(actx, discr.nodes())
+
+    actx.np.power(nodes, .5)
+    actx.np.power(nodes[0], .5)
+    actx.np.power(nodes[0][0], .5)
```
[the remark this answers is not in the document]
MTCam: Just a reminder that we have no `actx.np.power` routine, which is used by the power-law transport model.  Since then, I've been advised to try the `**` operator - just haven't gotten around to doing that as it is not a part of this change set.

mirgecom/flux.py
```diff
@@ -30,6 +31,59 @@
...
 """
 import numpy as np  # noqa
+from meshmode.dof_array import DOFArray
+from mirgecom.fluid import (
+    ConservedVars,
+    make_conserved
+)
+
+
+def central_scalar_flux(trace_pair, normal):
```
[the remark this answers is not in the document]
MTCam: This function calculates the central flux for a vector of scalars, so its name is appropriate.  The fact that we use it when calculating the gradient of a vector of scalars is independent of this actual routine and what it does.  We could just as well use a different numerical flux for calculating the grad.  In general I do not agree with the notion that we should name routines after what they are used for - but instead about what they *do* or what they *are*.

The numerical fluxes in H+W are defined in terms of the `average` and `jump` operators, which are explained for both vector and scalar quantities here:
Screen Shot 2021-08-19 at 1 31 17 PM
   
This is essentially what drives my choice of whether I name the thing a "scalar" flux or a "vector" flux.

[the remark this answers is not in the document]
majosm: > The fact that we use it when calculating the gradient of a vector of scalars is independent of this actual routine and what it does. We could just as well use a different numerical flux for calculating the grad.

I don't think this is true. The `u*normal` form of the flux pops out as the thing getting multiplied by the test function inside the surface integral when deriving the DG scheme for `s = grad(u)`.

> In general I do not agree with the notion that we should name routines after what they are used for - but instead about what they _do_ or what they _are_.

I don't know how to make sense of a flux without reference to some operator/equation/physics...

> The numerical fluxes in H+W are defined in terms of the `average` and `jump` operators, which are explained for both vector and scalar quantities here:
> Screen Shot 2021-08-19 at 1 31 17 PM
> 
> This is essentially what drives my choice of whether I name the thing a "scalar" flux or a "vector" flux.

The numerical fluxes are defined by taking the term in the surface integral and modifying it with averages, jumps, etc. in order to couple the elements together.

mirgecom/viscous.py
```diff
@@ -0,0 +1,377 @@
...
+    j: numpy.ndarray
+        The species diffusive flux vector
+
+    Returns
+    -------
+    numpy.ndarray
+        The total diffusive heath flux vector
+    """
+    nspec = cv.nspecies
+    if nspec > 0:
```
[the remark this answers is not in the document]
MTCam: I don't _think_.

mirgecom/viscous.py
```diff
@@ -0,0 +1,377 @@
...
+        Gradient of the fluid temperature
+
+    Returns
+    -------
+    :class:`~mirgecom.fluid.ConservedVars`
+        The viscous transport flux vector
+    """
+    transport = eos.transport_model()
+    if transport is None:
+        return 0
```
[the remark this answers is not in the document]
MTCam: The intent is that if there are no viscous transport properties provided, then the viscous fluxes are zero.

mirgecom/viscous.py
```diff
@@ -0,0 +1,377 @@
...
+    viscous_mass_flux = 0 * cv.momentum
+
+    nspecies = cv.nspecies
+    if nspecies > 0:
+        j = diffusive_flux(discr, eos, cv, grad_cv)
+        heat_flux_diffusive = diffusive_heat_flux(discr, eos, cv, j)
+    else:
+        # passes the right (empty) shape for diffusive flux when no species
+        j = cv.momentum * cv.species_mass.reshape(-1, 1)
+        heat_flux_diffusive = 0
```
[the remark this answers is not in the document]
MTCam: Maybe, I'll be trying some things out to address this.

mirgecom/operators.py
```diff
@@ -1,5 +1,6 @@
 r""":mod:`mirgecom.operators` provides helper functions for composing DG operators.
 
+.. autofunction:: grad_operator
```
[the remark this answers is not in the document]
MTCam: So added.

mirgecom/flux.py
```diff
@@ -30,6 +31,59 @@
...
+
+    .. math::
+
+        h(\mathbf{u}^-, \mathbf{u}^+; \mathbf{n}) = \frac{1}{2}
+        \left(\mathbf{u}^{+}+\mathbf{u}^{-}\right)\hat{n}
+
+    where $\mathbf{u}^-, \mathbf{u}^+$, are the vector of independent scalar
+    components and scalar solution components on the interior and exterior of the
+    face on which the central flux is to be calculated, and $\hat{n}$ is the normal
+    vector.
```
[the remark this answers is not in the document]
MTCam: See what you think of the new docs.

test/test_operators.py
```diff
@@ -0,0 +1,283 @@
...
+        for i in range(dim):
+            dp[i] = _coord_test_func(actx, x_vec, order=2, fac=(i+1), grad=True)
+        return make_conserved(dim=dim, mass=dm, energy=de, momentum=np.stack(dp),
+                              species_mass=dy)
+    else:
+        mass = zeros + 1.0
+        energy = _trig_test_func(actx, x_vec)
+        for i in range(dim):
+            momentum[i] = _coord_test_func(actx, x_vec, order=2, fac=(i+1))
+        return make_conserved(dim=dim, mass=mass, energy=energy, momentum=momentum)
```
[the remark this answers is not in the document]
MTCam: The symbolic stuff is confusing to me.  I'm very slowly trying to grok it.

mirgecom/eos.py
```diff
@@ -129,6 +130,10 @@ def get_internal_energy(self, temperature, *, mass, species_mass_fractions):
         """Get the fluid internal energy from temperature and mass."""
         raise NotImplementedError()
 
+    def get_species_enthalpies(self, cv: ConservedVars):
+        """Get specific enthalpies for each mixture species."""
+        raise NotImplementedError()
+
```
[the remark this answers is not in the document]
MTCam: Oh, I was hoping it would return a NotImplemented error for non-mixture EOS. That way I have a specific exception to catch.  It is OK for the diffusive heat flux routine to be called with a non-mixture EOS, but as I have it this exception is the way that the routine knows to return an alternate structure.

mirgecom/viscous.py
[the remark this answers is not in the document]
MTCam: Ok, this is ready for you to give it another peak.

test/test_viscous.py
```diff
@@ -0,0 +1,520 @@
...
+        ones = x / x
+        u_x = dpdx*y*(h - y)/(2*mu)
+        p_x = p_hi - dpdx*x
+        mass = rho*ones
+        u_y = 0*x
+        velocity = make_obj_array([u_x, u_y])
+        ke = .5*np.dot(velocity, velocity)*mass
+        rho_e = p_x/(gamma-1) + ke
+        return make_conserved(2, mass=mass, energy=rho_e,
+                              momentum=mass*velocity)
```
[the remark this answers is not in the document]
MTCam: I made one that works for my tests and will work for the Poiseuille example in NS.   How's it looking for your purposes?

mirgecom/eos.py
```diff
@@ -79,6 +79,7 @@ class GasEOS:
     .. automethod:: gamma
     .. automethod:: transport_model
     .. automethod:: get_internal_energy
+    .. automethod:: get_species_enthalpies
```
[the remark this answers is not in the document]
MTCam: For at least one of them, it was because `internal_energy` was already taken, and the `get_internal_energy` version is an inversion routine with a different interface.  But mostly it was to match the underlying interface that the EOS is calling.

mirgecom/viscous.py
MTCam: Heh, well... see what you think of it now.

test/test_operators.py
```diff
@@ -0,0 +1,283 @@
...
+    tol = 1e-10 if dim < 3 else 1e-9
+    from pytools.convergence import EOCRecorder
+    eoc = EOCRecorder()
+
+    for nfac in [1, 2, 4]:
+
+        npts_axis = (nfac*4,)*dim
+        box_ll = (0,)*dim
+        box_ur = (1,)*dim
+        mesh = _get_box_mesh(dim, a=box_ll, b=box_ur, n=npts_axis)
```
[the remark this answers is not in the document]
MTCam: Ok, I changed these to use the actual h_max from the grid. FWIW, when I was looking over your operator tests to get a handle on the symbolic stuff - it looked like you were doing the exact same scaling for your grids as I was doing here.

test/test_viscous.py
```diff
@@ -0,0 +1,520 @@
...
+        bnd_tpair = TracePair(btag, interior=cv_bnd, exterior=cv_bnd)
+        flux_weak = central_scalar_flux(bnd_tpair, bnd_nhat)
+        return discr.project(bnd_tpair.dd, "all_faces", flux_weak)
+
+    for nfac in [1, 2, 4]:
+
+        npts_axis = nfac*(11, 21)
+        box_ll = (left_boundary_location, ybottom)
+        box_ur = (right_boundary_location, ytop)
+        mesh = _get_box_mesh(2, a=box_ll, b=box_ur, n=npts_axis)
```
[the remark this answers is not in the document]
MTCam: Switched to h_max.

test/test_operators.py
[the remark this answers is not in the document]
majosm: Oh, I was just suggesting here to use `nels_axis` instead of `npts_axis` (and also `nelement_per_axis` inside `_get_box_mesh`), so that it matches up with the spacing specified in `eoc.add_data_point(...)`. (If scaling by points instead of elements, I think it would need to be `1/(n-1)` instead of `1/n` in order to be consistent.)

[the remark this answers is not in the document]
MTCam: Ah, duh, ok.  As usual I went down a rabbit hole with this. Think it's ok now that it uses `h_max` routine?

mirgecom/eos.py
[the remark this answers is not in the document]
majosm: I dunno... I guess I'm not a huge fan of allowing partially-implemented interfaces. I think it's easier to reason about interfaces if a given class either fully implements the interface or doesn't implement it at all. One alternate way to do what you're currently doing here would be to create, say, a `MixtureEOS` base class that inherits from `GasEOS` and adds the mixture-specific stuff (and have `PyrometheusMixture` inherit from it, but `IdealSingleGas` not). Then you could check for `isinstance(eos, MixtureEOS)` in `diffusive_heat_flux` and return accordingly.

mirgecom/viscous.py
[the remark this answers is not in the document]
majosm: Ah ok. I think this is fine then, now that the transport model language has been weakened in the docstring. Maybe also consider returning a zero `ConservedVars` to match the documentation?

mirgecom/eos.py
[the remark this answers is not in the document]
MTCam: I like your suggested approach, and it occurred to me to do it this way but I didn't think it would fly in review. I'll toy with it for a minute.

mirgecom/viscous.py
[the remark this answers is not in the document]
majosm: What happens if
```python
    if len(d) == 0:
        return cv.momentum * cv.species_mass.reshape(-1, 1)
```
is left out of `diffusive_flux`? Does anything break?

mirgecom/viscous.py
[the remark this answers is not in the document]
MTCam: Please see what you think of this (cfe216d), and if you still think it would be better to construct a ConservedVars object full of zeros, then I'll switch it up.

test/test_operators.py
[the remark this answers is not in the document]
majosm: Not sure. @thomasgibson?

mirgecom/flux.py
```diff
@@ -30,6 +31,59 @@
...
+
+def central_scalar_flux(u_tpair, normal):
+    r"""Compute a central scalar flux.
+
+    The central scalar flux, $h$, of a scalar quantity $u$ is calculated as:
+
+    .. math::
+
+        h({u}^-, {u}^+; \mathbf{n}) = \frac{1}{2}
+        \left({u}^{+}+{u}^{-}\right)\hat{n}
```
majosm, line 49:
```suggestion
        \left({u}^{+}+{u}^{-}\right)\mathbf{\hat{n}}
```

mirgecom/flux.py
```diff
@@ -30,6 +31,59 @@
...
+
+    where ${u}^-, {u}^+$, are the scalar function values on the interior
+    and exterior of the face on which the central flux is to be calculated, and
+    $\hat{n}$ is the *normal* vector.
```
majosm, line 53:
```suggestion
    $\mathbf{\hat{n}}$ is the *normal* vector.
```

mirgecom/flux.py
```diff
@@ -30,6 +31,59 @@
...
+from mirgecom.fluid import (
+    ConservedVars,
+    make_conserved
+)
+
+
+def central_scalar_flux(u_tpair, normal):
+    r"""Compute a central scalar flux.
+
+    The central scalar flux, $h$, of a scalar quantity $u$ is calculated as:
```
majosm, line 44:
```suggestion
    The central scalar flux, $\mathbf{h}$, of a scalar quantity $u$ is calculated as:
```

mirgecom/viscous.py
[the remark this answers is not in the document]
MTCam: Yes.  The problem is that even for nspecies = 0, the cv.species_mass still has a data structure, albeit an empty one.  When I create a CV with vector quantities, the empty data structure used for the `species_mass` component needs to jibe with the other components.

mirgecom/flux.py
```diff
@@ -30,6 +31,59 @@
...
+
+    .. math::
+
+        h({u}^-, {u}^+; \mathbf{n}) = \frac{1}{2}
```
majosm, line 48:
```suggestion
        \mathbf{h}({u}^-, {u}^+; \mathbf{n}) = \frac{1}{2}
```

mirgecom/initializers.py
```diff
@@ -850,3 +851,116 @@ def __call__(self, x_vec, eos, **kwargs):
...
+        height: float
+            Height of the domain, (default = .02)
+        length: float
+            Length of the domain, (default = .1)
+        density: float
+            Constant density of the fluid, (default=1.0)
+        """
+        self.length = length
+        self.height = height
+        self.dpdx = (p_hi - p_low)/length
```
majosm, line 907: Reverse `hi` and `low`, I think? (And adjust the expressions in the other methods accordingly.)

mirgecom/initializers.py
```diff
@@ -850,3 +851,116 @@ def __call__(self, x_vec, eos, **kwargs):
...
+    P(x) &= P_{\text{hi}} + P'x\\
+    v_x &= \frac{-P'}{2\mu}y(h-y), v_y = 0\\
+    \rho &= \rho_0\\
+    \rho{E} &= \frac{P(x)}{(\gamma-1)} + \frac{\rho}{2}(\mathbf{v}\cdot\mathbf{v})
+    $$
+
+    Here, $P'$ is the constant slope of the linear pressure gradient from the inlet
+    to the outlet and is calculated as:
+    $$
+    P' = \frac{(P_{\text{hi}}-P_{\text{low}})}{\text{length}}
```
majosm, line 880:
```suggestion
    P' = \frac{(P_{\text{low}}-P_{\text{hi}})}{\text{length}}
```

majosm (commented):
Couple of doc/Poiseuille things. (Cool diagram BTW. 👍)

mirgecom/viscous.py
majosm: I think this looks OK now. Missing a definition for `q` though.

mirgecom/viscous.py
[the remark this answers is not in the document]
majosm: Isn't `-cv.mass*d.reshape(-1, 1)*grad_y` the right shape though, even when `nspecies` is 0?

mirgecom/eos.py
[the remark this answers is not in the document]
majosm: Whoah, I totally missed that we had both `internal_energy` and `get_internal_energy`. I shouldn't have let that pass review. 🙂 I guess it's not _so_ bad since they take different arguments, but there's a probably a more descriptive name we could use for `get_internal_energy`.

Could we remove the `get_` prefixes for the new methods here?

mirgecom/eos.py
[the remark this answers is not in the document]
MTCam: Taking a pass at doing the MixtureEOS base class in #484

mirgecom/viscous.py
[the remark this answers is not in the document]
MTCam: Apparently. I removed this in #484 as suggested here, and no tests failed.  Just running it through CI there to make sure.

mirgecom/initializers.py
MTCam: Well, the way I have initializer, the self.dpdx is just the magnitude. The sign is accounted for in `P(x) = P_hi - self.dpdx*x` expression.

MTCam: (4fb140c) 🤞

mirgecom/eos.py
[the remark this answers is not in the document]
MTCam: (64cb2e2) :)

mirgecom/viscous.py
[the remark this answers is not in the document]
MTCam: (64cb2e2)

mirgecom/eos.py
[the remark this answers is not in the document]
MTCam: (fabd833) and #485

mirgecom/viscous.py
```diff
@@ -0,0 +1,372 @@
...
+    transport = eos.transport_model()
+    if transport is None:
+        return 0
+
+    dim = cv.dim
+    viscous_mass_flux = 0 * cv.momentum
+
+    #     nspecies = cv.nspecies
+    # if nspecies > 0:
+
```
majosm, line 257:
```suggestion
```

mirgecom/viscous.py
```diff
@@ -0,0 +1,372 @@
...
+    # todo: user-supplied flux routine
+    f_int = viscous_flux(discr, eos, cv_tpair.int, grad_cv_tpair.int,
+                         grad_t_tpair.int)
+    f_ext = viscous_flux(discr, eos, cv_tpair.ext, grad_cv_tpair.ext,
+                         grad_t_tpair.ext)
+    # This bit hard-codes BR1 central flux, use mirgecom.flux.central?
+    f_avg = 0.5*(f_int + f_ext)
+    flux_weak = make_conserved(cv_tpair.int.dim, q=(f_avg.join() @ normal))
+
+    if local is False:
```
majosm, line 316:
```suggestion
    if not local:
```

mirgecom/viscous.py
```diff
@@ -0,0 +1,372 @@
...
+        on the faces
+    grad_cv_tpair: :class:`grudge.trace_pair.TracePair`
+        Trace pair of :class:`~mirgecom.fluid.ConservedVars` with the gradient of the
+        fluid solution on the faces
+    grad_t_tpair: :class:`grudge.trace_pair.TracePair`
+        Trace pair of temperature gradient on the faces.
+    local: bool
+        Indicates whether to skip projection of fluxes to "all_faces" or not. If
+        set to *False* (the default), the returned fluxes are projected to
+        "all_faces."  If set to *True*, the returned fluxes are not projected to
```
majosm, line 295:
```suggestion
        "all_faces".  If set to *True*, the returned fluxes are not projected to
```

test/test_viscous.py
```diff
@@ -0,0 +1,471 @@
...
+logger = logging.getLogger(__name__)
+
+
+# TODO: Bring back transport_model 0 when *actx.np.power* is fixed
+@pytest.mark.parametrize("transport_model", [1])
+def test_viscous_stress_tensor(actx_factory, transport_model):
+    """Test tau data structure and values against exact."""
+    actx = actx_factory()
+    dim = 3
+    nel_1d = 5
```
majosm, line 62:
```suggestion
    nel_1d = 4
```

test/test_viscous.py
```diff
@@ -0,0 +1,471 @@
...
+
+    from meshmode.mesh.generation import generate_regular_rect_mesh
+    mesh = generate_regular_rect_mesh(
+        a=(1.0,) * dim, b=(2.0,) * dim, n=(nel_1d,) * dim
```
majosm, line 66:
```suggestion
        a=(1.0,) * dim, b=(2.0,) * dim, nelements_per_axis=(nel_1d,) * dim
```

test/test_viscous.py
```diff
@@ -0,0 +1,471 @@
...
+
+# Box grid generator widget lifted from @majosm and slightly bent
+def _get_box_mesh(dim, a, b, n, t=None):
+    dim_names = ["x", "y", "z"]
+    bttf = {}
+    for i in range(dim):
+        bttf["-"+str(i+1)] = ["-"+dim_names[i]]
+        bttf["+"+str(i+1)] = ["+"+dim_names[i]]
+    from meshmode.mesh.generation import generate_regular_rect_mesh as gen
+    return gen(a=a, b=b, n=n, boundary_tag_to_face=bttf, mesh_type=t)
```
majosm, line 120:
```suggestion
    return gen(a=a, b=b, npoints_per_axis=n, boundary_tag_to_face=bttf, mesh_type=t)
```

majosm (commented):
Spotted a few more minor things.

mirgecom/flux.py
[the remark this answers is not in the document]
MTCam: I think I see the disconnect here. I'm loosely calling anything that appears as a surface integral over the element (i.e. as a result of divergence theorem, integration by parts, or similar action) as a "flux". Such flux terms can be constructed for arbitrary fields as things in-and-of-themselves at the element boundary - and are different from the transport fluxes.

Additionally, there is the reconstruction method for the quantities that we intend to integrate over the element surface, which are needed because the fields are not (in-general) single-valued at the element faces.  I am (currently) naming these routines according to the reconstruction method and the input field - not the intended use of the constructed quantity.  Describing fluxes according to the reconstruction method is a common practice in the literature, as well. 

We have 3 different numerical flux routines used in the fluid code right now.  

- One that implements what BR1 calls *H_s* in their eqn 13 - for the aux eqn.  I'm OK with calling that `grad_flux` if you want, but to me that name is less descriptive of what the routine is doing. They prescribe a central reconstruction of a scalar field there (i.e. treating each component of *u* as an independent scalar field).
- One that implements BR1's *h_e* in their eqn 10, which is the flux term resulting from applying the divergence theorem/IBP to the volume integral of the divergence of the inviscid part of the transport flux.  They suggest several, we use a LLF numerical flux there.
- One that implements BR1's *h_v* in their eqn 15, which is the flux term resulting from applying the divergence theorem/IBP to the volume integral of the divergence of the viscous part of the transport flux.  They prescribe a central reconstruction for the viscous flux vector, and then dot that with the face normals to form the "flux" that is to be integrated over the element surface.

If we rename, for example, the `central_vector_flux` to `divergence_flux`, then what to do with the LLF routine?  It too is a flux intended for use in a divergence integral. This change would suggest we should call both of these the `divergence_flux` and then pass in the reconstruction method as either central, or LLF, or other.  But this will be complicated by the presence of penalty terms in some reconstructions and not others.     

I am not opposed to reconstructing things like that - but I do suggest that we should take that flux refactoring on in a dedicated effort.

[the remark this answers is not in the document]
majosm: > I think I see the disconnect here. I'm loosely calling anything that appears as a surface integral over the element (i.e. as a result of divergence theorem, integration by parts, or similar action) as a "flux".

Me too. 🙂 

> Such flux terms can be constructed for arbitrary fields as things in-and-of-themselves at the element boundary - and are different from the transport fluxes.

Sure, but they're still induced by the operator being discretized, so I think it makes sense to tie them to the `grad`/`div` operators. Off the top of my head I can't picture a way that these particular flux forms would arise otherwise.

> Additionally, there is the reconstruction method for the quantities that we intend to integrate over the element surface, which are needed because the fields are not (in-general) single-valued at the element faces. I am (currently) naming these routines according to the reconstruction method and the input field - not the intended use of the constructed quantity. Describing fluxes according to the reconstruction method is a common practice in the literature, as well.
> 
> We have 3 different numerical flux routines used in the fluid code right now.
> 
> * One that implements what BR1 calls _H_s_ in their eqn 13 - for the aux eqn.  I'm OK with calling that `grad_flux` if you want, but to me that name is less descriptive of what the routine is doing. They prescribe a central reconstruction of a scalar field there (i.e. treating each component of _u_ as an independent scalar field).
> * One that implements BR1's _h_e_ in their eqn 10, which is the flux term resulting from applying the divergence theorem/IBP to the volume integral of the divergence of the inviscid part of the transport flux.  They suggest several, we use a LLF numerical flux there.
> * One that implements BR1's _h_v_ in their eqn 15, which is the flux term resulting from applying the divergence theorem/IBP to the volume integral of the divergence of the viscous part of the transport flux.  They prescribe a central reconstruction for the viscous flux vector, and then dot that with the face normals to form the "flux" that is to be integrated over the element surface.

Note that all of the BR1 names refer to the operator to which they correspond. I'm fine with `central_grad_flux` if that's your objection (it's unclear to me right now whether we could make a general `grad_flux` function that takes a `central_flux` function (say) as a parameter, or if maybe it would work the other way around; we only have one gradient flux at the moment, though, so I'm not really giving it a lot of thought). It's the `scalar`/`vector` terminology that I don't like; it doesn't tell me anything about what the flux represents.

> If we rename, for example, the `central_vector_flux` to `divergence_flux`, then what to do with the LLF routine? It too is a flux intended for use in a divergence integral. This change would suggest we should call both of these the `divergence_flux` and then pass in the reconstruction method as either central, or LLF, or other. But this will be complicated by the presence of penalty terms in some reconstructions and not others.
> 
> I am not opposed to reconstructing things like that - but I do suggest that we should take that flux refactoring on in a dedicated effort.

I just noticed that `lfr_flux` in the code is something different from what I thought. In the book it's just:

Screen Shot 2021-08-20 at 4 00 11 PM

which is (in principle at least) independent of the operator. But the current implementation explicitly encodes the flux for a divergence. I would advocate for either:

1. Renaming `lfr_flux` to something like `lfr_div_flux` and putting it in `operators.py` (along with renaming `central_vector_flux` to `central_div_flux` and sticking it in there too).
2. Passing `f` in as a function parameter, and leaving it as `lfr_flux` in `flux.py`.

(But this is probably a job for a separate PR. Let's limit the scope to `central_scalar_flux` here.)

mirgecom/simutil.py
majosm: (This one is still here BTW.)

MTCam: Yikes!

mirgecom/flux.py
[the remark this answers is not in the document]
MTCam: > > Such flux terms can be constructed for arbitrary fields as things in-and-of-themselves at the element boundary - and are different from the transport fluxes.
> 
> Sure, but they're still induced by the operator being discretized, so I think it makes sense to tie them to the `grad`/`div` operators. Off the top of my head I can't picture a way that these particular flux forms would arise otherwise.

Fair enough. I agree that it does make sense to tie them to the operators, and I am happy to rename them as such.

> Note that all of the BR1 names refer to the operator to which they correspond. 

I don't see that. Can you point it out?

> I'm fine with `central_grad_flux` if that's your objection (it's unclear to me right now whether we could make a general `grad_flux` function that takes a `central_flux` function (say) as a parameter, or if maybe it would work the other way around; we only have one gradient flux at the moment, though, so I'm not really giving it a lot of thought). 

I guess I was thinking we could have a `gradient_flux` and then pass in the reconstruction method such as: 

```python
def central_reconstruction(u_tpair):
    return u_tpair.avg
```

Clearly, the central reconstruction is trivial esp with the TracePair support, but more complicated reconstruction routines can easily be imagined. In the case of transport flux reconstructions there are clearly many varied choices.   

> It's the `scalar`/`vector` terminology that I don't like; it doesn't tell me anything about what the flux represents.

Absolutely true of course. Again, the name was designed to say what it was doing, not what it would be used for.

What I ended up doing was renaming both - since this is the easiest point at which to do so. We use them more extensively in the soon-to-follow-on developments. I hope they are making more sense from your perspective now.   (2c2c6e0)

test/test_operators.py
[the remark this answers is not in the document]
MTCam: Any chance you'd let this slide into a dedicated effort to replace this (and other) testing fixtures with symbolic constructs?  I really can't make heads or tails of it when I read it.

test/test_operators.py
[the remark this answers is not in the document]
MTCam: After discussing with @thomasgibson, I think we agree that the current EOC tests should be made consistent with other wrt how they scale the grids and what grid/spacing metric is used in the EOC.  (#486)

test/test_viscous.py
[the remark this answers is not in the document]
MTCam: (#486)

MTCam:
Sorry about this @majosm:  I noticed that y1-production had @jlevine18 mixture-diffusion-limited timestep work that needed to be a part of this.  We've all had eyes on this before.

mirgecom/flux.py
```diff
@@ -3,7 +3,8 @@
 Numerical Flux Routines
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-.. autofunction:: lfr_flux
+.. autofunction:: gradient_flux_central
+.. autofunction:: divergence_flux_lfr
```
thomasgibson, line 7: `divergence_flux_lfr` sounds strange. Technically none of these flux functions should care (in principle) whether you're computing fluxes for the primary or auxiliary equations. I know why you'd want to distinguish between the primary and auxiliary equations in DG, but I feel like it may be better to just leave `flux_lfr` unchanged for now. Thoughts?

mirgecom/viscous.py
```diff
@@ -0,0 +1,419 @@
...
+    normal = thaw(actx, discr.normal(cv_tpair.dd))
+
+    # todo: user-supplied flux routine
+    f_int = viscous_flux(discr, eos, cv_tpair.int, grad_cv_tpair.int,
+                         grad_t_tpair.int)
+    f_ext = viscous_flux(discr, eos, cv_tpair.ext, grad_cv_tpair.ext,
+                         grad_t_tpair.ext)
+    # This bit hard-codes BR1 central flux, use mirgecom.flux.central?
+    f_avg = 0.5*(f_int + f_ext)
+    flux_weak = make_conserved(cv_tpair.int.dim, q=(f_avg.join() @ normal))
```
thomasgibson, line 314: Shouldn't we use the `gradient_flux_central` routine here?

test/test_operators.py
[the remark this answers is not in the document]
thomasgibson: This is the reason I think we should consider moving _away_ from the symbolic stuff in mirgecom (not add more of it).

mirgecom/flux.py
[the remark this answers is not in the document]
majosm: > > Note that all of the BR1 names refer to the operator to which they correspond.
> 
> I don't see that. Can you point it out?

The subscripts: `_s` for S (aka `grad(u)`), `_e` for Euler, `_v` for viscous.

> What I ended up doing was renaming both - since this is the easiest point at which to do so. We use them more extensively in the soon-to-follow-on developments. I hope they are making more sense from your perspective now. ([2c2c6e0](https://github.com/illinois-ceesd/mirgecom/commit/2c2c6e0393474d71c34de8efe18f334a2e184b02))

I think this looks pretty good for now. Eventually I think it would be good to have these in the same module as `div_operator`/`grad_operator` and name them similar to the other numerical fluxes (i.e., call them `div_facial_flux`/`grad_facial_flux`). And then we would just put the operator-independent stuff (like the generalized LFR function and `central_reconstruction`, etc.) in `flux.py`. But we can work on that later.

mirgecom/flux.py
majosm: Not sure I follow... Note that the current LFR implementation here encodes a divergence with the `@ normal`; it's not quite the same as the book.

test/test_operators.py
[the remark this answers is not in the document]
majosm: I dunno, I found the symbolic capabilities super helpful for avoiding messy algebra while writing the diffusion tests. IMO not hard to use for this once you understand the gist. Happy to give a quick tutorial @MTCam if you want.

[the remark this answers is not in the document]
majosm: I re-implemented `_coord_test_func` symbolically as an example: #490. Want to take a stab at `_trig_test_func`? After that we can make some further simplifications (like computing gradients and evaluating expressions inside the test instead of inside each function).

mirgecom/flux.py
MTCam: I share your perspective on this @thomasgibson, I think.  My take is that regardless of what the operator or set of equations is, one will be either integrating a vector or a scalar at the element boundary (i.e. either `a*n_hat` or `v.dot.nhat`), and it made sense to me to name them accordingly.  The different element flux routines seem to just come down to how the `a` or `v` get reconstructed before scaling, or dotting (respectively) `n_hat`).

FWIW, I am just fine leaving this how it is. It doesn't break anything, it isn't wrong,  and we can talk at length about how to name them.

mirgecom/viscous.py
MTCam: Of course not.  The operand here is a vector so we need to use `divergence_flux_central`.

mirgecom/viscous.py
```diff
@@ -0,0 +1,419 @@
...
+            # spatial locality, causing fewer cache misses
+            return arraycontext.make_loopy_program(
+                "{ [i1,i0,i2]: 0<=i1<ni1 and 0<=i0<ni0 and 0<=i2<n_species}",
+                "out[i1,i0] = max(i2, a[i2,i1,i0])"
+            )
+
+        return_dof.append(
+            actx.call_loopy(make_max_kernel(), a=stacked_diffusivity)["out"])
+
+    return DOFArray(actx, tuple(return_dof))
```
majosm, line 419: #492

MTCam: I am really confused why we don't need a kernel here?

mirgecom/viscous.py
```diff
@@ -0,0 +1,419 @@
...
+    d_alpha: numpy.ndarray
+        Species diffusivities
+    """
+    if len(d_alpha) <= 0:
+        return 0
+    if not isinstance(d_alpha[0], DOFArray):
+        return max(d_alpha)
+
+    return_dof = []
+    for i in range(len(d_alpha[0])):
```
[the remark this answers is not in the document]
majosm: (Superseded by #492; I forgot to remove this comment.)

mirgecom/viscous.py
majosm: The kernel is inside `actx.np.maximum`. It computes the elementwise maximum between two arrays. `reduce` calls it successively on each species component, i.e.:

```python
max01 = actx.np.maximum(d_alpha[0], d_alpha[1])
max012 = actx.np.maximum(max01, d_alpha[2])
max0123 = actx.np.maximum(max012, d_alpha[3])
...etc...
```

mirgecom/viscous.py
MTCam: (8a6bb2d)

mirgecom/viscous.py
MTCam: Tricksies. 👍

test/test_operators.py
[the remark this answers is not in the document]
MTCam: How's this? (#493)

mirgecom/flux.py
MTCam: > Not sure I follow... Note that the current LFR implementation here encodes a divergence with the `@ normal`; it's not quite the same as the book.

In my H+W, in the Beyond One Dimension chapter, where they implement a 2D Euler solver, they lay out the Lax-Friedrichs as this:

`{{Flux}}.dot.n_hat + lam[[CV]]/2`

This is also how it is implemented in their example code in both the 1D and the 2D example, so I duplicated.

majosm: > In my H+W, in the Beyond One Dimension chapter, where they implement a 2D Euler solver, they lay out the Lax-Friedrichs as this:
> 
> `{{Flux}}.dot.n_hat + lam[[CV]]/2`
> 
> This is also how it is implemented in their example code in both the 1D and the 2D example, so I duplicated.

Sorry, I should have been more specific; I meant the general form that the book describes (the one I screenshotted in the other discussion thread). The version used for Euler is just that applied to the divergence operator. I'm guessing @thomasgibson is thinking about the more general version here (?). If we had a function that computed the general form, it would make sense to call it `lfr_flux`, but since it's currently specific to divergence I think `divergence_flux_lfr` is appropriate.

MTCam: I am not sure.  I think the one you screenshotted was a 1D version? It is extended to multiple dimensions as you read on in the same chapter - wrapping up in section 2.3 with eqns 2.16,2.17 and the following definition of the Lax-Friedrichs (the one we currently implement).

majosm: > Edit: I think I see now, Matt. You mean I'm returning (f*.dot.n, as appropriate for the divergence operator) and not just f*?

Yep, exactly.

MTCam: #494 🤔

majosm: > #494 🤔

Neat. Yeah, I like this. It would be cool to eventually generalize this, adding a `flux_central`, etc. Then we could have a single flux function for each operator and pass these guys in as parameters.

MTCam: Trouble is that it does not work as intended.  I have not been able to piece together what is going wrong yet.  I think this part at least could be safely pushed off to a dedicated flux-refactoring effort.  Whadda you think?

majosm: > Trouble is that it does not work as intended. I have not been able to piece together what is going wrong yet. I think this part at least could be safely pushed off to a dedicated flux-refactoring effort. Whadda you think?

https://github.com/illinois-ceesd/mirgecom/pull/494#discussion_r694926647

MTCam: > > Trouble is that it does not work as intended. I have not been able to piece together what is going wrong yet. I think this part at least could be safely pushed off to a dedicated flux-refactoring effort. Whadda you think?
> 
> [#494 (comment)](https://github.com/illinois-ceesd/mirgecom/pull/494#discussion_r694926647)

Awesome!  You fixed it! This exact issue has been plaguing me for some time.  It is why the LFR flux routine is how it ~is~ was.  Otherwise, I would have applied the .dot.normal up inside `inviscid_facial_flux` - but I could not make that work at the time because I couldn't figure out how to form f^* in a way that worked.

MTCam: @majosm The tests no longer work right after the LFR change. I am not sure what is wrong here, PyOpenCL claims to run out of memory?!

majosm: Yeah. It looks like it's only failing for `test_multilump_rhs` for order 4 in 3D (which has a largest mesh size of 16 elements per axis). My guess is that it's either running out of memory due to a bunch of temporary arrays floating around or the memory is just getting too fragmented. I'm trying a CI run with a memory pool allocator to see if fragmentation is the issue: #496

test/test_viscous.py
```diff
@@ -0,0 +1,521 @@
...
+    ytop = .02
+    nspecies = 0
+    spec_diffusivity = 0 * np.ones(nspecies)
+    transport_model = SimpleTransport(viscosity=mu, thermal_conductivity=kappa,
+                                      species_diffusivity=spec_diffusivity)
+
+    xlen = right_boundary_location - left_boundary_location
+    p_low = base_pressure
+    p_hi = pressure_ratio*base_pressure
+    dpdx = (p_hi - p_low) / xlen
```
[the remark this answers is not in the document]
MTCam: (437af0b)

mirgecom/flux.py
```diff
@@ -3,7 +3,10 @@
 Numerical Flux Routines
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-.. autofunction:: lfr_flux
+.. autofunction:: gradient_flux_central
+.. autofunction:: divergence_flux_central
+.. autofunction:: flux_lfr
+.. autofunction:: divergence_flux_lfr
```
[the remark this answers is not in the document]
MTCam: #502

mirgecom/flux.py
```diff
@@ -30,9 +33,93 @@
...
+        Trace pair for the face upon which flux calculation is to be performed
+    normal: numpy.ndarray
+        object array of :class:`meshmode.dof_array.DOFArray` with outward-pointing
+        normals
+
+    Returns
+    -------
+    numpy.ndarray
+        object array of `meshmode.dof_array.DOFArray` with the central scalar flux
+        for each scalar component.
```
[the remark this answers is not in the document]
MTCam: (bb9087b)

mirgecom/flux.py
```diff
@@ -30,9 +33,93 @@
...
```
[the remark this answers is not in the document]
MTCam: (bb9087b)

mirgecom/flux.py
```diff
@@ -30,9 +33,94 @@
...
+
+    Returns
+    -------
+    numpy.ndarray
+        object array of :class:`~meshmode.dof_array.DOFArray` with the flux for each
+        scalar component.
+    """
+    tp_avg = u_tpair.avg
+    tp_join = tp_avg
+    if isinstance(tp_avg, DOFArray):
```
inducer, line 79: Leave a `FIXME` here to make sure this gets improved once generalized `outer` arrives. cc @majosm

mirgecom/flux.py
```diff
@@ -62,15 +150,61 @@ def lfr_flux(cv_tpair, f_tpair, normal, lam):
...
+        Physical flux trace pair on faces on which numerical flux is to be calculated
+
+    normal: numpy.ndarray
+
+        object array of :class:`~meshmode.dof_array.DOFArray` with outward-pointing
+        normals
+
+    lam: :class:`~meshmode.dof_array.DOFArray`
 
         lambda parameter for Lax-Friedrichs/Rusanov flux
```
inducer, line 199: `lam` is a local/global upper bound on the wave speed. The docs should say so, here and elsewhere.

mirgecom/flux.py
```diff
@@ -30,9 +33,94 @@
...
+        object array of :class:`~meshmode.dof_array.DOFArray` with outward-pointing
+        normals
+
+    Returns
+    -------
+    numpy.ndarray
+        object array of :class:`~meshmode.dof_array.DOFArray` with the flux for each
+        scalar component.
+    """
+    return trace_pair.avg@normal
```
inducer, line 120: I wonder whether this routine is worth having (vs inlining the implementation), given the balance between code and documentation.

mirgecom/initializers.py
```diff
@@ -850,3 +851,116 @@ def __call__(self, x_vec, eos, **kwargs):
...
+        eos: :class:`~mirgecom.eos.GasEOS`
+            A gas equation of state
+        cv: :class:`~mirgecom.fluid.ConservedVars`
+            Optional fluid state to supply fluid density and velocity if needed.
+
+        Returns
+        -------
+        :class:`~mirgecom.fluid.ConservedVars`
+            The Poiseuille solution state
+        """
```
inducer, line 929: Does this work in 2D and 3D? If it's 2D-only, should it yell if it's being handed a 3D array? If it's 2D or 3D, maybe let the docs say so?

mirgecom/initializers.py
```diff
@@ -850,3 +851,116 @@ def __call__(self, x_vec, eos, **kwargs):
...
+            mass = self.rho*x/x
+            u_x = -self.dpdx*y*(self.height - y)/(2*self.mu)
+            velocity = make_obj_array([u_x, 0*x])
+
+        ke = .5*np.dot(velocity, velocity)*mass
+        rho_e = p_x/(eos.gamma(cv)-1) + ke
+        return make_conserved(2, mass=mass, energy=rho_e,
+                              momentum=mass*velocity)
+
+    def exact_grad(self, x_vec, eos, cv_exact):
```
inducer, line 947: The symbolic test infrastructure could probably be used to compute this. (Make an issue/add a FIXME?)

mirgecom/flux.py
```diff
@@ -62,15 +150,61 @@ def lfr_flux(cv_tpair, f_tpair, normal, lam):
...
+    -------
+    numpy.ndarray
+
+        object array of :class:`~meshmode.dof_array.DOFArray` with the
+        Lax-Friedrichs/Rusanov flux.
+    """
+    return flux_lfr(cv_tpair, f_tpair, normal, lam)@normal
+
+
+def flux_lfr(cv_tpair, f_tpair, normal, lam):
```
inducer, line 167: I don't find it super meaningful (and potentially confusing) to offer this functionality of computing the "un-dotted-with-a-normal" flux. Was there a use case? A quick grep suggested it's only used just above here.

mirgecom/flux.py
```diff
@@ -30,9 +33,94 @@
...
+    Returns
+    -------
+    numpy.ndarray
+        object array of :class:`~meshmode.dof_array.DOFArray` with the flux for each
+        scalar component.
+    """
+    return trace_pair.avg@normal
+
+
+def divergence_flux_lfr(cv_tpair, f_tpair, normal, lam):
```
inducer, line 123: `divergence_flux_*` rubs me the wrong way somehow. LF is something that applies to the whole conservation law, and the divergence is only a small part of that. `claw_flux_lfr`?

mirgecom/simutil.py
```diff
@@ -177,7 +176,11 @@ def write_visfile(discr, io_fields, visualizer, vizname,
 
 
 def allsync(local_values, comm=None, op=None):
-    """Perform allreduce if MPI comm is provided."""
+    """Perform allreduce if MPI comm is provided.
+
+    .. note::
+        This is a collective routine and must be called by all MPI ranks.
+    """
```
inducer, line 183: Document `op`?

mirgecom/simutil.py
```diff
@@ -177,7 +176,11 @@ def write_visfile(discr, io_fields, visualizer, vizname,
 
 
 def allsync(local_values, comm=None, op=None):
```
inducer, line 178: Kind of not loving the function name? Make an issue to rename to `allreduce`?

mirgecom/viscous.py
```diff
@@ -0,0 +1,402 @@
...
+.. autofunction:: conductive_heat_flux
+.. autofunction:: diffusive_heat_flux
+.. autofunction:: viscous_facial_flux
+
+Viscous Time Step Computation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. autofunction:: get_viscous_timestep
+.. autofunction:: get_viscous_cfl
+.. autofunction:: get_local_max_species_diffusivity
```
inducer, line 18: `get` vs not? (for consistency with the above)

mirgecom/viscous.py
```diff
@@ -0,0 +1,402 @@
...
+
+    Returns
+    -------
+    numpy.ndarray
+        The total diffusive heat flux vector
+    """
+    if isinstance(eos, MixtureEOS):
+        h_alpha = eos.species_enthalpies(cv)
+        return sum(h_alpha.reshape(-1, 1) * j)
+    return 0
```
inducer, line 206: This feels like an abstraction fail? Should there be an abstract method that helps avoid the conditional?

mirgecom/viscous.py
```diff
@@ -0,0 +1,402 @@
...
+
+
+# TODO: We could easily make this more general (dv, grad_dv)
+# so that user can define arbitrary array containers for *dv*,
+# enabling support for other models and simultaneous eqns.
+def viscous_flux(discr, eos, cv, grad_cv, grad_t):
```
inducer, line 212: Is this comment really true? At least `species_diffusivity` seems to need to know "who it's talking to".

inducer (commented):
Thanks! This is looking good. A few thoughts below, but I saw nothing critical IIRC.

mirgecom/eos.py
```diff
@@ -137,6 +143,40 @@ def dependent_vars(self, cv: ConservedVars) -> EOSDependentVars:
...
+        """Get the species specific enthalpies."""
+        raise NotImplementedError()
+
+    def get_production_rates(self, cv: ConservedVars):
+        """Get the production rate for each species."""
+        raise NotImplementedError()
+
+    def get_species_source_terms(self, cv: ConservedVars):
+        r"""Get the species mass source terms to be used on the RHS for chemistry."""
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
MTCam: #485

mirgecom/simutil.py
MTCam: When we rolled this, we preferred `allsync` over `allreduce` is already an MPI routine.

mirgecom/viscous.py
MTCam: #485

mirgecom/flux.py
MTCam: I agree, but we've already beat this to death on this PR.  Can we defer to #502?

mirgecom/viscous.py
MTCam: > Is this comment really true? At least `species_diffusivity` seems to need to know "who it's talking to".

Isn't species diffusivity just a function of (CV, grad(CV))? My only thought here, and perhaps it is misguided, is that if we had a DV array container (like CV), we could use to send around 
DV, and the gradient of DV.  Regardless, I will remove the comment.

mirgecom/flux.py
MTCam: IMO, it is important to explicitly delineate this as function with supporting documentation so that: a) the user/developer can easily read what it is and does, and b) so that users can *eventually* swap out for alternate implementations.

mirgecom/flux.py
MTCam: The reason I let it be a passed-in parameter is because the LFR reads that lambda = the max eigenvalue of the system flux jacobian. For single gas/fluid, it is the max wave speed, but not for other systems and maybe not even for our system with species diffusion.  

See if this added documentation clears it up. (2ec3cc0)

mirgecom/eos.py
```diff
@@ -137,6 +143,40 @@ def dependent_vars(self, cv: ConservedVars) -> EOSDependentVars:
...
+    .. automethod:: get_density
+    .. automethod:: get_species_molecular_weights
+    .. automethod:: get_production_rates
+    .. automethod:: species_enthalpies
+    .. automethod:: get_species_source_terms
+    """
+
+    def get_density(self, pressure, temperature, species_mass_fractions):
+        """Get the density from pressure, temperature, and species fractions (Y)."""
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
MTCam: Changing this here is grossly out-of-scope for this PR isn't it?  Probably should have added a follow-on PR for this - but I didn't (182bb45).

mirgecom/initializers.py
MTCam: It *is* named *PlanarPoiseuille*, with a snazzy diagram an everything.  Heh, regardless I tried to sharpen it up a bit in (4c9f5ee).

mirgecom/simutil.py
MTCam: (c08bbbe)

mirgecom/viscous.py
MTCam: Not sure.  We could detect whether we have a mixture at the caller level - but I think that the code is cleaner with it down here. In general, it is perfect valid for non-mixture EOS to be used in viscous simulations - but non-mixture EOS should have 0 for species enthalpies.  

I'm unsure how I would address this other than something very similar to this.

mirgecom/flux.py
majosm: Leaving it un-dotted lets you pass it in as a parameter to the operator flux routine, e.g.:

```python
flux = inviscid_facial_flux(discr, eos, cv_tpair, flux_func=flux_lfr)
```

If the dot was already applied, I think you would need to define and pass in different functions depending on what is being done on the inside of `inviscid_facial_flux`. Not sure if this applies so much to LFR, but for, say, `flux_central`, you'd need separate versions for `div` and `grad`.

None of this is implemented yet, so this is just an attempt to look ahead a little. Some of what's here will probably end up going away when #502 is implemented (like the separate `divergence_flux_lfr` and `divergence_flux_central`; they can be unified).

I think we could use the `*_facial_flux` convention to distinguish between the two kinds of fluxes we have here (this is already being done in `inviscid.py` and `viscous.py`). I guess really `divergence_flux_*` should be `divergence_facial_flux_*`... but maybe we can save that change for #502. 🤷‍♂️

mirgecom/flux.py
majosm: I'm inclined to leave it for now. I think the implementation will grow somewhat with the `flux_func` parametrization stuff in #502. It would also maybe be nice to add an optional projection step as in, e.g., `inviscid_facial_flux`.

mirgecom/flux.py
MTCam: From my perspective integrating *any* quantity over the surface of the element in the surface integral term is going to involve dotting with the facial normals. That surface integral is called a "flux" - and it is separate from the "transport fluxes", or any other quantity that we intend to integrate. I think the different choices come down to how you reconstruct the given quantity at the surface (e.g. LFR, central difference, Roe state, etc). The rest, i.e. dotting or scaling the normal just comes down to whether you are integrating a scalar, or a vector.   Is this a misguided view?

majosm: > From my perspective integrating _any_ quantity over the surface of the element in the surface integral term is going to involve dotting with the facial normals.

I don't know if you meant to specifically imply "dotting" to mean a dot product -- but some kind of product, I would imagine, yeah.

> That surface integral is called a "flux" - and it is separate from the "transport fluxes", or any other quantity that we intend to integrate.

If we consider the "facial fluxes" in the code to be the things that appear inside the surface integrals, then we can treat them as generalizations of the transport fluxes (w.r.t. a surface). In the case of a conservation law (divergence) they're the same, and in the case of something else like `s = grad(u)`, it generalizes to something new that isn't a `dot(f, n)`. To me `inviscid_facial_flux`, `divergence_flux`, `gradient_flux`, etc., are all computing the same type of thing, just for different operators.

> I think the different choices come down to how you reconstruct the given quantity at the surface (e.g. LFR, central difference, Roe state, etc). The rest, i.e. dotting or scaling the normal just comes down to whether you are integrating a scalar, or a vector. Is this a misguided view?

I think it comes down to what the form of the integral looks like when you convert the volume integral into a surface integral. For a divergence it becomes `dot(u, n)`, for a gradient it becomes `u*n`. I'm not sure what it would look like for, say, a curl. But I don't think it could be either of these.

MTCam: > > That surface integral is called a "flux" - and it is separate from the "transport fluxes", or any other quantity that we intend to integrate.
> 
> If we consider the "facial fluxes" in the code to be the things that appear inside the surface integrals, then we can treat them as generalizations of the transport fluxes (w.r.t. a surface).

I mean in general the surface integral "of some quantity" over the surface of the element (or any closed volume) is, in general, called a flux integral. If you are integrating a vector quantity, the normal component (i.e. that part of the vector sticking in the nhat direction) is obtained by dotting with the normal.

> In the case of a conservation law (divergence) they're the same, and in the case of something else like `s = grad(u)`, it generalizes to something new that isn't a `dot(f, n)`. 

Right, in this case we integrate a *scalar* quantity over the surface. Any scalar we want to compute the "element flux" integral for will integrate the same.  This is why I made the two routines a "scalar" surface integral vs. a "vector" surface integral.

> 
> > I think the different choices come down to how you reconstruct the given quantity at the surface (e.g. LFR, central difference, Roe state, etc). The rest, i.e. dotting or scaling the normal just comes down to whether you are integrating a scalar, or a vector. Is this a misguided view?
> 
> I think it comes down to what the form of the integral looks like when you convert the volume integral into a surface integral. For a divergence it becomes `dot(u, n)`, for a gradient it becomes `u*n`. I'm not sure what it would look like for, say, a curl. But I don't think it could be either of these.

In the case of "curl" operator, the result is a *vector* at the surface and computing it on both sides (+,-) of the face and reconstructing it due to the discontinuity at the face is one part, then for the surface integral it would be dotted with the normal.

majosm: > In the case of "curl" operator, the result is a _vector_ at the surface and computing it on both sides (+,-) of the face and reconstructing it due to the discontinuity at the face is one part, then for the surface integral it would be dotted with the normal.

This doesn't sound right to me (unless I'm misunderstanding). If you subsequently converted that back into a volume integral, you'd have a divergence.

MTCam: > > In the case of "curl" operator, the result is a _vector_ at the surface and computing it on both sides (+,-) of the face and reconstructing it due to the discontinuity at the face is one part, then for the surface integral it would be dotted with the normal.
> 
> This doesn't sound right to me (unless I'm misunderstanding). If you subsequently converted that back into a volume integral, you'd have a divergence.

I could be wrong here, but I'm viewing this in the context of like a generalized stokes:

volume_integral(d[f]) = surface_integral([f] dA)

If f is a scalar, then d[f] is a gradient
If f is a one-form vector then d[f] is a divergence
If f is a two-form vector then d[f] is a curl

In the surface integral it is always [f]dA.  So that's why I keep saying that in the end it is how the quantity [f] gets reconstructed that is different across systems and use-cases, but the surface integral is always done the same regardless of the physical or numerical content of [f].

FWIW, I'm not arguing against keeping them as-is or continuing along with the operator-specific naming scheme; only offering my perspective on why I named them how they previously were.

MTCam: > > In the case of "curl" operator, the result is a _vector_ at the surface and computing it on both sides (+,-) of the face and reconstructing it due to the discontinuity at the face is one part, then for the surface integral it would be dotted with the normal.
> 
> This doesn't sound right to me (unless I'm misunderstanding). If you subsequently converted that back into a volume integral, you'd have a divergence.

After trying to remind myself how the stokes generalization works, I realized that you are of course right and the way in which the surface integrand is formed is operator-specific.

mirgecom/flux.py
inducer: Sure.

mirgecom/flux.py
inducer: I'm guessing this'll get left as-is and done later via #502?

mirgecom/simutil.py
inducer: It carries out a reduction. It should have "reduction" in the name IMO. "sync" kind of sounds like synchronization primitive. It could be called `global_reduction` or something.

mirgecom/eos.py
```diff
@@ -58,7 +60,7 @@ class EOSDependentVars:
     pressure: np.ndarray
 
 
-class GasEOS:
+class GasEOS(metaclass=ABCMeta):
```
inducer, line 63: You can also just inherit from `ABC`. (Metaclasses might be confusing to folks.)

mirgecom/viscous.py
inducer: These aren't part of the EOS though.

mirgecom/viscous.py
```diff
@@ -0,0 +1,399 @@
...
+
+def diffusive_flux(discr, eos, cv, grad_cv):
+    r"""Compute the species diffusive flux vector, ($\mathbf{J}_{\alpha}$).
+
+    The species diffusive flux is defined by:
+
+    .. math::
+
+        \mathbf{J}_{\alpha} = -\rho{d}_{(\alpha)}\nabla{Y_{\alpha}}~~
+        (\mathtt{no~implied~sum}),
```
inducer, line 106:
```suggestion
        \mathbf{J}_{\alpha} = -\rho{d}_{(\alpha)}\nabla{Y_{\alpha}},
```
(`\alpha` occurs on both sides of the equation, also wouldn't the same remark be needed lots of other places for consistency? Also `\mathtt` for human-readable text is weird.)

mirgecom/viscous.py
```diff
@@ -0,0 +1,399 @@
...
+
+    .. math::
+
+        \mathbf{\tau} = \mu\left(\nabla{\mathbf{v}}
+        +\left(\nabla{\mathbf{v}}\right)^T\right) + (\mu_B - \frac{2\mu}{3})
+        \left(\nabla\cdot\mathbf{v}\right)
+
+    Parameters
+    ----------
+    discr: :class:`grudge.eager.EagerDGDiscretization`
```
inducer, line 71: This should be `dcoll` and `grudge.DiscretizationCollection`. (Here and below)

mirgecom/viscous.py
inducer: Fair enough.

mirgecom/viscous.py
```diff
@@ -0,0 +1,399 @@
...
+    grad_cv_tpair: :class:`grudge.trace_pair.TracePair`
+        Trace pair of :class:`~mirgecom.fluid.ConservedVars` with the gradient of the
+        fluid solution on the faces
+    grad_t_tpair: :class:`grudge.trace_pair.TracePair`
+        Trace pair of temperature gradient on the faces.
+    local: bool
+        Indicates whether to skip projection of fluxes to "all_faces" or not. If
+        set to *False* (the default), the returned fluxes are projected to
+        "all_faces".  If set to *True*, the returned fluxes are not projected to
+        "all_faces"; remaining instead on the boundary restriction.
```
inducer, line 294: I'd kind of prefer if this flag wasn't here. First, `local` as a name doesn't suggest what it does (at least to me, but maybe that's subjective). Second, typing out the `project` call at the call site seems clearer.

inducer (commented):
Thanks! A few comments. No showstoppers. Feel free to merge and defer everything to issues if preferred.

mirgecom/viscous.py
MTCam: Can you explain your reasoning here?  The document states Einstein summation convention is used throughout except where otherwise noted.  This is one of those wheres, and since \alpha is a repeated index here in the term on the right side, the comment is needed afaict.

mirgecom/viscous.py
MTCam: (#509)  I'll get to this quickly - just wanted to note it in an issue.

mirgecom/viscous.py
MTCam: (#509) I added this to a housecleaning list because we need to update the interfaces, too, so that the documentation matches the interface.  But, when I tried to update the interfaces straightforwardly - it did not work.  There are some supporting changes needed, out-of-scope for this change set.

mirgecom/eos.py
MTCam: Will address in #471.

mirgecom/simutil.py
MTCam: (#509) Added this to the housecleaning issue because it is used by all the drivers, and is a change unrelated to this PR.

PR #468, opened 2021-08-02
Add lazy operator tests for NS

majosm:


test/test_lazy.py
```diff
@@ -274,6 +278,107 @@ def lazy_to_eager(u):
     assert is_close, f"{lhs} not <= {rhs}"
 
 
+def _get_poiseuille():
+    p_low = 1
+    p_hi = 1.001
```
majosm, line 283: @MTCam I turned the pressure way down from where it was in the Poiseuille example, and that seemed to fix the differences I was seeing between lazy and eager. Guessing that having that value be so large in relation to everything else was amplifying the rounding error. Is setting it like this ok?

MTCam: I can't think of an immediate problem with this.  The pressure ratio I used was pretty much arbitrary.

majosm:
```
_ test_lazy_op_diffusion[<context factory for <pyopencl.Device 'pthread-Intel(R) Xeon(R) Platinum 8272CL CPU @ 2.60GHz' on 'Portable Computing Language'>>-1] _
  File "/home/runner/work/mirgecom/mirgecom/test/test_lazy.py", line 205, in test_lazy_op_diffusion
    assert is_close, f"{lhs} not <= {rhs}"
AssertionError: 496.33304448274185 not <= 5.244507934888335e-10
assert False
```

:eyes:

Hrmm... it works on my laptop. Let's try running CI again.

majosm:
@inducer I'm seeing some suspicious behavior from the `compose_index_maps` kernel inside meshmode's `InterpolationBatch._global_from_element_indices`. It seems to be giving nondeterministic results: sometimes I get the right answer, and sometimes I get all `-1`s. I'm seeing it happen both between different tests inside the same CI run and between different CI runs of the same test. Check out the CI results here: https://github.com/illinois-ceesd/mirgecom/pull/468/checks?check_run_id=3266172182. Any ideas what it could be? Or otherwise, are there any other things I can try to help find out?

inducer:
I can reproduce this locally, but I haven't yet made a great deal of progress figuring out what's wrong. The generated code is here:

https://gist.github.com/inducer/021185f1e3fdf6bf5f9dc89530a64986

One possible lead is that [this freeze](https://github.com/inducer/meshmode/blob/877b89df01f71a2fd318243fa4dfdebc84a5fddd/meshmode/discretization/connection/direct.py#L151) has to deal with a loopy kernel with a global barrier in a subkernel. The generated code looks fine (I think?), but I wonder if there are other gotchas I'm not remembering. @kaushikcfd, does anything come to mind?

kaushikcfd:
Unfortunately, I'm not able to reproduce this locally as of yet. (Ran quite a few times on the scicomp-machines). The generated code seems fine to me.

inducer:
I seem to have the most success reproducing this right after I clear all caches. (PyOpenCL, Loopy)

inducer:
I think I've made a little bit of progress. If I run this code (which matches the generated CL code) right after clearing caches on pocl:

https://gist.github.com/inducer/48ffc06b957cc5f5a813b0b9ea449206

I get
```
[-1 -1 -1 -1]
```

inducer:
Another thing that potentially incriminates pocl/LLVM is that things execute as expected if I enable `trace_assignment_value=True` in loopy.

inducer:
On Linux, I've only been able to reproduce this on my Ice Lake laptop, not on any of the machines in the machine shop. @majosm * Can you reproduce this using the pure pyopencl script?
* On what machines were you observing this?

majosm:
> * On what machines were you observing this?

Only on CI.

> * Can you reproduce this using the pure pyopencl script?

Yep. I added it as a test (ea577ca). CI run is here: https://github.com/illinois-ceesd/mirgecom/runs/3327322272.

inducer:
This seems to be a pocl bug, I've reported it here: https://github.com/pocl/pocl/issues/979

In terms of workarounds, we could (at least):
* Create this index map on the CPU and transfer it.
* Make pytato smart enough to avoid the copy.
* Somehow add a global barrier between the scatter and the copy.

Any thoughts?

majosm:
> In terms of workarounds, we could (at least):
> 
> * Create this index map on the CPU and transfer it.
> * Make pytato smart enough to avoid the copy.
> * Somehow add a global barrier between the scatter and the copy.
> 
> Any thoughts?

Option 1 seems reasonable to me at a glance (I don't know what would all involved with the other two). This computation is only done once per `InterpolationBatch` instance, right? Can we compute it on the CPU, then memoize the thawed result on the `InterpolationBatch` instance (with `actx` as a key I guess)?

inducer:
> Can we compute it on the CPU, then memoize the thawed result on the InterpolationBatch instance (with actx as a key I guess)?

This may not play well with lazy though, as it needs to copy index data to the CPU. @kaushikcfd, what do you think?

kaushikcfd:
Over here: https://github.com/kaushikcfd/ci_playground/commits/main (see commits `4df1549d` and `6ade023`) on passing -cl-opt-disable to `clBuildProgram` I'm not able to reproduce the bug. If this behavior is reproducible, we can just pass this in for this one kernel.

kaushikcfd:
> Create this index map on the CPU and transfer it.
>> Can we compute it on the CPU, then memoize the thawed result on the InterpolationBatch instance (with actx as a key I guess)?
>>>This may not play well with lazy though, as it needs to copy index data to the CPU. @kaushikcfd, what do you think?

Since the final result is frozen, I think this should be fine.

inducer:
> we can just pass this in for this one kernel.

How would we do that? Tags?

kaushikcfd:
Since this a problem almost only on the CI systems, we could fix it by:
```diff
diff --git a/test/test_lazy.py b/test/test_lazy.py
index d42486c2..c09cd180 100644
--- a/test/test_lazy.py
+++ b/test/test_lazy.py
@@ -42,6 +42,19 @@ import pytest  # noqa
 import logging
 logger = logging.getLogger(__name__)
 
+# {{{ Band-aid fix for https://github.com/pocl/pocl/issues/979
+
+# See https://github.com/illinois-ceesd/mirgecom/pull/468
+import arraycontext.loopy as actx_loopy
+import loopy as lp
+
+actx_loopy._DEFAULT_LOOPY_OPTIONS = lp.Options(
+        no_numpy=True,
+        return_dict=True,
+        build_options=["-cl-opt-disable"])
+
+# }}}
+
```

kaushikcfd:
>  Tags?

Could be done, but I think it might be better not to hurt the performance for all builds. It was almost too difficult for me to find a system to reproduce this. :)

inducer:
I'm not a fan of messing with the options. I'd much rather compute the index array on the CPU.

majosm:
Superseded by #483.

PR #469, opened 2021-08-02
Add a floating point array comparison module

jlevine18:


mirgecom/float_comparisons.py
```diff
@@ -0,0 +1,70 @@
...
+from mirgecom.fluid import ConservedVars
+from pytools.obj_array import obj_array_vectorize
+import numpy as np
+
+
+def componentwise_norm(discr, a):
+    """Calculate a component-wise infinity norm."""
+    if isinstance(a, ConservedVars):
+        return componentwise_norm(discr, a.join())
+    return obj_array_vectorize(lambda b: discr.norm(b, np.inf), a)
```
inducer, line 42: This should use array containers.

mirgecom/float_comparisons.py
```diff
@@ -0,0 +1,70 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+"""
+from mirgecom.fluid import ConservedVars
+from pytools.obj_array import obj_array_vectorize
+import numpy as np
+
+
+def componentwise_norm(discr, a):
```
inducer, line 38: Specify order of the norm.

mirgecom/float_comparisons.py
```diff
@@ -0,0 +1,70 @@
...
+        return componentwise_norm(discr, a.join())
+    return obj_array_vectorize(lambda b: discr.norm(b, np.inf), a)
+
+
+def componentwise_err(discr, lhs, rhs, relative=True):
+    """Calculate the component-wise error."""
+    lhs_norm = componentwise_norm(discr, lhs)
+    rhs_norm = componentwise_norm(discr, rhs)
+
+    err_norm = componentwise_norm(discr, rhs - lhs)
```
inducer, line 50: This is not how you measure error. The error between 1 and -1 would be 0 under your formula.

majosm: ?

jlevine18: I don't think it should- if lhs = [1], rhs = [-1], lhs norm would be 1, rhs norm should be 1, rhs - lhs = [-2] so its norm is 2, so relative error should be 2/1 = 2. I'll add a unit test to make sure, though.

jlevine18: If I'm understanding your concern correctly, the assertion added in (010265c) should check the case you brought up. I'm subtracting arrays, not subtracting norms, so it should be able to tell apart [1] and [-1].

mirgecom/float_comparisons.py
jlevine18: Added norm order as a parameter in (e35c875). Thanks for the suggestion.

mirgecom/float_comparisons.py
jlevine18: I think I switched over, but there's some subtle differences between `map_array_container` and `obj_array_vectorize` that I don't entirely understand so I'm not sure if the way I do it by (33e37cc) is the most efficient.

majosm: So, I think what @inducer is getting at is that `ConservedVars` is an array container, so in theory you should be able to replace the whole `componentwise_norm` implementation with `return map_array_container(lambda b: discr.norm(b, order), a)`. However, if you subsequently call this version on a `ConservedVars`, you'll get a `ConservedVars` back (containing the norm of each component), so I think you'll need to do an additional array container reduction inside `within_tol`. I don't have much hands-on experience with array container reductions, but it looks like there is a `rec_map_reduce_array_container` that might do the job.

jlevine18:
Oh, that's a good point. Thank you so much. There's some weird differences
that I don't completely understand, but they make `map_array_container` not
a drop-in replacement for `obj_array_vectorize`, though. I'll do some more
testing and see what I can do.

mirgecom/float_comparisons.py
jlevine18: Oh, that's a good point. Thank you so much. There's some weird differences that I don't completely understand, but they make `map_array_container` not a drop-in replacement for `obj_array_vectorize`, though. I'll do some more testing and see what I can do.

PR #473, opened 2021-08-03
enable more lazy examples

matthiasdiener:
Missing:
- sod-mpi (needs better slicing support in pytato)
- mixture-mpi (needs pyrometheus support)
- autoignition-mpi (needs pyrometheus support)

### DIFF
--- a/examples/heat-source-mpi-lazy.py
+++ b/examples/heat-source-mpi-lazy.py
@@ -1,0 +1,1 @@
+heat-source-mpi.py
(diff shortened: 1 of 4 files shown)


MTCam (changes requested):
I am all for this.  Need to figure out what to do to keep the examples from taking an hour, though. Current main runs the examples in 10 minutes. Maybe I can tweak the number of steps down or something.

MTCam (approved):

PR #476, opened 2021-08-04
profiling: fix elwise kernel exec hook

matthiasdiener:
This restores profiling of the elwise kernels.

Grudge e.g. creates temporary (?) actx copies when building face connections that will point the exec hook to an actx that isn't used throughout the actual execution.

### DIFF
--- a/mirgecom/profiling.py
+++ b/mirgecom/profiling.py
@@ -114,6 +114,8 @@ class PyOpenCLProfilingArrayContext(PyOpenCLArrayContext):
 
     def clone(self):
         """Return a semantically equivalent but distinct version of *self*."""
+        from warnings import warn
+        warn("Cloned PyOpenCLProfilingArrayContexts can not profile elementwise PyOpenCL kernels.")
         return type(self)(self.queue, self.allocator, self.logmgr)
 
     def __del__(self):
--- a/mirgecom/profiling.py
+++ b/mirgecom/profiling.py
@@ -1,4 +1,4 @@
-"""An array context with profiling capabilities."""
+"""An array context with kernel profiling capabilities."""
 
 __copyright__ = """
 Copyright (C) 2020 University of Illinois Board of Trustees
@@ -73,7 +73,7 @@ class ProfileEvent:
 
 
 class PyOpenCLProfilingArrayContext(PyOpenCLArrayContext):
-    """An array context that profiles kernel executions.
+    """An array context that profiles OpenCL kernel executions.
 
     .. automethod:: tabulate_profiling_data
     .. automethod:: call_loopy
@@ -81,6 +81,13 @@ class PyOpenCLProfilingArrayContext(PyOpenCLArrayContext):
     .. automethod:: reset_profiling_data_for_kernel
 
     Inherits from :class:`arraycontext.PyOpenCLArrayContext`.
+
+    .. note::
+
+       Profiling of :mod:`pyopencl` kernels (that is, kernels that do not get
+       called through :meth:`call_loopy`) is restricted to a single instance of
+       this class. If there are multiple instances, only the first one created
+       will be able to profile these kernels.
     """
 
     def __init__(self, queue, allocator=None, logmgr: LogManager = None) -> None:
@@ -114,8 +121,6 @@ class PyOpenCLProfilingArrayContext(PyOpenCLArrayContext):
         del self.profile_events[:]
         self.profile_results.clear()
         self.kernel_stats.clear()
-
-        cl.array.ARRAY_KERNEL_EXEC_HOOK = None
 
     def array_kernel_exec_hook(self, knl, queue, gs, ls, *actual_args, wait_for):
         """Extract data from the elementwise array kernel."""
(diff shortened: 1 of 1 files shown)


mirgecom/profiling.py
```diff
@@ -101,7 +101,9 @@ def __init__(self, queue, allocator=None, logmgr: LogManager = None) -> None:
         self.kernel_stats = {}
         self.logmgr = logmgr
 
-        cl.array.ARRAY_KERNEL_EXEC_HOOK = self.array_kernel_exec_hook
+        # Only store the first kernel exec hook for elwise kernels
+        if cl.array.ARRAY_KERNEL_EXEC_HOOK is None:
+            cl.array.ARRAY_KERNEL_EXEC_HOOK = self.array_kernel_exec_hook
```
kaushikcfd, line 106: Wouldn't this lead to incorrect stats collection for the cloned instance?

matthiasdiener: Well, we are collecting stats only for the 'main' actx instance anyway, so any CL kernels running on the clones won't be considered in any case.

kaushikcfd: I think for our use-case it's fine, but generally we don't mention anywhere that cloning results in a "secondary" arraycontext instance. Not quite sure what's the best interface for this.

Another concern is the `__del__` method would make `cl.array.ARRAY_KERNEL_EXEC_HOOK` as `None` which would lead to incorrect stats collection if one of the array contexts goes out of scope but the other is active.

matthiasdiener: > I think for our use-case it's fine, but generally we don't mention anywhere that cloning results in a "secondary" arraycontext instance. Not quite sure what's the best interface for this.

Yeah, I'm not sure if this has other impacts on e.g. lazy eval.

> Another concern is the `__del__` method would make `cl.array.ARRAY_KERNEL_EXEC_HOOK` as `None` which would lead to incorrect stats collection if one of the array contexts goes out of scope but the other is active.

I think I'll just disable that hook unsetting.

kaushikcfd: Thanks!

mirgecom/profiling.py
```diff
@@ -113,8 +122,6 @@ def __del__(self):
         self.profile_results.clear()
```
kaushikcfd, line 122: Probably a warning here, that cloned instances have spurious stats collection?

kaushikcfd (approved):
Thanks, LGTM! Just one minor request.

mirgecom/profiling.py
```diff
@@ -101,7 +108,9 @@ def __init__(self, queue, allocator=None, logmgr: LogManager = None) -> None:
         self.kernel_stats = {}
         self.logmgr = logmgr
 
-        cl.array.ARRAY_KERNEL_EXEC_HOOK = self.array_kernel_exec_hook
+        # Only store the first kernel exec hook for elwise kernels
+        if cl.array.ARRAY_KERNEL_EXEC_HOOK is None:
+            cl.array.ARRAY_KERNEL_EXEC_HOOK = self.array_kernel_exec_hook
 
     def clone(self):
         """Return a semantically equivalent but distinct version of *self*."""
```
kaushikcfd, line 116: Probably a warning here, that cloned instances have spurious stats collection?

mirgecom/profiling.py
matthiasdiener: Just as a comment, right?

kaushikcfd: > Just as a comment, right?

Meant to say warn via `warnings.warn`.

mirgecom/profiling.py
matthiasdiener: Done in https://github.com/illinois-ceesd/mirgecom/pull/476/commits/8ce79338807af61111f5933156165860a037a92d

kaushikcfd: Thanks, good to go form my end.

thomasgibson (approved):
LFTM. I'm not sure what you mean by:
>Grudge e.g. creates temporary (?) actx copies when building face connections that will point the exec hook to an actx that isn't used throughout the actual execution.

Grudge doesn't create array contexts, unless you're referring to the "setup array context" used in `DiscretizationCollection`? It technically creates a new actx (via `clone()`). The reason for this was to avoid potential conflicts when mixing array contexts that don't support the same type of arithmetic.

Meshmode (the actual package that makes the connections), does something similar with its `Discretization` class I believe.

mirgecom/profiling.py
```diff
@@ -101,10 +108,14 @@ def __init__(self, queue, allocator=None, logmgr: LogManager = None) -> None:
...
+        from warnings import warn
+        warn("Cloned PyOpenCLProfilingArrayContexts can not profile elementwise PyOpenCL kernels.")
```
thomasgibson, line 118: Pesky flake8 :)
```suggestion
        warn("Cloned PyOpenCLProfilingArrayContexts can not "
             "profile elementwise PyOpenCL kernels.")
```

matthiasdiener:
> LFTM. I'm not sure what you mean by:
> 
> > Grudge e.g. creates temporary (?) actx copies when building face connections that will point the exec hook to an actx that isn't used throughout the actual execution.
> 
> Grudge doesn't create array contexts, unless you're referring to the "setup array context" used in `DiscretizationCollection`? It technically creates a new actx (via `clone()`). 

Yes, this is the scenario I meant. The last one of these cloned actx's collected all the elwise kernel profiles, and these were essentially 'lost' since the main actx didn't ever see them.

PR #478, opened 2021-08-06
Logging arbitrary fields

sipieta2:
This PR combines #390 and #391 and adjusts all relevant examples


thomasgibson:
This PR will require https://github.com/illinois-ceesd/logpyle/pull/74

PR #483, opened 2021-08-17
Add lazy operator tests for NS (again)

majosm:
Starting fresh here to preserve the debugging stuff in #468.

### DIFF
--- a/test/test_lazy.py
+++ b/test/test_lazy.py
@@ -98,30 +98,42 @@ def _isclose(discr, x, y, rel_tol=1e-9, abs_tol=0, return_operands=False):
         return is_close
 
 
-# FIXME: Re-enable and fix up if/when standalone gradient operator exists
-# def test_lazy_op_gradient(ctx_factory):
-#     cl_ctx = ctx_factory()
-#     actx, discr = _op_test_fixture(cl_ctx)
-#
-#     from grudge.dof_desc import DTAG_BOUNDARY, DISCR_TAG_BASE
-#     from mirgecom.diffusion import (
-#         _gradient_operator,
-#         DirichletDiffusionBoundary,
-#         NeumannDiffusionBoundary)
-#
-#     boundaries = {
-#         DTAG_BOUNDARY("x"): DirichletDiffusionBoundary(0),
-#         DTAG_BOUNDARY("y"): NeumannDiffusionBoundary(0)
-#     }
-#
-#     def op(alpha, u):
-#         return _gradient_operator(
-#             discr, DISCR_TAG_BASE, alpha, boundaries, u)
+@pytest.mark.parametrize("order", [1, 2, 3])
+def test_lazy_op_gradient(op_test_data, order):
+    eager_actx, lazy_actx, get_discr = op_test_data
+    discr = get_discr(order)
 
-#     compiled_op = actx.compile(op)
-#     alpha = discr.zeros(actx) + 1
-#     u = discr.zeros(actx)
-#     compiled_op(alpha, u)
+    from grudge.trace_pair import interior_trace_pair
+    from mirgecom.operators import grad_operator
+
+    def get_flux(u_tpair):
+        dd = u_tpair.dd
+        dd_allfaces = dd.with_dtag("all_faces")
+        normal = thaw(discr.normal(dd), u_tpair.int.array_context)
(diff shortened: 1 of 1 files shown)


majosm:
Tests are pretty slow at the moment FYI, but it sounds like that should be changing soon?

MTCam (approved):
lgtm!  I guess we will just fix this up when we have the Poiseuille init routine.  Any interest in adding a lazy test for the grad operator like the div one?

majosm:
> Any interest in adding a lazy test for the grad operator like the div one?

Yep -- thanks for the reminder. Added.

MTCam:
> > Any interest in adding a lazy test for the grad operator like the div one?
> 
> Yep -- thanks for the reminder. Added.

er.. that was fast.

majosm:
> er.. that was fast.

90% copy/paste 🙂

MTCam:
FYI, per your review, I added a Poiseuille init case [here](https://github.com/illinois-ceesd/mirgecom/blob/1c920f57a755737558050ffb2c071ae084fb1cca/mirgecom/initializers.py#L856).  If it works for you, feel free to cut-paste-and-tweak it into this.

test/test_lazy.py
```diff
@@ -274,6 +290,107 @@ def lazy_to_eager(u):
...
+    isclose = partial(
+        _isclose, discr, rel_tol=tol, abs_tol=tol, return_operands=True)
+
+    def lazy_to_eager(u):
+        return thaw(freeze(u, lazy_actx), eager_actx)
+
+    eager_result = op(*get_inputs(eager_actx))
+    lazy_result = lazy_to_eager(lazy_op(*get_inputs(lazy_actx)))
+    is_close, lhs, rhs = isclose(lazy_result, eager_result)
+    assert is_close, f"{lhs} not <= {rhs}"
```
inducer, line 391: Why does this say "<="?

test/test_lazy.py
```diff
@@ -274,6 +290,107 @@ def lazy_to_eager(u):
     assert is_close, f"{lhs} not <= {rhs}"
 
 
+def _get_poiseuille():
```
inducer, line 293: Could this be done by parametrizing the 'normal' Navier-Stokes tests? (What's preventing that? Any way to avoid the blob'o'copy-paste?)

test/test_lazy.py
```diff
@@ -274,6 +290,107 @@ def lazy_to_eager(u):
...
+
+    isclose = partial(
+        _isclose, discr, rel_tol=tol, abs_tol=tol, return_operands=True)
+
+    def lazy_to_eager(u):
+        return thaw(freeze(u, lazy_actx), eager_actx)
+
+    eager_result = op(*get_inputs(eager_actx))
+    lazy_result = lazy_to_eager(lazy_op(*get_inputs(lazy_actx)))
+    is_close, lhs, rhs = isclose(lazy_result, eager_result)
```
inducer, line 390: I feel that "check that lazy and eager return the same thing" (with customizable tolerances) is probably something that can be generalized and written once rather than multiple times.

inducer (commented):

test/test_lazy.py
majosm: `lhs` and `rhs` here are the LHS/RHS of the comparison inside `_isclose`, which is:
```python
norm(a-b) <= max(rel_tol * max(norm(a), norm(b)), abs_tol)
```
(analogous to `math.isclose`).

test/test_lazy.py
majosm: I guess the intent of these tests is to compare lazy to eager, rather than lazy to exact. I suspect we can use tighter tolerances here than we can in the regular NS tests. Whether this is actually useful... I'm not sure. @kaushikcfd Any thoughts?

test/test_lazy.py
majosm: True. I can take a stab at doing that.

test/test_lazy.py
kaushikcfd: If the 'normal' NS tests take a long time to execute (for reasons like large problem size), it might be better to check the equivalence of the lazy v eager operator evaluations for a smaller problem as the PytatoArrayContextForTests results in a single-threaded operator application. Otherwise, parametrizing seems like the cleaner approach.

PR #494, opened 2021-08-24
Add generic LLF flux reconstruction

MTCam:
Add a general LFR flux reconstruction function

### DIFF
--- a/mirgecom/flux.py
+++ b/mirgecom/flux.py
@@ -204,4 +204,6 @@ def flux_lfr(cv_tpair, f_tpair, normal, lam):
         object array of :class:`meshmode.dof_array.DOFArray` with the
         Lax-Friedrichs/Rusanov flux.
     """
-    return f_tpair.avg - lam*cv_tpair.diff*normal/2
+    return make_conserved(
+        dim=len(normal),
+        q=f_tpair.avg.join() - lam*np.outer(cv_tpair.diff.join(), normal)/2)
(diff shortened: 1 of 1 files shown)


mirgecom/flux.py
```diff
@@ -159,4 +160,48 @@ def divergence_flux_lfr(cv_tpair, f_tpair, normal, lam):
...
+        lambda parameter for Lax-Friedrichs/Rusanov flux
+
+    Returns
+    -------
+    numpy.ndarray
+
+        object array of :class:`meshmode.dof_array.DOFArray` with the
+        Lax-Friedrichs/Rusanov flux.
+    """
+    return f_tpair.avg - lam*cv_tpair.diff*normal/2
```
majosm, line 207:
```suggestion
    return make_conserved(
        dim=len(normal),
        q=f_tpair.avg.join() - lam*np.outer(cv_tpair.diff.join(), normal)/2)
```
I think. (It worked for the Euler tests that I tried, at least.)

MTCam: I do not understand why the outer is needed here.  Is it because without it it will do cv.mass*nx + cv.energy*ny + ...?

majosm: If you just did `*` it would probably produce an error (since `cv_tpair.diff.join()` and `normal` are both 1D arrays, with different lengths). An `np.outer` or a `.reshape(-1, 1)` is needed here to get what you want I think.

MTCam: > If you just did `*` it would probably produce an error 

Confirmed.

I wonder why `cv_tpair.diff*normal` doesn't do what I want, however.

majosm: My guess is it's because of `cv.momentum` being a vector.

MTCam: 👍 I guessed similarly.

PR #495, opened 2021-08-24
Move symbolic evaluation inside `test_grad_operator`

majosm:
Changes all of the `*_test_func` functions to return symbolic expressions and moves the evaluation part inside the gradient test. As a result, the test functions don't need to optionally compute and return the gradient; the test itself can compute it from the symbolic expressions.

### DIFF
--- a/test/test_operators.py
+++ b/test/test_operators.py
@@ -31,7 +31,7 @@ from meshmode.array_context import (  # noqa
     pytest_generate_tests_for_pyopencl_array_context
     as pytest_generate_tests
 )
-from pytools.obj_array import make_obj_array
+from pytools.obj_array import make_obj_array, obj_array_vectorize
 import pymbolic as pmbl  # noqa
 import pymbolic.primitives as prim
 from meshmode.dof_array import thaw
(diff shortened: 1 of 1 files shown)


MTCam:
This is clearly better and more succinct, but it is harder for me to understand it.  It is going to take some getting used to.

test/test_operators.py
```diff
@@ -243,12 +198,24 @@ def test_grad_operator(actx_factory, dim, order, test_func):
...
         h_max = h_max_from_volume(discr)
 
+        def sym_eval(expr, x_vec):
+            # FIXME: When pymbolic supports array containers
+            mapper = sym.EvaluationMapper({"x": x_vec})
+            from arraycontext import rec_map_array_container
+            result = rec_map_array_container(mapper, expr)
+            # If expressions don't depend on coords (e.g., order 0), evaluated result
+            # will be scalar-valued, so promote to DOFArray(s) before returning
+            return result * (0*x_vec[0] + 1)
```
MTCam, line 208: This may be the only real difficult piece for me. It isn't apparent to me (if I were writing a new test, for example) that I need this, or how to write it if I did need it.

majosm: Yeah, this is a little ugly in its current state.

In an ideal world, it would just be:

```python
        def sym_eval(expr, x_vec):
            return sym.EvaluationMapper({"x": x_vec})(expr)
```

where `sym.EvaluationMapper` is a function-like object that takes in a symbolic expression and plugs in the corresponding data (defined by the mapping passed to its constructor, the `{"x": x_vec}` here).

The other junk is just to: 1) support evaluating expressions inside `ConservedVars` objects (because pymbolic doesn't seem to support array containers yet), and 2) make sure the result is made up of `DOFArray`s and not just constants (because something in the rest of test expects it to be; I didn't look closely to see what it was).

MTCam: Ok, this is terrific. It is much less error prone than writing those expressions by hand as I was doing - and I did not get them right the first time, or the second, ...

It would be useful to have your words (like above) pasted to the call and construct sites of this stuff.  I like reading code for documentation - especially for stuff like this.  What would be really cool is a `test_sym.py` that is basically just a demonstration of stuff especially in the context of test fixtures for MIRGE-Com.  I also wonder about just having a general `test_fixtures.py` that demos templates a bunch of patterns for writing tests.

Can this be used to build applications and have them work through lazy?  Seems like it could. Is that the sort of thing that was extricated from grudge @thomasgibson?

thomasgibson: > Is that the sort of thing that was extricated from grudge

Yes, but that infrastructure is still there! It's just deprecated and no longer used.

MTCam (approved):
👍

test/test_operators.py
majosm: > What would be really cool is a `test_sym.py` that is basically just a demonstration of stuff especially in the context of test fixtures for MIRGE-Com.

#500

PR #496, opened 2021-08-24
Try using a memory pool in the multilump test

majosm:
...to see if the allocation failure from #466 (CI run [here](https://github.com/illinois-ceesd/mirgecom/pull/466/checks?check_run_id=3413260375)) goes away.

### DIFF
--- a/test/test_euler.py
+++ b/test/test_euler.py
@@ -317,7 +317,14 @@ def test_multilump_rhs(actx_factory, dim, order, v0):
     The case is tested against the analytic expressions of the RHS. Checks several
     different orders and refinement levels to check error behavior.
     """
-    actx = actx_factory()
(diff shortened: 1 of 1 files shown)


MTCam:
It is a bit of a mystery to me that the memory footprint would change as a result of the rather slight refactor in #494

majosm:
> It is a bit of a mystery to me that the memory footprint would change as a result of the rather slight refactor in #494

I think the fact that we're doing `np.outer(..., normal) @ normal` instead of just `...` will introduce some extra temporary arrays. Probably not significant on its own, but if we were already close to the limit then it might push us over.

majosm:
Yeah, seems like it's probably just fragmentation (maybe something like #212 again?). I bet we can fix it by reducing the largest mesh size a little.

PR #506, opened 2021-09-17
Create a PR template

matthiasdiener:


MTCam:
I like this, but am a little alarmed that the lazy/eager compare is failing.  Does this indicate that lazy results are drifting away from eager?

matthiasdiener:
> I like this, but am a little alarmed that the lazy/eager compare is failing. Does this indicate that lazy results are drifting away from eager?

yeah, it started failing yesterday: https://github.com/illinois-ceesd/mirgecom/actions/runs/1243651411

inducer:
https://github.com/inducer/grudge/pull/161 merged yesterday. To get it to pass, I submitted https://github.com/illinois-ceesd/mirgecom/pull/505. Lazy is now doing different math than eager (it can take advantage of affine mappings), so I'm not too concerned about a little bit of difference. P^0 is probably more accurate anyhow.

inducer:
` 5e-15` seems a wee bit tight as tolerances go anyhow.

.github/pull_request_template.md
```diff
@@ -0,0 +1,8 @@
+- [ ] Is the scope and purpose of the PR clear?
```
inducer, line 1: Maybe this could be separated out into what the PR author should cover and what the reviewer should ask? (with some duplication maybe?)

MTCam:
> ` 5e-15` seems a wee bit tight as tolerances go anyhow.

Thanks @inducer for that explanation above.  You are right that it is a little tight, but at least it let us see and catch how some changes are requiring these bumps in tol, and put some developers eyes on and even better prompted a possible explanation of what caused it.  I guess I'm arguing for keeping the tols as tight as they can go - just to error on the side of caution.

.github/pull_request_template.md
matthiasdiener: I tried to separate them out in https://github.com/illinois-ceesd/mirgecom/pull/506/commits/d01aea41259eb4ce849f9aa4b99ce3ef9948b453, but it seems to be a lot of duplication.

MTCam (approved):

.github/pull_request_template.md
inducer: Agree, not loving the duplication.

matthiasdiener: Undid the duplication. This is ready for review @inducer.

PR #508, opened 2021-09-17
print error value

matthiasdiener:


MTCam:
Partially addresses https://github.com/annabellej/mirgecompare/issues/21

bin/mirgecompare.py
```diff
@@ -46,7 +46,7 @@ def compare_files_vtu(first_file, second_file, file_type, tolerance = 1e-12):
         # verify individual values w/in given tolerance
         for j in range(arr1.GetSize()):
             if abs(arr1.GetValue(j) - arr2.GetValue(j)) > tolerance: 
-                print("Tolerance:", tolerance)
+                print("Tolerance:", tolerance, "Error:", abs(arr1.GetValue(j) - arr2.GetValue(j)))
                 raise ValueError("Fidelity test failed: Mismatched data array values with given tolerance")
```
MTCam, line 50: This should probably compare them all and record the max difference, then raise if the tolerance was violated.  That way we can see what the max error was.

MTCam, line 50:
```suggestion
        field_errs = [abs(arr1.GetValue(j) - arr2.GetValue(j)) for j in range(arr1.GetSize())]
        max_err = max(field_errs)
        if max_err > tolerance: 
                raise ValueError(f"Fidelity test failed: Mismatched data array values with (tol,err) = ({tolerance}, {max_err})")
```

Would this work?

matthiasdiener:
Done in #710.

PR #513, opened 2021-09-23
Fix restart reading of `num_parts` in examples

majosm:
Fixes the restart reading to use the same name for the partition count when reading as when writing (`num_parts`).


MTCam (approved):
👍

PR #517, opened 2021-10-11
Use `arraycontext.outer` in flux functions

majosm:
~~:warning: Depends on inducer/arraycontext#46. :warning:~~ (Merged.)

**Questions for the review**:

### DIFF
--- a/mirgecom/flux.py
+++ b/mirgecom/flux.py
@@ -33,11 +33,6 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 import numpy as np  # noqa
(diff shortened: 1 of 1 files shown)


majosm:
@inducer Just in case you were wondering, it looks like `np.ndarray` needs to be in `bcast_container_types` for `ConservedVars` in order to support the matmuls [here](https://github.com/illinois-ceesd/mirgecom/blob/f932ff5b5616dc357fa97fed4effc5d1c6dcd51a/mirgecom/flux.py#L123) and [here](https://github.com/illinois-ceesd/mirgecom/blob/f932ff5b5616dc357fa97fed4effc5d1c6dcd51a/mirgecom/flux.py#L177). As far as I can tell from the tests these seem to be the only reason it's there.

PR #519, opened 2021-10-13
Check for `-inf` too in `check_naninf_local`

majosm:
**Questions for the review**:


majosm:
> Why not use isfinite() ?

🤷‍♂️ (Trying it.)

PR #518, opened 2021-10-11
Account for grudge nodal reductions returning device scalars

majosm:
:warning: Merge together with inducer/grudge#170. ~~Restore requirements.txt first. :warning:~~ (Done.)

**Questions for the review**:
- [ ] Is the scope and purpose of the PR clear?
  - [ ] The PR should have a description.
  - [ ] The PR should have a guide if needed (e.g., an ordering).
- [ ] Is every top-level method and class documented? Are things that should be documented actually so?
- [ ] Is the interface understandable? (I.e. can someone figure out what stuff does?) Is it well-defined?
- [ ] Does the implementation do what the docstring claims?
- [ ] Is everything that is implemented covered by tests?
- [ ] Do you see any immediate risks or performance disadvantages with the design? Example: what do interface normals attach to?

PR #523, opened 2021-10-14
Unify examples

majosm:
1. Removes serial wave example (redundant after #522).
2. Modifies `run_examples.sh` to use each example's `--help` message to determine whether it can be run with MPI/lazy.
3. Removes `<example>-lazy.py` symlinks.
4. Renames `<example>-mpi.py` to `<example>.py`.

Depends on #522, #521, and #518.

Since this renames all of the examples, it may be disruptive to any existing PRs that also modify the examples. So, depending on what's currently outstanding, I'm not sure if this can be merged right now or not.

**Questions for the review**:


majosm:
More or less superseded by #950 (and way out of date anyway). Closing.

PR #530, opened 2021-10-26
Remove explicit `actx` argument from `advance_state`

majosm:
Get the array context from the state if possible, and if it doesn't have one (such as in the time integrator tests, where the state is just a float) treat the array context operations as no-ops.

**Questions for the review**:


mirgecom/steppers.py
```diff
@@ -57,13 +64,18 @@ def get_rhs():
     return get_rhs()
 
 
+def _evaluate_state(actx, state):
```
inducer, line 67:
```suggestion
def _force_evaluation(actx, state):
```

inducer (approved):
Thanks! IMO this is an improvement.

PR #534, opened 2021-10-27
Persistent temperature guess

### LINKED ISSUE #310: Initial guess for temperature
This issue is intended to continue a conversation about how to handle the initial temperature guess that we pass into `pyro.get_temperature` in _Pyrometheus_-based EOS in _MIRGE-Com_.

Currently, if we do not have a reasonable guess, `get_temperature` takes many iterations, and can sometimes apparently get the wrong answer.  

Current approach: just pass 300K (expensive, potentially wrong)
Suggested: pass the last successful temperature (ideal, but tricky, requires restarting temperature)
Suggested: pass a deterministic guess that we compute based on current state (promising, but tricky to craft the guess)

MTCam:
This change set demonstrates a possible solution for maintaining a persistent copy of the domain temperature throughout stepping, and using it to provide a starting place for the next call to `get_temperature`.

- `steppers.py` is updated to:
  - accept and pass-through a `reference_state` to the advancers
- time integrators changed to:
  - accept optional reference state
  - if reference state is present, then pass it to RHS
- `autoignition-mpi.py` is updated to:
  - accept a reference state in the RHS
  - pass the reference state to EOS for DV computation
  - store/restart the temperature field 
- `eos.py` is updated to:
  - Accept an optional reference state in `eos.temperature`
  - Use reference state temperature for guess/seed when present, else const/uniform `tguess`

TODO:
- Fix up restart to memoize temperature on reference_state with restarted field as seed
- Port to lazy

fixes #310 

@kaushikcfd 

**Questions for the review**:


mirgecom/integrators/explicit_rk.py
```diff
@@ -28,11 +28,18 @@
 """
 
 
-def rk4_step(state, t, dt, rhs):
+def rk4_step(state, t, dt, rhs, reference_state=None):
```
inducer, line 31: It seems that this might all be avoidable by `partial`'ing the `reference_state` into `rhs` as a keyword argument. (For all the steppers.)

inducer:

MTCam:
Superseded by #524.   Requested changes from here have been addressed there.

PR #547, opened 2021-11-04
Convert results to bool in `check_range_local` and `check_naninf_local`

majosm:
So, apparently `isinstance(..., Number)` returns `False` for numpy's boolean type `np.bool_`. This is causing issues in at least one of the examples in `y1-production` where the result of one of these functions is passed into `global_reduce` (which subsequently checks if the input is a `Number`).

We could alternatively change `global_reduce` to check for this, but we already use `isinstance(..., Number)` all over the place, so changing these two return values instead seemed possibly safer?

**Questions for the review**:

### DIFF
--- a/mirgecom/simutil.py
+++ b/mirgecom/simutil.py
@@ -270,7 +270,7 @@ def allsync(local_values, comm=None, op=None):
 def check_range_local(discr, dd, field, min_value, max_value):
     """Check for any negative values."""
     actx = field.array_context
-    return (
+    return bool(  # Convert to bool because np.bool_ is not a Number
         actx.to_numpy(op.nodal_min_loc(discr, dd, field)) < min_value
         or actx.to_numpy(op.nodal_max_loc(discr, dd, field)) > max_value
     )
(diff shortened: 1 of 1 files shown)


majosm:
@inducer Any objections? Not sure if there's a more general way we should be dealing with this (e.g., using something other than `Number`?).

inducer:
I've hit a few of these, and I think a better think to do is to use `np.isscalar` instead of `isinstance(..., Number)`. It does solve this issue, too:
```
>>> np.isscalar(np.bool_(True))
True
```

majosm:
> I've hit a few of these, and I think a better thing to do is to use `np.isscalar` instead of `isinstance(..., Number)`. It does solve this issue, too:
> 
> ```
> >>> np.isscalar(np.bool_(True))
> True
> ```

After reading the documentation for [np.isscalar](https://numpy.org/doc/stable/reference/generated/numpy.isscalar.html), it sounds like we may actually want to use `np.ndim(...) == 0` in this case. Implemented in #554.

PR #554, opened 2021-11-08
Avoid using isinstance(..., Number) to detect scalars

majosm:
Supersedes #547.

**Questions for the review**:

### DIFF
--- a/mirgecom/simutil.py
+++ b/mirgecom/simutil.py
@@ -47,7 +47,6 @@ THE SOFTWARE.
 import logging
 import numpy as np
 import grudge.op as op
-from numbers import Number
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


mirgecom/simutil.py
```diff
@@ -187,8 +186,8 @@ def global_reduce(local_values, op, *, comm=None):
 
     Parameters
     ----------
-    local_values: numbers.Number or numpy.ndarray
-        The (MPI-compatible) value or array of values on which the
+    local_values:
+        The (mpi4py-compatible) value or array of values on which the
```
inducer, line 190:
```suggestion
        The (:mod:`mpi4py`-compatible) value or array of values on which the
```

mirgecom/symbolic.py
```diff
@@ -89,22 +90,22 @@ def map_call(self, expr):
             raise ValueError("Unrecognized function '%s'" % expr.function)
 
     def _sin(self, val):
```
inducer, line 92:
```suggestion
    @staticmethod
    def _np_like_for(val):
       actx = get_container_context_recursively(val)
       if actx is None:
           return  np
       else:
           return actx.np

    def _sin(self, val):
```

inducer (approved):
LGTM with these.

PR #555, opened 2021-11-11
Y1 production eos bug fix

anderson2981:
**Questions for the review**:


anderson2981:
Update get_interal_energy to return a unit-consistent value across single- and multi-species implementations. Current behavior was to return energy density (rho*e, J/m^3) in single species and energy (e J/kg) in multi-species.

The function now returns e always.

Also removed the dependence on mass.

PR #557, opened 2021-11-19
Add tests for nonlinear diffusion

majosm:
Seems like it works?

cc @dshtey2

**Questions for the review**:


majosm:
I also refactored the `HeatProblem` stuff because the case factory functions (`get_decaying_trig`, etc.) were starting to look like subclasses.

inducer (approved):
Thanks! LGTM

PR #558, opened 2021-11-25
print actx type on startup

matthiasdiener:
**Questions for the review**:

### DIFF
--- a/mirgecom/mpi.py
+++ b/mirgecom/mpi.py
@@ -121,6 +121,20 @@ def mpi_entry_point(func):
             raise RuntimeError("mpi4py.MPI imported before designated MPI entry "
                         "point. Check for prior imports.")
 
+        actx_class = kwargs.pop("actx_class", None)
+        if actx_class:
+            from arraycontext import PyOpenCLArrayContext as NewEagerActx
+            from meshmode.array_context import PyOpenCLArrayContext as OldEagerActx
+            if actx_class == OldEagerActx:
+                print(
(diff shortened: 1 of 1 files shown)


mirgecom/mpi.py
```diff
@@ -121,6 +121,20 @@ def wrapped_func(*args, **kwargs):
...
+            from meshmode.array_context import PyOpenCLArrayContext as OldEagerActx
+            if actx_class == OldEagerActx:
+                print(
+                    f"*** Eager mode with deprecated '{actx_class}. ***")
+            elif actx_class == NewEagerActx:
+                print(
+                    f"*** Eager mode with '{actx_class}. ***")
+            else:
+                print(
+                    f"*** Lazy mode with '{actx_class}. ***")
```
inducer, line 136: I think this ought to use `logging`.

mirgecom/mpi.py
```diff
@@ -121,6 +121,20 @@ def wrapped_func(*args, **kwargs):
             raise RuntimeError("mpi4py.MPI imported before designated MPI entry "
                         "point. Check for prior imports.")
 
+        actx_class = kwargs.pop("actx_class", None)
+        if actx_class:
```
inducer, line 125: I'm not sure about this being mixed in with the MPI functionality. I can see *why*, but I'm not sure it's a good idea.

matthiasdiener:
https://github.com/inducer/grudge/pull/230 implements something similar in grudge.

matthiasdiener:
Closing this since https://github.com/inducer/grudge/pull/230 has been merged.

PR #561, opened 2021-12-06
Generalize symbolic helper code so it can be used more easily with other tests

majosm:
Mostly minor changes here; the only real noteworthy change is the addition of ~~the `MathMapper` class, which is intended to be a numpy-like object~~ the math module and its `__getattr__` function that dispatches math function calls to either `np`, `np.actx`, or `pymbl.var(...)` depending on the types of the arguments.

I'm not sure if having the math mapper be a module instance is the best way to implement this, so I'm open to changing it if there's another way that's cleaner. I mainly just wanted to be able to `import mirgecom.math as mm` and then `mm.sin(...)` instead of having to `from mirgecom.math import sin`, etc. for each function used. Maybe this isn't such a big deal? I dunno. 🤷‍♂️ 

**Questions for the review**:

### DIFF
--- a/mirgecom/math.py
+++ b/mirgecom/math.py
@@ -1,6 +1,6 @@
 """Utilities and functions for math.
 
-.. autodata:: math_mapper
+.. autofunction:: __getattr__
 """
 
 __copyright__ = """Copyright (C) 2021 University of Illinois Board of Trustees"""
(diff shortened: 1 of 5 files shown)


mirgecom/math.py
```diff
@@ -0,0 +1,60 @@
...
+from pymbolic.primitives import Expression
+from arraycontext import get_container_context_recursively
+
+
+class _SymbolicMathContainer:
+    def __getattr__(self, name):
+        return pmbl.var(name)
+
+
+class MathMapper:
```
inducer, line 41:
```suggestion
class MathFunctionDispatcher:
```

mirgecom/math.py
```diff
@@ -0,0 +1,60 @@
...
+    """Dispatcher for math operations on symbolic/numeric input."""
```
inducer, line 42: Needs more:

* What's it for?
* How is it used?

mirgecom/math.py
```diff
@@ -0,0 +1,60 @@
...
+        if any(isinstance(arg, Expression) for arg in args):
+            return _SymbolicMathContainer()
+        else:
+            actx = get_container_context_recursively(make_obj_array(args))
+            if actx is not None:
+                return actx.np
+            else:
+                return np
+
+    def __getattr__(self, name):
```
inducer, line 55: Maybe use module-level `__getattr__`.

mirgecom/math.py
```diff
@@ -0,0 +1,64 @@
+"""Utilities and functions for math.
+
+.. autofunction:: __getattr__
```
inducer, line 3: Add a usage example.

inducer (approved):
Thanks! LGTM with an example added.

mirgecom/math.py
majosm: Added.

inducer (approved):
:tada: LGTM, thanks!

PR #570, opened 2021-12-12
Operator overhaul and overintegration support for all fluid-based operators

thomasgibson:
This PR add overintegration support for the main production branches (based on the new state-handling). In the process of adding overintegration to the AV and NS operators, I also did some rewriting to (hopefully) make it a bit easier to follow.

Update: since we plan to eventually add a NS-style artificial viscosity operator, I've taken this opportunity to rename `av_operator` to `av_laplacian_operator`. There is a backwards compact interface as well: https://github.com/illinois-ceesd/mirgecom/pull/570/commits/af8cccf7b088861649e02ded5623aded32dcb2aa

**Questions for the review**:

### DIFF
--- a/examples/doublemach-mpi.py
+++ b/examples/doublemach-mpi.py
@@ -130,8 +130,8 @@ def get_doublemach_mesh():
 
 @mpi_entry_point
 def main(ctx_factory=cl.create_some_context, use_logmgr=True,
-         use_leap=False, use_profiling=False, casename=None,
-         rst_filename=None, actx_class=PyOpenCLArrayContext):
+         use_leap=False, use_profiling=False, use_overintegration=False,
+         casename=None, rst_filename=None, actx_class=PyOpenCLArrayContext):
     """Drive the example."""
     cl_ctx = ctx_factory()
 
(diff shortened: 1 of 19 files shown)


examples/doublemach-mpi.py
```diff
@@ -399,12 +414,21 @@ def my_post_step(step, t, dt, state):
...
+            quadrature_tag=quadrature_tag
+        ) + av_operator(
+            discr,
+            cv=fluid_state.cv,
+            boundaries=boundaries,
+            boundary_kwargs={"time": t, "gas_model": gas_model},
+            alpha=alpha,
+            s0=s0,
+            kappa=kappa,
+            quadrature_tag=quadrature_tag
```
MTCam, line 431: I'm not wild about the splayed out style where every argument gets its own line.  Also I don't care enough to make a fuss about it.  In general, I have an aversion to having so much vertical space and so little of the available real-estate used.

MTCam (commented):
This is a great change.  So little required to enable OI.  Cool.

examples/doublemach-mpi.py
thomasgibson: Compacted it down a bit here: https://github.com/illinois-ceesd/mirgecom/pull/570/commits/78f00f488a1ca49117d2bc529e5c9b5ab955a899

PR #573, opened 2021-12-14
Flux-differencing for Euler and Navier-Stokes operators

thomasgibson:
This PR adds both Euler and Navier-Stokes operators that use flux-differencing for inviscid computations.

**Questions for the review**:


mirgecom/euler.py
```diff
@@ -193,6 +207,177 @@ def interp_to_surf_quad(utpair):
...
+        discretization to use during operator evaluations.
+        The default value is *None*.
+
+    Returns
+    -------
+    :class:`mirgecom.fluid.ConservedVars`
+
+        Agglomerated object array of DOF arrays representing the RHS of the Euler
+        flow equations.
+    """
```
inducer, line 247: Add some citations here on where the math comes from.

mirgecom/euler.py
```diff
@@ -193,6 +207,177 @@ def interp_to_surf_quad(utpair):
...
+    dd_base = as_dofdesc("vol")
+    dd_vol = DOFDesc("vol", quadrature_tag)
+    dd_faces = DOFDesc("all_faces", quadrature_tag)
+    # NOTE: For single-gas this is just a fixed scalar.
+    # However, for mixtures, gamma is a DOFArray. For now,
+    # we are re-using gamma from here and *not* recomputing
+    # after applying entropy projections. It is unclear at this
+    # time whether it's strictly necessary or if this is good enough
+    gamma = gas_model.eos.gamma(state.cv, state.temperature)
```
inducer, line 256: Add something like:
```py
if not isinstance(gas_model, eos.IdealGas):
    from warnings import warn
    warn("We're not sure about this.")
```

mirgecom/euler.py
inducer, line 247: Add a comment on your view of how mature multi-species is.

PR #577, opened 2021-12-15
Pulse: make it 3D by default

matthiasdiener:
Previously, all examples were 2D by default, thus not checking 3D
in CI e.g.

**Questions for the review**:

### DIFF
--- a/examples/pulse-mpi.py
+++ b/examples/pulse-mpi.py
@@ -36,7 +36,7 @@ from meshmode.array_context import (
     PytatoPyOpenCLArrayContext
 )
 from mirgecom.profiling import PyOpenCLProfilingArrayContext
-from meshmode.dof_array import thaw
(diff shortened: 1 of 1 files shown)


matthiasdiener:
> In principal, I am OK with this, but more is changed here than just dimensionality. There is a change in the grid spacing. We'll just have to check that the solution remains well-behaved, and that it doesn't make CI take an hour.

Agreed, from what I can see, the solution is well-behaved both for 2D and 3D, and CI time doesn't seem to be affected.

MTCam:
> Edit: Just to be clear, I changed grid spacing because 3D wasn't well-behaved with the previous grid.

Hrm, that seems strange.   Usually one would refine to fix stability issues, but in this case  it was coarsened.

MTCam (approved):
👍

PR #584, opened 2021-12-26
Force source build for mpi4py during CI

inducer:
CI is currently broken, with lots of mpi4py build failures that, potentially, are due to outdated generated Cython code. This forces a source build for mpi4py. This may be necessary in a number of other code paths, including emirge.

cc @matthiasdiener 

See https://stackoverflow.com/a/65696724 for context.


inducer:
Provisionally closing this in favor of #585.

PR #586, opened 2021-12-27
Drop Vtk from conda-env

inducer:
Vtk pulls in a lot of heavyweight dependencies. Its only purpose at this point seems to be to enable `mirgecompare`, but that doesn't justify a few extra gigabytes and a few extra minutes of install time, at least to me.

@matthiasdiener How do you feel about that? `mirgecompare` could simply have instructions to `conda install vtk` if the relevant packages are missing. Happy to add that if you agree.

Production changes to be merged alongside this:

https://github.com/illinois-ceesd/mirgecom/compare/production...production-merge/drop-vtk-from-env

### DIFF
--- a/.ci-support/merge-install-production-branch.sh
+++ b/.ci-support/merge-install-production-branch.sh
@@ -12,31 +12,67 @@ set -x
 #
 # PRODUCTION_BRANCH = The production branch (default=production)
 # PRODUCTION_FORK = The production fork (default=illinois-ceesd)
-#
+
 MIRGE_HOME=${1:-"."}
-
-PRODUCTION_BRANCH=${PRODUCTION_BRANCH:-"production"}
-PRODUCTION_FORK=${PRODUCTION_FORK:-"illinois-ceesd"}
-
 echo "MIRGE_HOME=${MIRGE_HOME}"
+
+if [[ "$GITHUB_HEAD_REF" == "$PRODUCTION_PR_MAIN_BRANCH" ]]; then
+  # GITHUB_HEAD_REF is only set for PR builds.
+  # https://docs.github.com/en/actions/learn-github-actions/environment-variables
+  PRODUCTION_BRANCH="production-merge/${PRODUCTION_PR_MAIN_BRANCH}"
+  PRODUCTION_FORK=${PRODUCTION_PR_FORK:-"illinois-ceesd"}
+else
+  PRODUCTION_BRANCH="${PRODUCTION_BRANCH:-"production"}"
+  PRODUCTION_FORK=${PRODUCTION_FORK:-"illinois-ceesd"}
+fi
+
+
 echo "PRODUCTION_FORK=$PRODUCTION_FORK"
 echo "PRODUCTION_BRANCH=$PRODUCTION_BRANCH"
 
 cd "${MIRGE_HOME}" || exit 1
 git status
 
+# $CI is set by Github Actions
+# https://docs.github.com/en/actions/learn-github-actions/environment-variables
 if [[ "$CI" == "true" ]]; then
   # This is needed in order for git to create a merge commit
   git config user.email "ci-runner@ci.machine.com"
   git config user.name "CI Runner"
 fi
 
-# Making a dedicated production remote adds production forks
-git remote add production "https://github.com/${PRODUCTION_FORK}/mirgecom"
-git fetch production
+git fetch -f \
+        "https://github.com/${PRODUCTION_FORK}/mirgecom.git" \
+        "$PRODUCTION_BRANCH":ci-prod-test
 
-# Merge the production branch for testing the production drivers
-git merge "production/${PRODUCTION_BRANCH}" --no-edit
+git fetch -f \
+        "https://github.com/illinois-ceesd/mirgecom.git" \
+        production:ci-prod-upstream
+
+if ! git merge ci-prod-test --no-edit; then
+  echo "*** The branch of this CI run failed to merge with"
+  echo "*** the current production-targeted branch:"
+  echo "*** ${PRODUCTION_BRANCH}"
+  if [[ "$PRODUCTION_BRANCH" == "production" && -n "$GITHUB_HEAD_REF" ]]; then
+    echo "*** You must create a branch off of '$GITHUB_HEAD_REF'"
+    echo "*** that is up-to-date with respect to the 'production' branch in the main"
+    echo "*** mirgecom repo and make it known to the CI in"
+    echo "*** .ci-support/production-testing-env.sh. See the comments there."
+  else
+    echo "*** You must update this branch to be up-to-date with"
+    echo "*** respect to the branch of this test run."
+  fi
+  exit 1
+fi
+
+if ! git merge ci-prod-upstream --no-edit; then
+  echo "*** The current mirgecom branch failed to merge with"
+  echo "*** mirgecom's current production branch."
+  echo "*** You must update this branch to be up-to-date with"
+  echo "*** respect to 'production' in the main mirgecom repository."
+  exit 1
+fi
+
 # Pick up any requirements.txt
 pip install -r requirements.txt
 cd - || exit 1
--- a/.ci-support/merge-install-production-branch.sh
+++ b/.ci-support/merge-install-production-branch.sh
@@ -14,6 +14,7 @@ set -x
 # PRODUCTION_FORK = The production fork (default=illinois-ceesd)
 #
 MIRGE_HOME=${1:-"."}
+
 PRODUCTION_BRANCH=${PRODUCTION_BRANCH:-"production"}
 PRODUCTION_FORK=${PRODUCTION_FORK:-"illinois-ceesd"}
 
@@ -21,19 +22,21 @@ echo "MIRGE_HOME=${MIRGE_HOME}"
 echo "PRODUCTION_FORK=$PRODUCTION_FORK"
 echo "PRODUCTION_BRANCH=$PRODUCTION_BRANCH"
 
-cd ${MIRGE_HOME}
+cd "${MIRGE_HOME}" || exit 1
 git status
 
-# This junk is needed to be able to execute git commands properly
-git config user.email "ci-runner@ci.machine.com"
-git config user.name "CI Runner"
+if [[ "$CI" == "true" ]]; then
+  # This is needed in order for git to create a merge commit
+  git config user.email "ci-runner@ci.machine.com"
+  git config user.name "CI Runner"
+fi
 
 # Making a dedicated production remote adds production forks
-git remote add production https://github.com/${PRODUCTION_FORK}/mirgecom
+git remote add production "https://github.com/${PRODUCTION_FORK}/mirgecom"
 git fetch production
 
 # Merge the production branch for testing the production drivers
-git merge production/${PRODUCTION_BRANCH} --no-edit
+git merge "production/${PRODUCTION_BRANCH}" --no-edit
 # Pick up any requirements.txt
 pip install -r requirements.txt
-cd -
+cd - || exit 1
--- a/bin/mirgecompare.py
+++ b/bin/mirgecompare.py
@@ -1,8 +1,17 @@
 #!/usr/bin/env python
+import sys
+import numpy as np
 
 
-def compare_files_vtu(first_file, second_file, file_type, tolerance = 1e-12):
-    import vtk
+class MissingVtk(Exception):
+    pass
+
+
+def compare_files_vtu(first_file, second_file, file_type, tolerance=1e-12):
+    try:
+        import vtk
+    except ImportError:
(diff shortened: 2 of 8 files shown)


inducer:
@matthiasdiener I'll wait for your feedback before sorting out the production merge failure.

matthiasdiener:
Imho, if it is only needed for mirgecompare, dropping `vtk` is a good idea.

matthiasdiener (approved):

bin/mirgecompare.py
```diff
@@ -1,8 +1,17 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
matthiasdiener, line 1: Most things installed through emirge seem to have `#!/usr/bin/env python` but I'm agnostic on this.

inducer:
@MTCam How bad would it be to merge this? AFAICT, this would only cause a trivial merge conflict in `conda-env.yml`.

bin/mirgecompare.py
inducer: Fair. Reverted for consistency.

MTCam:
> @MTCam How bad would it be to merge this? AFAICT, this would only cause a trivial merge conflict in `conda-env.yml`. (If it's bad, what's the suggested procedure for getting this in?)

I apologize for taking so long to get back with you about this.   I only just now noticed this mention.   This change set is pretty innocuous.  I'd say merge at will.

FWIW: If it did have the potential to affect the production cases, then the path for making it production-ready is outlined in the docs:  https://mirgecom.readthedocs.io/en/latest/development/pullrequests.html#production-testing-in-ci

inducer:
The main issue is that the production branch doesn't merge cleanly. I'm not sure the document you point to addresses the situation of a merge conflict and/or how to address it.

MTCam:
> The main issue is that the production branch doesn't merge cleanly. I'm not sure the document you point to addresses the situation of a merge conflict and/or how to address it.

This is going to sound like an RTFM-type answer, but that is not my intent.     

The first point I want to make before continuing; it is not strictly necessary to resolve this before merging.   Merging stuff that breaks production tests makes me grumpy, but it is not a giant deal.  If the merge is simple, and does not hurt production, then production will pick it up when it updates to pick up the new main (typically as soon as I see production tests on main start failing).   

Now for the rtfm part: Steps 1 and 2 outlined in the documentation above are specifically designed to discover and deal with merge conflicts and offer a path to resolution.

In summary of steps 1 and 2: 
- Check out the branch
- merge production branch (resolve conflicts yourself)
- push the resulting code as your "custom" production branch 
- specify the custom production branch in your production testing env for the branch

This procedure is typically only required for developments that deeply affect production.

For developments such as yours, that don't affect production at all, you can just merge with no worries and let me deal with it, or you can pre-merge your development to production and viola: merge conflicts resolved.   Not only does a pre-merge to production fix your own development's merge conflicts, but it also let's main test out production tests with your changes applied before you merge them.  These two paths to resolving merge conflicts before merge-to-main are really identical.  If you are confident that your changes don't hurt production - then pre-merge is best as it avoids production env customization.   Again, for developments that profoundly affect production code - the best path is production env customization as outlined above.

I am more than happy to help with this, and beef up the documentation if you find it lacking.

inducer:
Thanks for explaining. I'd like to propose an improvement to that process, which I've realized in 5e3ce201. The main changes are that:

- There's a naming convention for production branches, namely `production-merge/MAIN_BRANCH_NAME`.
- The production testing branch switch deactivates itself automatically once merged to main, so that it doesn't need to be backed out before merging. (And, potentially, there could be a on-push Github workflow that automatically resets production to the relevant branch.)

@matthiasdiener @MTCam LMK what you think of this. I've added it to this PR as a test, I'm happy to separate it out.

:warning: If we do merge as part of this, it please don't squash, so that unrelated changes don't get mashed into the same commit. 

L

MTCam:
> Thanks for explaining. I'd like to propose an improvement to that process, which I've realized in [5e3ce20](https://github.com/illinois-ceesd/mirgecom/commit/5e3ce201e251de56888e9fbe33b50def1ff8ee6a). The main changes are that:
> 
> * There's a naming convention for production branches, namely `production-merge/MAIN_BRANCH_NAME`.

I like this.  It'd be great if it could be automatically deleted MAIN_BRANCH_NAME merge to main.  Alas, probably too much to ask for.

> * The production testing branch switch deactivates itself automatically once merged to main, so that it doesn't need to be backed out before merging. (And, potentially, there could be a on-push Github workflow that automatically resets production to the relevant branch.)

This would be terrific, if it works!

> 
> @matthiasdiener @MTCam LMK what you think of this. I've added it to this PR as a test, I'm happy to separate it out.
> 
> ⚠️ If we do merge as part of this, it please don't squash, so that unrelated changes don't get mashed into the same commit.

inducer:
@matthiasdiener Could you look over my shell scripting to see if there's something fishy?

.ci-support/merge-install-production-branch.sh
```diff
@@ -12,28 +12,65 @@ set -x
...
+
+if [[ "$GITHUB_HEAD_REF" == "$PR_MAIN_BRANCH" ]]; then
+  # GITHUB_HEAD_REF is only set for PR builds.
+  # https://docs.github.com/en/actions/learn-github-actions/environment-variables
+  PRODUCTION_BRANCH="production-merge/${PR_MAIN_BRANCH}"
+  PRODUCTION_FORK=${PR_PRODUCTION_FORK:-"illinois-ceesd"}
+else
+  PRODUCTION_BRANCH="${PRODUCTION_BRANCH:-"production"}"
+  PRODUCTION_FORK=${PRODUCTION_FORK:-"illinois-ceesd"}
+fi
```
MTCam, line 27: This seems to indicate PRs *must* have a compatible production branch instead of allowing a default to regular ole production.   If it merges, it seems it will break all production feeder branches - none of which require such a custom production branch.

.ci-support/merge-install-production-branch.sh
```diff
@@ -12,28 +12,65 @@ set -x
...
 
-cd ${MIRGE_HOME}
+cd "${MIRGE_HOME}" || exit 1
 git status
 
-# This junk is needed to be able to execute git commands properly
-git config user.email "ci-runner@ci.machine.com"
-git config user.name "CI Runner"
+if [[ "$CI" == "true" ]]; then
+  # This is needed in order for git to create a merge commit
```
matthiasdiener, line 37: Could mention here where $CI is set

.ci-support/merge-install-production-branch.sh
```diff
@@ -12,28 +12,65 @@ set -x
...
+if ! git merge ci-prod-test --no-edit; then
+  echo "*** The branch of this CI run failed to merge with"
+  echo "*** the current production-targeted branch:"
+  echo "*** ${PRODUCTION_BRANCH}"
+  if [[ "$PRODUCTION_BRANCH" == "production" && -n "$GITHUB_HEAD_REF" ]]; then
+    echo "*** You must create a branch off of '$GITHUB_HEAD_REF'"
+    echo "*** that is up-to-date with respect to the 'production' branch in the main"
+    echo "*** mirgecom repo and make it known to the CI in"
+    echo "*** .ci-support/production-testing-env.sh. See the comments there."
+else
```
matthiasdiener, line 59:
```suggestion
  else
```

matthiasdiener:
> @matthiasdiener Could you look over my shell scripting to see if there's something fishy?

Looks good, except the two minor comments I made. I think MikeC's comment still needs addressing though.

.ci-support/merge-install-production-branch.sh
inducer: That's not the intention. What makes you think that? To me, the logic is that "if we're currently building a PR *and* if the production env settings indicate that a special branch is needed (by setting `PR_MAIN_BRANCH` to the name of the current branch... come to think of it, I'll probably rename that variable, likely `PRODUCTION_PR_MAIN_BRANCH`), then use that branch. Otherwise, fall back to the old logic.

.ci-support/merge-install-production-branch.sh
inducer: https://github.com/illinois-ceesd/mirgecom/compare/5e3ce201e251de56888e9fbe33b50def1ff8ee6a..e3725cd0d5cfb267c7deb3b0906977a0c38d5bf9

.ci-support/merge-install-production-branch.sh
inducer: https://github.com/illinois-ceesd/mirgecom/compare/5e3ce201e251de56888e9fbe33b50def1ff8ee6a..e3725cd0d5cfb267c7deb3b0906977a0c38d5bf9

.ci-support/merge-install-production-branch.sh
MTCam: The previous logic was if the PR requires a special production branch, then they set PRODUCTION_BRANCH.  I am unsure why additional variables are needed.

inducer: No, the original variables stay unchanged. The additional variable (now `PRODUCTION_PR_MAIN_BRANCH`) contains the name of the branch headed into `main`. That way, the CI script can tell whether it's currently testing that branch and, if so, test the corresponding production branch (found from `PRODUCTION_PR_MAIN_BRANCH` by convention) by appropriately populating the existing variables. Likewise, `PRODUCTION_PR_FORK` is needed because it's only supposed to be active *if* we're still on the specific branch.

MTCam: I do trust you know what you are doing; I just don't understand.   

When CI runs these scripts, it is always testing the current branch, so I didn't find it necessary to indicate to CI what branch it is currently testing.   The production testing procedure was to merge "production" into the current branch (i.e. the branch being tested, regardless of what branch it is)  and run the production cases, and any branch could specify a custom branch to use for "production" through the PRODUCTION_BRANCH variable.

It appeared that this change set removes the PRODUCTION_BRANCH variable hooks from `production_testing_env.sh`, and force the setting of PRODUCTION_PR_MAIN_BRANCH.   The previous logic would use `production` proper iff the PRODUCTION_BRANCH var was not set.

.ci-support/merge-install-production-branch.sh
```diff
@@ -12,28 +12,67 @@ set -x
...
+
+if [[ "$GITHUB_HEAD_REF" == "$PRODUCTION_PR_MAIN_BRANCH" ]]; then
+  # GITHUB_HEAD_REF is only set for PR builds.
+  # https://docs.github.com/en/actions/learn-github-actions/environment-variables
+  PRODUCTION_BRANCH="production-merge/${PRODUCTION_PR_MAIN_BRANCH}"
+  PRODUCTION_FORK=${PRODUCTION_PR_FORK:-"illinois-ceesd"}
+else
+  PRODUCTION_BRANCH="${PRODUCTION_BRANCH:-"production"}"
+  PRODUCTION_FORK=${PRODUCTION_FORK:-"illinois-ceesd"}
+fi
```
MTCam, line 27: I think the `else` is unneeded.  Use what was set just above or default to production proper.

```suggestion
PRODUCTION_BRANCH="${PRODUCTION_BRANCH:-"production"}"
PRODUCTION_FORK=${PRODUCTION_FORK:-"illinois-ceesd"}
```

inducer: I'm not sure I understand what you're asking for here. Your first comment seems to aim at removing `PRODUCTION_{FORK,BRANCH}`, and your second one seems to push for reinstating them. Which of the two would you like? And if you're interested in reinstating them, for what use case that isn't a PR?

MTCam: This change set forces a custom production branch called: `production-merge/${PR_MAIN_BRANCH}`.   I am accustomed to calling my custom production branch whatever I want:
`PRODUCTION_BRANCH=my-custom-production-branch`.    

I would like to retain that capability.   Nothing in this change set prevents my continuing to do that, except the comment snippet that indicates something like that is possible has been removed from `production_test_env.sh`.  I am merely suggesting that snippet should be restored.

inducer: > I am merely suggesting that snippet should be restored.

If it's restored, wouldn't that confuse new users who might see two ways to do the same thing and be confused on which of them they should use? (To my mind, the one with the naming convention is preferable, so I tried to make that the path of least resistance.)

MTCam: To me it isn't the same thing.  It simply allows a user to specify outright the name of the custom production branch to use at their option.

With the new way you are introducing, one is forced to have an independent custom production for each branch/PR, with the advantage that it does not need to be "undone" before merge.   In the way that I typically operate, I can and do use a single custom production branch for several developments/branches/PRs.   I would like to retain that ability.

PR #587, opened 2022-01-07
add possibility to debug multi-rank mirgecom with pudb

matthiasdiener:
**Questions for the review**:

### DIFF
--- a/mirgecom/mpi.py
+++ b/mirgecom/mpi.py
@@ -151,8 +151,8 @@ def mpi_entry_point(func):
 
         _check_gpu_oversubscription()
 
-        # Set this to 'if 1:' in order to debug mirgecom with pudb on rank 0.
-        # Connect to pudb on rank 0 with 'telnet 127.0.0.1 6899'
+        # Change this to 'if 1:' in order to debug mirgecom with pudb on rank 0.
+        # Connect to pudb on rank 0 with 'telnet 127.0.0.1 6899'.
         if 0:
             from pudb.remote import debug_remote_on_single_rank
             debug_remote_on_single_rank(MPI.COMM_WORLD, 0, func, *args, **kwargs)
--- a/mirgecom/mpi.py
+++ b/mirgecom/mpi.py
@@ -151,6 +151,12 @@ def mpi_entry_point(func):
 
         _check_gpu_oversubscription()
 
-        func(*args, **kwargs)
+        # Set this to 'if 1:' in order to debug mirgecom with pudb on rank 0.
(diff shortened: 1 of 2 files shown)


mirgecom/mpi.py
```diff
@@ -151,6 +151,12 @@ def wrapped_func(*args, **kwargs):
 
         _check_gpu_oversubscription()
 
-        func(*args, **kwargs)
+        # Change this to 'if 1:' in order to debug mirgecom with pudb on rank 0.
+        # Connect to pudb on rank 0 with 'telnet 127.0.0.1 6899'.
+        if 0:
+            from pudb.remote import debug_remote_on_single_rank
+            debug_remote_on_single_rank(MPI.COMM_WORLD, 0, func, *args, **kwargs)
```
matthiasdiener, line 158: I'm not sure if there is a better way to enable debugging. Unfortunately, this can not be done in the driver, as the code above prevents access to an MPI communicator.

```python
if "mpi4py.MPI" in sys.modules:
        raise RuntimeError("mpi4py.MPI imported before designated MPI entry "
                           "point. Check for prior imports.")
```

majosm: Could you add something like:

```python
def pudb_remote_on_single_rank(func):
    @wraps(func)
    def wrapped_func(*args, **kwargs):
        # pylint: disable=import-error
        from pudb.remote import debug_remote_on_single_rank
        from mpi4py import MPI
        debug_remote_on_single_rank(MPI.COMM_WORLD, 0, func, *args, **kwargs)

    return wrapped_func
```

and then in the driver do:

```python
@mpi_entry_point
@pudb_remote_on_single_rank
def main(...):
```
?

matthiasdiener: This looks better

matthiasdiener: 0cf62c1

matthiasdiener:
This is ready for review

MTCam (approved):
Any chance we could get an example, howto, or walk-through in the docs?

matthiasdiener:
> Any chance we could get an example, howto, or walk-through in the docs?

Does bd219ae make this more clear?

inducer (approved):
Not opposed. :)

MTCam:
> > Any chance we could get an example, howto, or walk-through in the docs?
> 
> Does [bd219ae](https://github.com/illinois-ceesd/mirgecom/commit/bd219ae432ca6d2fe20ded2996b1668cfedac4c4) make this more clear?

Yes, perfect, thanks!

PR #589, opened 2022-01-10
prepend rank number to all print() invocations

matthiasdiener:
**Questions for the review**:

### DIFF
--- a/mirgecom/mpi.py
+++ b/mirgecom/mpi.py
@@ -199,7 +199,6 @@ if 0:
 
         __builtins__["oldprint"](out_str, *args, **kwargs)
 
-
     if "oldprint" not in __builtins__:
         __builtins__["oldprint"] = __builtins__["print"]
     __builtins__["print"] = rank_print
--- a/mirgecom/mpi.py
+++ b/mirgecom/mpi.py
@@ -158,7 +158,6 @@ def mpi_entry_point(func):
 
 def dprint(*args, **kwargs):
     """Prepends the rank number to the print function."""
-
     from mpi4py import MPI
     out_str = f"[{MPI.COMM_WORLD.Get_rank()}]"
 
--- a/mirgecom/mpi.py
+++ b/mirgecom/mpi.py
@@ -158,8 +158,11 @@ def mpi_entry_point(func):
 
 def dprint(*args, **kwargs):
     """Prepends the rank number to the print function."""
-    from mpi4py import MPI
-    out_str = f"[{MPI.COMM_WORLD.Get_rank()}]"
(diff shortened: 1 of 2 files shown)


inducer (changes requested):
Thanks for working on this, but sorry, I'm not a fan. Changing builtins is kind of a no-no IMO. A better approach might be to use logging more pervasively and then install a logging formatter that automatically includes the rank.

matthiasdiener:
> Thanks for working on this, but sorry, I'm not a fan. Changing builtins is kind of a no-no IMO. A better approach might be to use logging more pervasively and then install a logging formatter that automatically includes the rank.

I was looking at logging formatters as well but didn't find a solution that propagates the formatting across packages; i.e., we can add a formatter to mirgecom logging, but that won't affect grudge etc.

MTCam:
> > Thanks for working on this, but sorry, I'm not a fan. Changing builtins is kind of a no-no IMO. A better approach might be to use logging more pervasively and then install a logging formatter that automatically includes the rank.
> 
> I was looking at logging formatters as well but didn't find a solution that propagates the formatting across packages; i.e., we can add a formatter to mirgecom logging, but that won't affect logging in grudge etc.

It is a miserable situation to need to debug at the level where you need to see the rank of the writer, but I agree that this is extremely useful when one needs.   If we can turn it on-and-off, it seems only slightly sinful to tweak the internal function.

matthiasdiener:
Maybe it would be enough to wrap this in an `if 0:` so that it must be manually enabled?

inducer:
I'm OK with this if it's not turned on unconditionally.

matthiasdiener:
> I'm OK with this if it's not turned on unconditionally (i.e. on import).

What do you think of 5915121?

MTCam (approved):
Looks useful.

mirgecom/mpi.py
```diff
@@ -184,3 +184,21 @@ def wrapped_func(*args: Any, **kwargs: Any):
         debug_remote_on_single_rank(MPI.COMM_WORLD, 0, func, *args, **kwargs)
 
     return wrapped_func
+
+
+# Change this to `if 1:` in order to prepend the rank number to every message
+# printed with print().
+if 0:
```
inducer, line 191:
```suggestion
def enable_rank_labeled_print():
    """Add docstring here.
    """
```
(and hook into docs)

inducer (approved):

mirgecom/mpi.py
matthiasdiener: Done, thanks!

MTCam: Even better. This is cool.

PR #591, opened 2022-01-26
Positivity-preserving limiter and SSPRK timestepping

thomasgibson:
**Questions for the review**:

### DIFF
--- a/mirgecom/integrators/ssprk.py
+++ b/mirgecom/integrators/ssprk.py
@@ -1,7 +1,10 @@
-"""Time-integration module for Mirgecom."""
+"""Timestepping routines for strong-stability preserving Runge-Kutta methods.
+
+.. autofunction:: ssprk43_step
+"""
 
 __copyright__ = """
-Copyright (C) 2020 University of Illinois Board of Trustees
+Copyright (C) 2022 University of Illinois Board of Trustees
 """
 
 __license__ = """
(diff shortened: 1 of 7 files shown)


mirgecom/integrators/ssprk.py
```diff
@@ -0,0 +1,56 @@
...
+def ssprk43_step(state, t, dt, rhs, limiter=None):
+    """Take one step using an explicit 4-stage, 3rd-order, SSPRK method."""
+    actx = state.array_context
+
+    def rhs_update(t, y):
+        return y + dt*rhs(t, y)
+
+    y1 = 1/2*state + 1/2*rhs_update(t, state)
+    if limiter is not None:
+        y1 = thaw(freeze(limiter(y1), actx), actx)
```
inducer, line 42: Move the `thaw(freeze(...))` inside the limiter. Has nothing to do with time stepping.

mirgecom/integrators/ssprk.py
```diff
@@ -0,0 +1,56 @@
...
+
+def ssprk43_step(state, t, dt, rhs, limiter=None):
+    """Take one step using an explicit 4-stage, 3rd-order, SSPRK method."""
+    actx = state.array_context
+
+    def rhs_update(t, y):
+        return y + dt*rhs(t, y)
+
+    y1 = 1/2*state + 1/2*rhs_update(t, state)
+    if limiter is not None:
```
inducer, line 41: Ditch the `if limiter is None` for:
```py
if limiter is None:
    limiter = lambda x: x
```

mirgecom/limiter.py
```diff
@@ -0,0 +1,98 @@
...
+    actx = state.array_context
+
+    def compute_limited_field(field):
+        if not isinstance(field, DOFArray):
+            # vecs is not a DOFArray -> treat as array container
+            return map_array_container(compute_limited_field, field)
+
+        # Compute nodal and elementwise max/mins of the field
+        _mmax = op.nodal_max(dcoll, "banana", field)
+        _mmin = op.nodal_min(dcoll, "avocado", field)
```
inducer, line 72: :banana: :avocado:

mirgecom/limiter.py
```diff
@@ -0,0 +1,98 @@
...
+from meshmode.dof_array import DOFArray
+
+from grudge.discretization import DiscretizationCollection
+
+from mirgecom.fluid import ConservedVars
+
+import grudge.op as op
+
+
+def limiter_liu_osher(dcoll: DiscretizationCollection, state):
```
inducer, line 39: Take in two `dd`s, and two `state`s:
- One on which the nodal minima/maxima are taken.
- One that actually has the limiting applied.

mirgecom/limiter.py
```diff
@@ -0,0 +1,98 @@
...
+            )
+        )
+        return theta*(field - cell_avgs) + cell_avgs
+
+    return ConservedVars(
+        mass=compute_limited_field(state.mass),
+        energy=state.energy,
+        momentum=state.momentum,
+        species_mass=compute_limited_field(state.species_mass)
+    )
```
inducer, line 98: Make it apply to single scalars and/or generic array containers.

mirgecom/limiter.py
```diff
@@ -0,0 +1,98 @@
...
+        An array container containing the filtered field(s).
+    """
+    from grudge.geometry import area_element
+
+    actx = state.array_context
+
+    def compute_limited_field(field):
+        if not isinstance(field, DOFArray):
+            # vecs is not a DOFArray -> treat as array container
+            return map_array_container(compute_limited_field, field)
```
inducer, line 68: use `multi_map_array_container`

mirgecom/limiter.py
inducer, line 39: The base limiter might be a candidate for grudge?

MTCam:
I like this limiter, but would like to suggest some changes along the lines of #600. This change set puts the limiting into the hands of the user so they can choose what fields to limit.   Consider that the original function is incorrectly limiting both mass and species_mass, and offers no user control over what is limited.   

The change set suggested in #600 also should work with any time-integrator.   Note the added combustion example in `autoignition-mpi.py`, and allowing driver control allows the two different limited fields in the combustion case and the `pulse-mpi.py` case.

PR #592, opened 2022-01-27
Add OI to euler and update related tests and utils

MTCam:
Ports @thomasgibson's OI changes to main instead of production.

**Questions for the review**:

### DIFF
--- a/mirgecom/euler.py
+++ b/mirgecom/euler.py
@@ -53,16 +53,27 @@ THE SOFTWARE.
 """
 
 import numpy as np  # noqa
+
+from grudge.dof_desc import DOFDesc, as_dofdesc
+from grudge.trace_pair import (
+    TracePair,
+    interior_trace_pairs
+)
+import grudge.op as op
+
 from mirgecom.inviscid import (
     inviscid_flux,
     inviscid_facial_flux
 )
-
-from grudge.trace_pair import interior_trace_pairs
 from mirgecom.operators import div_operator
+from mirgecom.gas_model import (
+    project_fluid_state,
+    make_fluid_state_trace_pairs
+)
 
 
-def euler_operator(discr, state, gas_model, boundaries, time=0.0):
+def euler_operator(discr, state, gas_model, boundaries, time=0.0,
+                   quadrature_tag=None):
     r"""Compute RHS of the Euler flow equations.
 
     Returns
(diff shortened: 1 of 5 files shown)


thomasgibson:
Looks good, but could you update one of the convergence tests in `test_euler.py` to use overintegration? You can just add a new pytest fixture parameter like `overintegration` which takes on True/False values for the relevant tests.

mirgecom/euler.py
```diff
@@ -95,44 +106,86 @@ def euler_operator(discr, state, gas_model, boundaries, time=0.0):
...
+        discretization to use during operator evaluations.
+        The default value is *None*.
+
     Returns
     -------
-    numpy.ndarray
+    :class:`mirgecom.fluid.ConservedVars`
 
         Agglomerated object array of DOF arrays representing the RHS of the Euler
         flow equations.
```
inducer, line 120:
```suggestion
```

mirgecom/euler.py
```diff
@@ -95,44 +106,86 @@ def euler_operator(discr, state, gas_model, boundaries, time=0.0):
...
+    dd_quad_faces = DOFDesc("all_faces", quadrature_tag)
+
+    def _interp_to_surf_quad(utpair):
+        local_dd = utpair.dd
+        local_dd_quad = local_dd.with_discr_tag(quadrature_tag)
+        return TracePair(
+            local_dd_quad,
+            interior=op.project(discr, local_dd, local_dd_quad, utpair.int),
+            exterior=op.project(discr, local_dd, local_dd_quad, utpair.ext)
+        )
```
inducer, line 133: Does `op.project(utpair)` work? It should.

mirgecom/operators.py
```diff
@@ -28,45 +28,66 @@
...
+        calculated
     flux: numpy.ndarray
-        the boundary fluxes across the faces of the element
+        the boundary flux for each function in v
+
     Returns
     -------
-    meshmode.dof_array.DOFArray or numpy.ndarray
-        the dg divergence operator applied to vector-valued function *u*.
+    float or numpy.ndarray
```
inducer, line 87: Why is this a `float` now?

mirgecom/euler.py
thomasgibson: @inducer It doesn't update `dd`, which is why this is here.

mirgecom/operators.py
thomasgibson: This doesn't make sense. It should return an array container (DOFArray, ConservedVars, object array, etc)

MTCam: Right this was just a mistake.  I'll fix it up.

MTCam: (3916e5f)

MTCam:
> Looks good, but could you update one of the convergence tests in `test_euler.py` to use overintegration? You can just add a new pytest fixture parameter like `overintegration` which takes on True/False values for the relevant tests

(3f7e404)

mirgecom/euler.py
inducer: OIC. Add a comment?

MTCam: (781b8fc)

thomasgibson (approved):
LGTM!

Issue #614, opened 2022-02-25
Verification test for time step estimation of full Navier-Stokes


inducer:
cc @MTCam @lukeolson @jbfreund

inducer:
Discussed possible unit mismatch in https://github.com/illinois-ceesd/mirgecom/blob/9b19b749de2c7d099f4cd5aee67727eda8349225/mirgecom/viscous.py#L370-L373

inducer:
![image](https://user-images.githubusercontent.com/352067/155743051-b404dc9b-684e-49d6-a650-831fc5a355ec.png)

(7.32) in JSH/TW

inducer:
@anderson2981 highlights a units issue in that expression.

MTCam:
JSH/TW make assumptions about the kinematic viscosity and thermal conductivity that make their NS not like ours.  They assume kinematic viscosity is independent of density, and neglect thermal diffusion.   That make their formula OK for their use - but in our case we need to use the actual kinematic viscosity.  Corrected in #615.  

Now to cover it with some tests.

MTCam:
Re: Validation for our timestep estimation.....

Using the `/examples/poiseuille-mpi.py`, a good example of a single gas CNS problem,  we can see that the solution diverges within a few hundred steps for CFL >  ~.3.   Two inviscid examples, `vortex-mpi.py` and `pulse-mpi.py` run stably up to CFL = ~.6.

We'll add a blurb to the documentation about guidelines for selecting time-step size, citing these examples.  We should add some mixture examples, and some verbiage about some of the pertinent dimensionless flow characterization numbers that one can use to estimate the limits for given problems.

PR #615, opened 2022-02-25
Use kinematic not dynamic viscosity for dt calc

MTCam:
Fixes the viscous dt calculation where wrong viscosity was used.

**Questions for the review**:

### DIFF
--- a/mirgecom/viscous.py
+++ b/mirgecom/viscous.py
@@ -357,10 +357,10 @@ def get_viscous_timestep(discr, state):
 
     length_scales = characteristic_lengthscales(state.array_context, discr)
 
-    mu = 0
+    nu = 0
     d_alpha_max = 0
     if state.is_viscous:
-        mu = state.viscosity
+        nu = state.viscosity / state.density
         d_alpha_max = \
             get_local_max_species_diffusivity(
                 state.array_context,
@@ -369,7 +369,7 @@ def get_viscous_timestep(discr, state):
 
     return(
         length_scales / (state.wavespeed
-        + ((mu + d_alpha_max) / length_scales))
+        + ((nu + d_alpha_max) / length_scales))
     )
 
 


inducer:
Would it make sense to include a test (i.e. effectively fix #614) as part of this?

MTCam:
> Would it make sense to include a test (i.e. effectively fix, or start to fix, #614) as part of this?

Yes, I would like that.  Do you have any ideas about how we can effectively test it?

test/test_viscous.py
```diff
@@ -502,45 +503,51 @@ def test_viscous_timestep(actx_factory, dim, mu, vel):
...
     dt_field = get_viscous_timestep(discr, fluid_state)
+    print(f"{dt_field=}")
 
-    speed_total = fluid_state.wavespeed
-    dt_expected = chlen / (speed_total + (mu / chlen))
-
-    error = (dt_expected - dt_field) / dt_expected
-    assert actx.to_numpy(discr.norm(error, np.inf)) == 0
+    error = (dt_field - dt_expected) / dt_expected
+    assert actx.to_numpy(discr.norm(error, np.inf)) < tol
```
inducer, line 553: IIUC this doesn't actually check the stability of the PDE solver at these time steps. Is that something we could incorporate?

MTCam: That's right, it only checks that we implemented the formula correctly.   I think we decided at today's meetings that the validation test(s) for the solver do not need to be integrated and we can add a documentation blurb about (a) reproducing the tests using the provided examples, and (b) how to estimate stability limits for your given problem.

MTCam: > (...) Is that something we could incorporate?

I imagine we can test for a contrived case at what CFL it goes unstable (i.e. when infs, or nans appear in the derived quantitites), and make sure that is somewhere between .1 - 1.1 , or even tighter for inviscid problems.  This is how I've checked it by-hand.

To do this in the unit testing, we need a NS and Euler test steppers that support constant CFL timestepping.  One for Euler is here:
https://github.com/illinois-ceesd/mirgecom/blob/9b19b749de2c7d099f4cd5aee67727eda8349225/test/test_euler.py#L506,

but I do not believe I have tested it with constant CFL yet.   An NS stepper is not something that can go into this PR as NS is downstream of this.

I wonder if it is worth it to test it at this level?  Our timestep estimator is designed to return a conservative estimate for the limiting simulation-wide DT, or CFL, so I think as long as it gives a ballpark measure of the {DT, CFL}, i.e. if the simulation is "stable" at CFL < [.1, 1.1] , then it's ok for this purpose.    

A useful thing to have would be a routine that returns precisely (up to `characterstic_length_scales`) the limiting timestep for each of advection, thermal diffusion, species diffusion, and chemistry, at least so they can be visualized.  Taking the min DT among these should ensure stability.  What we have currently globs them all together to get a conservative value that should ensure stability for the current state.

inducer: https://github.com/inducer/grudge/pull/229

PR #624, opened 2022-03-21
add script to run on multiple GPU ranks on a single node

matthiasdiener:
**Questions for the review**:


matthiasdiener:
Maybe we can find a better name for this file - `multigpu.sh`?

MTCam:
I like these platform-specific scripts that summarize a preferred way to run (e.g. `run_lassen.sh`).   Can we put them in a place like (`/examples/platforms`) or some such thing?  And also make their names consistent?

The script added by this change set appears to introduce the equivalent script for porter (i.e. `run_porter.sh`). 

Both scripts I think will work properly on their respective platforms whether we run on a single or on multiple GPUs, right?

matthiasdiener:
> I like these platform-specific scripts that summarize a preferred way to run (e.g. `run_lassen.sh`). Can we put them in a place like (`/examples/platforms`) or some such thing? And also make their names consistent?

I think it's good to put them in a separate dir, maybe `examples/scripts` ?

> The script added by this change set appears to introduce the equivalent script for porter (i.e. `run_porter.sh`).

Right, but this script could be used on other machines than just porter, maybe `run_gpus.sh`?

> Both scripts I think will work properly on their respective platforms whether we run on a single or on multiple GPUs, right?

Right, but the specific use case of this script is to run on a system without a dedicated launcher that can "separate" GPUs between MPI ranks (ie, this script is not necessary on Lassen, which can do the separation in `jsrun`). It's probably not very useful when you want to run on CPU, or when there is only 1 GPU.

MTCam:
> I think it's good to put them in a separate dir, maybe `examples/scripts` ?
👍 

> Right, but this script could be used on other machines than just porter, maybe `run_gpus.sh`?
> Right, but the specific use case of this script is to run on a system without a dedicated launcher that can "separate" GPUs between MPI ranks (ie, this script is not necessary on Lassen, which can do the separation in `jsrun`). It's probably not very useful when you want to run on CPU, or when there is only 1 GPU.

But it won't work on _all_ platforms, so can the name of it try to reflect that like:
`run_gpus_linux.sh` or `run_gpus_generic.sh` or something like that?   Not a big deal if you don't like it.

MTCam (approved):
How about `run_gpus_mpirun.sh`?  Final and favorite suggestion.

PR #627, opened 2022-03-28
Bump tolerance for `test_lazy_op_euler` (lump)

inducer:
Based on this failure: https://gitlab.tiker.net/inducer/meshmode/-/jobs/392559

### DIFF
--- a/test/test_lazy.py
+++ b/test/test_lazy.py
@@ -254,7 +254,7 @@ def _get_scalar_lump():
         BTAG_ALL: PrescribedFluidBoundary(boundary_state_func=_my_boundary)
     }
 
-    return gas_model, init, boundaries, 5e-12
+    return gas_model, init, boundaries, 1e-11
 
 
 @pytest.mark.parametrize("order", [1, 2, 3])


MTCam (approved):
Isn't 1e-11 starting to be a significant difference between lazy and eager?  -1e-11 is gigantic if we're talking species fractions.

inducer:
> Isn't 1e-11 starting to be a significant difference between lazy and eager? -1e-11 is gigantic if we're talking species fractions.

:shrug: It's only a factor of two over 5e-12. At the same time, yeah, it's a ways away from DP machine epsilon. I agree with your suggestion from chat that these thresholds should be relative, not absolute.

MTCam:
> > Isn't 1e-11 starting to be a significant difference between lazy and eager? -1e-11 is gigantic if we're talking species fractions.
> 
> 🤷 It's only a factor of two over 5e-12. At the same time, yeah, it's a ways away from DP machine epsilon. I agree with your suggestion from chat that these thresholds should be relative, not absolute.

This is true, but our previous bump was from 5e-15 to 5e-12, and we grumbled a little bit over that.   Heh.   Regardless, as we know absolute differences have limited value for this comparison.   I'm just not inclined to go fix it this instant.  I already have an issue in about this and some other stuff with the comparison util.

PR #631, opened 2022-04-04
Standalone NS gradients

majosm:
Splits CV gradient and temperature gradient computations into standalone functions so that they can be called outside of NS.

**Questions for the review**:

### DIFF
--- a/mirgecom/navierstokes.py
+++ b/mirgecom/navierstokes.py
@@ -28,6 +28,8 @@ where:
 RHS Evaluation
 ^^^^^^^^^^^^^^
 
+.. autofunction:: grad_cv_operator
+.. autofunction:: grad_t_operator
 .. autofunction:: ns_operator
 """
 
@@ -55,11 +57,13 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
+from functools import partial
+
 from grudge.trace_pair import (
     TracePair,
     interior_trace_pairs
 )
-from grudge.dof_desc import DOFDesc, as_dofdesc
+from grudge.dof_desc import DOFDesc, as_dofdesc, DISCR_TAG_BASE
 
 import grudge.op as op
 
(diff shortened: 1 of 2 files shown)


MTCam:
I just wanted to point out that `make_operator_fluid_states` is expensive:

- It communicates `CV`, and `Tseed` across the partition boundaries *expensive* 
- It projects `CV` and `Tseed` to all the domain boundaries
- It computes fluid temperature on each boundary using 'CV', 'Tseed' *expensive* for mixtures 

It should not be called multiple times from the same (or even from different) operators.

majosm:
> I just wanted to point out that `make_operator_fluid_states` is expensive:
> 
> * It communicates `CV`, and `Tseed` across the partition boundaries _expensive_
> * It projects `CV` and `Tseed` to all the domain boundaries
> * It computes fluid temperature on each boundary using 'CV', 'Tseed' _expensive_ for mixtures
> 
> It should not be called multiple times from the same (or even from different) operators.

Hmm. Should I add an `operator_states_quad` argument to `ns_operator` too, then? I need to compute the gradients (well, temperature gradient at least) outside of `ns_operator` in order to set up BCs in the coupled case.

MTCam:
> > I just wanted to point out that `make_operator_fluid_states` is expensive:
> > 
> > * It communicates `CV`, and `Tseed` across the partition boundaries _expensive_
> > * It projects `CV` and `Tseed` to all the domain boundaries
> > * It computes fluid temperature on each boundary using 'CV', 'Tseed' _expensive_ for mixtures
> > 
> > It should not be called multiple times from the same (or even from different) operators.
> 
> Hmm. Should I add an `operator_states_quad` argument to `ns_operator` too, then? I need to compute the gradients (well, temperature gradient at least) outside of `ns_operator` in order to set up BCs in the coupled case.

Yeah, there's no great solution jumping out at me.   @inducer and I talked about taking the state preparation out of the operator itself before, and there is some light pressure to do so because the AV operator also just repeats all the same communication and gradient calculation that the NS operator does.  

I see no other choice for us than to go ahead and prepare the "operator states", grad(CV), and grad(T) and then share them among components.  Potentially, we don't need to tear apart the existing monolithic operators to do this; we _could_ write a specialized "multiphysics" super-operator or some such.  What about something like that?

PR #632, opened 2022-04-04
Modernize `diffusion_operator`

majosm:
Applies some updates to `diffusion_operator` originating from the multi-physics work.

**Questions for the review**:

### DIFF
--- a/examples/heat-source-mpi.py
+++ b/examples/heat-source-mpi.py
@@ -153,7 +153,7 @@ def main(actx_class, ctx_factory=cl.create_some_context, use_logmgr=True,
 
     def rhs(t, u):
         return (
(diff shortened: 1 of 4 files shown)


MTCam (approved):
Not a giant deal: I sort of don't like how the `diffusion_operator` doesn't advertise its interface in the prototype.   I sort of do like your argument normalization - at least as an example of what you can do to meddle with an interface/api.

majosm:
> Not a giant deal: I sort of don't like how the `diffusion_operator` doesn't advertise its interface in the prototype.

Same. The normalization stuff is a yucky hack to handle a deprecated positional parameter (`quad_tag`). It will go away whenever `quad_tag` is fully removed.

PR #636, opened 2022-04-07
git cantera

MTCam:
Try installing Cantera from git instead of using conda.

**Questions for the review**:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,4 +22,4 @@ pyyaml
 --editable git+https://github.com/inducer/pytato.git#egg=pytato
 --editable git+https://github.com/ecisneros8/pyrometheus.git#egg=pyrometheus
 --editable git+https://github.com/illinois-ceesd/logpyle.git#egg=logpyle
---editable git+https://github.com/Cantera/cantera.git#egg=cantera
+git+https://github.com/Cantera/cantera.git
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -22,3 +22,4 @@ pyyaml
 --editable git+https://github.com/inducer/pytato.git#egg=pytato
 --editable git+https://github.com/ecisneros8/pyrometheus.git#egg=pyrometheus
 --editable git+https://github.com/illinois-ceesd/logpyle.git#egg=logpyle
+--editable git+https://github.com/Cantera/cantera.git#egg=cantera
```
inducer, line 25: Given how much complexity there is in the cantera conda feedstock:

https://github.com/conda-forge/cantera-feedstock/tree/main/recipe

I'm not hopeful that this will get anywhere close to working...

inducer: (And even if it does, I think it'll impose a substantial build time penalty.)

MTCam: > (And even if it does, I think it'll impose a substantial build time penalty.)

Well it doesn't appear to work at all, so ugh.  Back to the drawerring board for finding a way to use the transport features that we need from it.

matthiasdiener: Is there a specific patch that is needed? If so, we could add it to the cantera-feedstock for conda.

PR #638, opened 2022-04-12
Move forced evaluation inside `logmgr` tick interval and make it smarter

majosm:
I think this is how it should be, right?

To-do:
- [x] :warning: Merge production version and reset CI back to production before merging

**Questions for the review**:

### DIFF
--- a/.ci-support/production-testing-env.sh
+++ b/.ci-support/production-testing-env.sh
@@ -8,8 +8,8 @@ set -x
 # The production capability may be in a CEESD-local mirgecom branch or in a
 # fork, and is specified through the following settings:
 #
-export PRODUCTION_BRANCH="fix-forced-evaluation-production"   # The base production branch to be installed by emirge
-export PRODUCTION_FORK="majosm"  # The fork/home of production changes (if any)
+# export PRODUCTION_BRANCH=""   # The base production branch to be installed by emirge
+# export PRODUCTION_FORK=""  # The fork/home of production changes (if any)
 #
 # Multiple production drivers are supported. The user should provide a ':'-delimited
 # list of driver locations, where each driver location is of the form:


examples/heat-source-mpi.py
```diff
@@ -172,6 +173,9 @@ def rhs(t, u):
                         ], overwrite=True)
 
         u = rk4_step(u, t, dt, compiled_rhs)
+        # Force evaluation once per timestep
+        u = thaw(freeze(u), actx)
```
matthiasdiener, line 177: Why not use _force_evaluation here?

majosm: How's https://github.com/illinois-ceesd/mirgecom/pull/638/commits/19f50233371b10a94cb971dfee203f9c0fa6e1cf?

matthiasdiener: LGTM, thanks!

MTCam:
I am not sure I am a fan of `force_eval` down inside the stepper.  What if we don't need to force eval at all?  Having it down here makes it mandatory for every step - regardless of whether the result will be used.   Does that matter?

inducer:
Given https://github.com/inducer/grudge/pull/252, I think `force_eval` will be a thing of the past rather sooner than later.

MTCam:
We need force eval to be optional to support the use of pre-compiled time steppers.   Can we tweak this slightly to include changes like these:

https://github.com/illinois-ceesd/mirgecom/blob/021c1e86575d48a11f071352927d6391ab579a12/mirgecom/steppers.py#L131

majosm:
> We need force eval to be optional to support the use of pre-compiled time steppers. Can we tweak this slightly to include changes like these:
> 
> https://github.com/illinois-ceesd/mirgecom/blob/021c1e86575d48a11f071352927d6391ab579a12/mirgecom/steppers.py#L131

Added. Got me wondering, though -- since the need for evaluation forcing depends on the driver and the choice of timestepper, should it maybe just be done inside a driver callback instead? What do you think about something like https://github.com/illinois-ceesd/mirgecom/pull/638/commits/700ada56cd063d3d7b5cabb8247766d2cbaac71a? (Pretend the `force_eval` argument is gone, that's just in there for compatibility.)

inducer:
> We need force eval to be optional to support the use of pre-compiled time steppers. Can we tweak this slightly to include changes like these:

To be clear: Precompiled time steppers do not preclude use of force-eval. But: force-eval imposes a Python-side overhead that is unnecessary in that setting and should be avoided if at all possible.

MTCam:
> > We need force eval to be optional to support the use of pre-compiled time steppers. Can we tweak this slightly to include changes like these:
> 
> To be clear: Precompiled time steppers do not preclude use of force-eval. But: force-eval imposes a Python-side overhead that is unnecessary in that setting and should be avoided if at all possible.

Oh sure!  We just don't want to do it if we don't have to, though!

majosm:
Can someone remind me what exactly happens in the non-precompiled case when you don't force evaluation? Is the DAG just snowballing over multiple steps, or is it something else? I don't fully understand what `compiled_lsrk45_step` is doing differently.

inducer:
It snowballs.

majosm:
What does the precompiled version do to prevent that?

inducer:
All computation happens inside of `compile`.

MTCam:
This production driver issue (https://github.com/illinois-ceesd/mirgecom/runs/7139424067?check_suite_focus=true#step:4:187) is , I think, due to Kaushik's branches need to get updated since this stuff landed:

https://github.com/inducer/arraycontext/pull/183

and requiring these fixes to grudge:

https://github.com/inducer/grudge/pull/270

kaushikcfd:
> I think, due to @kaushikcfd branches need to get updated since this stuff landed:

Rebased my branches!

MTCam:
> > I think, due to @kaushikcfd branches need to get updated since this stuff landed:
> 
> Rebased my branches!

Thanks, that did it!

majosm:
@inducer @MTCam What do you think about the current approach for deducing whether evaluation forcing is necessary? It appears to give the desired behavior for the current MIRGE-Com integrators and the pre-compiled LSRK45.

One slightly questionable aspect of this implementation is that it depends on a certain behavior of `actx.np.array_equal` that I'm not sure is intentional. Namely, it checks whether the contained pytato expressions are the same, not whether they evaluate to the same result.

mirgecom/steppers.py
```diff
@@ -64,17 +61,26 @@ def get_rhs():
...
+    If *ary1* and *ary2* are :mod:`pytato` arrays, additionally checks whether the
+    contained expressions are the same (not just whether they evaluate to the same
+    result).
+    """
     if actx is None:
-        return state
-    return thaw(freeze(state, actx), actx)
+        return np.array_equal(ary1, ary2)
+
+    return actx.to_numpy(actx.np.array_equal(ary1, ary2))[()]
```
inducer, line 75: Pytato arrays use `==` for structural equality. This is one intentional departure from numpy. The ability to use them in mappings in mappings without gymnastics was the overriding concern.

mirgecom/steppers.py
```diff
@@ -136,6 +144,22 @@ def _advance_state_stepper_func(rhs, timestepper,
 
         state = timestepper(state=state, t=t, dt=dt, rhs=compiled_rhs)
 
+        if force_eval is None:
+            unevaluated_state = state
+            state = force_evaluation(actx, state)
+            if not _strict_array_equal(actx, state, unevaluated_state):
```
inducer, line 150: I don't think it's necessary to base this on equality. You could simply traverse the array container and look for the first instance of `pytato.Array` that's not a `pytato.DataWrapper`. If you find any, it means that you would need force-eval. We could/should probably offer this predicate in `arraycontext`.

MTCam:
> @inducer @MTCam What do you think about the current approach for deducing whether evaluation forcing is necessary? It appears to give the desired behavior for the current MIRGE-Com integrators and the pre-compiled LSRK45.

@majosm I like this approach but I do have some concerns about what we're doing that are not made worse by this development, but I want to bring them up since this might be a good opportunity to fix it up.

The biggest problems (to my knowledge) we've encountered that were of-and-around force_evaluation:
 - gigantic compile times:   Caused when an "unevaluated" state expression is passed to `advance_state` initially.  Now we are careful to `force_evaluate` _before_ calling `advance_state`, but it would be great if it (advance routine) detected and evaluated any unevaluated expn automatically.  We still find lingering drivers that have this problem.
 - real bad runtime performance in lazy:  Sometimes the `pre_step_callback` (or even `post_step_callback` modifies the `state` and/or the `dt` inside and must `force_evaluate` before passing back to the stepper else they incur very large performance penalties.   I think we could save the user from themselves by checking the callbacks similar to how you are checking the timestepper.
 -  spurious timing results:  When force eval is not down inside the step-to-step timer (i.e. between the callbacks where the timing is started/stopped).
 
The first bullet issue can easily be fixed by this development if a check is added to the incoming state (and DT), to make sure neither of those have unevaluated expns.

This change set already addresses the last bullet issue, and makes the handling of setting force_eval automatically do the right thing (if the user didn't already force the wrong thing) w.r.t  the timestepping routine (i.e. whether it is pre-compiled or not).  I think that the choice of timestepper is not the only reason for needing to do `force_evaluation` down inside the stepping, however.  

Example: Tulio's spatially varying DT for steady-state problems, or even plain-ole constant CFL mode.   Both of these send back an "expression" stuffed into DT!  Force evaluating DT inside the callbacks fixed the observed performance problem with those modes, big time.  This also solved the ongoing memory issues he was having.   This problem is more around the 2nd bullet.   

The second bullet issues are trickier (to me).  I really like how this development only needs to evaluate whether force eval is required after the timestepper only once, for all of (simulated) time.   It would be terrific if we could check the same for the callbacks and automatically do that too.   I imagine we could actually do one test call to each of pre, post, and timestepper _outside_ of stepping to check whether any force eval is needed  - but we'd risk missing some cases that only periodically update state or dt.  (e.g. filtering or constant_cfl mode).   Ha, currently we miss them *all*, and it is up to the driver writer to make sure they force_evaluate when needed.

We can handle (i.e. continue handling) 2nd bullet issues entirely driver-side, by force evaluating on input and output of callbacks.   But not all driver writers do this, and even less of them know when to do this, and less still know why.  It might be good if the infrastructure assumed some of the responsibility for doing it - or at least issued warnings about it like this development does.  The other thing I really like about your approach here is that the user still has the option to turn it on and off at will.  It would be nice to address the 2nd bullet issues similarly.

MTCam:
> > * gigantic compile times:   Caused when an "unevaluated" state expression is passed to `advance_state` initially.  Now we are careful to `force_evaluate` _before_ calling `advance_state`, but it would be great if it (advance routine) detected and evaluated any unevaluated expn automatically.  We still find lingering drivers that have this problem.
> 
> It looks like there's already a warning emitted when users pass unevaluated state to a compiled function, e.g. `rhs`:
> 
> ```
> mirgecom/integrators/explicit_rk.py:33: DeprecationWarning: Argument array '(1, 0, 'energy', 0)' to a compiled function is unevaluated. Evaluating just-in-time, at considerable expense. This is deprecated and will stop working in 2023. To avoid this warning, force evaluation of all arguments via freeze/thaw.
>   k1 = rhs(t, state)
> ```
> 
> Do you think that's enough?
> 

I think that depends on what happens after sunset of this feature.   There should be no unevaluated states being passed to the advancing routine - so it seems reasonable (to me) to just call `force_evaluation` on any incoming state.   It really doesn't need to detect whether it is needed as it does no harm when it is not needed, but saves the developer a heap of trouble when it is.


> > * real bad runtime performance in lazy:  Sometimes the `pre_step_callback` (or even `post_step_callback` modifies the `state` and/or the `dt` inside and must `force_evaluate` before passing back to the stepper else they incur very large performance penalties.   I think we could save the user from themselves by checking the callbacks similar to how you are checking the timestepper.
> 
> If I understand the situation correctly from what AK has told me, forcing evaluation is more of a fallback than a solution. Ideally we shouldn't be doing it at all in the timestepping loop because it involves array container traversals. The optimal solution is to only do computation inside compiled functions. I could possibly add these fallbacks/warnings for the pre/post-step callbacks though, if you think that would be useful to have.

The situation occurs because user/developers don't know when to make sure to do it.   When we have constant_cfl mode, we *must* do it for DT, else the performance is very bad.  There is no option to not force_eval if you compute [state, dt] in the callbacks.   That said, we can continue limping along without addressing that here, and simply try to make sure developers are educated enough to know this must be done.  It would be superior to do it inside the infrastructure, imo, and save me the trouble of going along and updating all the examples and drivers to ensure they are doing the right thing when constant CFL mode is used.

majosm:
> I think that depends on what happens after sunset of this feature. There should be no unevaluated states being passed to the advancing routine - so it seems reasonable (to me) to just call `force_evaluation` on any incoming state. It really doesn't need to detect whether it is needed as it does no harm when it is not needed, but saves the developer a heap of trouble when it is.

Sounds fair. I don't really see a downside since it only happens once per `advance_state` call.

> The situation occurs because user/developers don't know when to make sure to do it. When we have constant_cfl mode, we _must_ do it for DT, else the performance is very bad. There is no option to not force_eval if you compute [state, dt] in the callbacks. That said, we can continue limping along without addressing that here, and simply try to make sure developers are educated enough to know this must be done. It would be superior to do it inside the infrastructure, imo, and save me the trouble of going along and updating all the examples and drivers to ensure they are doing the right thing when constant CFL mode is used.

AFAIK you don't need to force evaluation if the code is compiled; that should probably be encouraged instead of adding `force_evaluation` calls if possible, since it will be more efficient. Possibly `advance_state` could be made to check for missing `compile`s (doing basically the same thing it's doing now with the timestepper), but there could be more pitfalls for this case (e.g., if the `state`/`dt` modification is not done every timestep).

Andreas also mentioned something about potentially just passing pointers between compiled functions instead of constructing python objects, in order to avoid the overhead of doing any work on the python side during timestepping. Do you know anything about that @kaushikcfd? Is that something that could become a reality soon, or is it just an idea at this point?

MTCam:
> > I think that depends on what happens after sunset of this feature. There should be no unevaluated states being passed to the advancing routine - so it seems reasonable (to me) to just call `force_evaluation` on any incoming state. It really doesn't need to detect whether it is needed as it does no harm when it is not needed, but saves the developer a heap of trouble when it is.
> 
> Sounds fair. I don't really see a downside since it only happens once per `advance_state` call.

👍  - Yeah just once per `advance_state` call, which may be just once per simulation job. 
 
> AFAIK you don't need to force evaluation if the code is compiled; that should probably be encouraged instead of adding `force_evaluation` calls if possible, since it will be more efficient. Possibly `advance_state` could be made to check for missing `compile`s (doing basically the same thing it's doing now with the timestepper), but there could be more pitfalls for this case (e.g., if the `state`/`dt` modification is not done every timestep).

OK have patience with me here please; I don't fully grok the freeze/thaw/compile/to_numpy stuff. What I think you are saying is that we should be using _compiled_ functions to modify the state and/or dt, so that we do not need to force_eval down in the callbacks.   I do _try_ to use compiled functions in the callbacks - but sometimes (e.g. in health checking), I already need to "evaluate" because I need numbers there.  Health checking is mostly local; reducing only the result of the health checking, but constant cfl timestep modification also requires reductions which (afaik) is done in actual Python and requires numbers.  I'm not sure how to get around needing to make sure to force eval before-and-after these operations.

> 
> Andreas also mentioned something about potentially just passing pointers between compiled functions instead of constructing python objects, in order to avoid the overhead of doing any work on the python side during timestepping. Do you know anything about that @kaushikcfd? Is that something that could become a reality soon, or is it just an idea at this point?

PR #643, opened 2022-04-21
CI: make lazy examples run distributed

matthiasdiener:
**Questions for the review**:

### DIFF
--- a/examples/compare_lazy_solution.sh
+++ b/examples/compare_lazy_solution.sh
@@ -21,13 +21,14 @@ else
 fi
 MPIARGS=
 if [[ "$SIM" == *"-mpi.py" ]]; then
+    MPIRUN="mpirun -n 2"
     MPIARGS=" -m mpi4py "
 fi
 casename_base=$(echo ${SIM/%.py})
-python ${MPIARGS} ${SIM} --casename ${casename_base}-eager
-python ${MPIARGS} ${SIM} --casename ${casename_base}-lazy --lazy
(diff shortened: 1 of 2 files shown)


MTCam:
Can we add comparisons for these example results for the eager vs. lazy runs?  I think we just need to add the names of the examples we want to test to `test_lazy_accuracy.sh`.

matthiasdiener:
> Can we add comparisons for these example results for the eager vs. lazy runs? I think we just need to add the names of the examples we want to test to `test_lazy_accuracy.sh`.

Which ones should we add? I think adding all of them will make this CI test very slow.

MTCam:
> > Can we add comparisons for these example results for the eager vs. lazy runs? I think we just need to add the names of the examples we want to test to `test_lazy_accuracy.sh`.
> 
> Which ones should we add? I think adding all of them will make this CI test very slow.

I'm not sure - if they run in lazy I think they are already generating all the data.  We just need to run the comparison util.  If the lazy compares run the examples multiple times then we should fix that so that it doesn't do that any more to speed up CI.   If they run both eager and lazy, I think we should check the results (i.e., if it is not costly).

inducer (approved):
More CI is always better. You guys decide what you'd like cost-wise. :)

MTCam:
I just want to make sure we keep our eye on this ().   When this change propagates up through main - running `mixture-mpi.py` currently hangs. (#645)

MTCam:
Safe to close?   The lazy examples are running distributed now.

matthiasdiener:
Sure, it's fine to close this. Should we start removing the `-lazy` symlinks?

Issue #509, opened 2021-09-20
 Housecleaning


MTCam:
There are several changes in MIRGE-Com that are easy to do, but touch a lot of pieces. Here's a list of the stuff we want to do:

- [x] rename `allsync` to something with reduction in the name (AK: `global_reduction`, or similar) (#521)
- [x] massage docs for `global_reduction` : https://github.com/illinois-ceesd/mirgecom/commit/c08bbbe76bc5458ae5bebca151cb42c272a64a02#r56757125 (#521)
- [x] remove the `local` argument for some routines that triggers projection
- [ ] Replace 'discr' -> 'dcoll' in documentation, and interfaces

MTCam:
Closed by #728

PR #648, opened 2022-04-27
Add HLL numerical flux

MTCam:
This change set:
- adds Harten-Lax-van Leer (HLL) numerical flux routines implemented by @anderson2981,
  - https://link.springer.com/chapter/10.1007/978-3-662-03490-3_10?noAccess=true
- and adds/refactors numerical flux interfaces supporting user-specified numerical flux routines
  - `euler.py`, `inviscid.py`, `flux.py`, `viscous.py`, `boundary.py`
- adds a suite of tests and extends some existing tests for the new fluxes
  - `test_bc.py`, `test_inviscid.py`, `test_flux.py` 
- adds Toro reference on approximate Riemann solvers and sharpens the documentation on BCs and numerical fluxes
   - `misc.rst`, `discretization.rst`

 **majosm review checklist**:
- [ ] ~~Is everything that is implemented covered by tests?~~
  - _Can't answer this; need @anderson2981 to look_

 **anderson2981 review checklist**:


mirgecom/inviscid.py
```diff
@@ -128,32 +128,183 @@ def inviscid_facial_flux(discr, state_tpair, local=False):
         the face normals as required by the divergence operator for which they
         are being computed.
     """
```
majosm, line 130: Documented arguments need updating (or maybe removal, to match `inviscid_flux_hll`); they don't match the function prototype.

mirgecom/inviscid.py
```diff
@@ -128,32 +128,183 @@ def inviscid_facial_flux(discr, state_tpair, local=False):
...
-    flux_tpair = TracePair(dd,
-                           interior=inviscid_flux(state_tpair.int),
-                           exterior=inviscid_flux(state_tpair.ext))
 
-    # This calculates the local maximum eigenvalue of the flux Jacobian
-    # for a single component gas, i.e. the element-local max wavespeed |v| + c.
-    w_int = state_tpair.int.speed_of_sound + state_tpair.int.speed
-    w_ext = state_tpair.ext.speed_of_sound + state_tpair.ext.speed
-    lam = actx.np.maximum(w_int, w_ext)
+def inviscid_flux_hll(state_pair, gas_model, normal, **kwargs):
```
majosm, line 140: Should this and `inviscid_flux_rusanov` maybe be called `inviscid_facial_flux_<rusanov/hll>`, since they return the normal flux?

majosm (commented):
Made a couple of passes, mostly looking at code clarity and documentation. Hopefully @anderson2981 can look more closely at the math.

mirgecom/inviscid.py
```diff
@@ -128,32 +128,183 @@ def inviscid_facial_flux(discr, state_tpair, local=False):
...
+        A CV object containing the scalar numerical fluxes at the input faces.
+        The returned fluxes are scalar because they've already been dotted with
+        the face normals as required by the divergence operator for which they
+        are being computed.
+    """
+    normal = thaw(discr.normal(state_pair.dd), state_pair.int.array_context)
+    num_flux = numerical_flux_func(state_pair, gas_model, normal)
+    dd = state_pair.dd
+    dd_allfaces = dd.with_dtag("all_faces")
+    return num_flux if local else discr.project(dd, dd_allfaces, num_flux)
```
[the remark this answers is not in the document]
MTCam: My thinking here is that the family of numerical fluxes we use are easily written as functions of (state_pair, gas_model, normal).  This interface allows physics users to easily add their own numerical flux routines (especially those described by Toro and Mengaldo) using quantities they understand well, without needing to worry about the more infrastructure,y stuff in this wrapper interface.  It also leaves less room to get this part wrong on accident, and eliminates repeated code in the fluxy suite of routines.

mirgecom/inviscid.py
MTCam: I am not opposed to such a renaming, but fwiw, the fluxes we are using according to their definitions in the approximate Riemann solver references (e.g., https://edanya.uma.es/NSPDE/cursos/toro/2-HLL+HLLC.pdf), are defined as interface-normal fluxes already.

mirgecom/inviscid.py
```diff
@@ -128,32 +128,183 @@ def inviscid_facial_flux(discr, state_tpair, local=False):
...
+    q_ext = 1 + (gamma_ext + 1)/(2*gamma_ext)*(p_star/p_ext - 1)
+
+    pres_check_int = actx.np.greater(p_star, p_int)
+    pres_check_ext = actx.np.greater(p_star, p_ext)
+
+    q_int = actx.np.where(pres_check_int, q_int, ones)
+    q_ext = actx.np.where(pres_check_ext, q_ext, ones)
+
+    q_int = actx.np.sqrt(q_int)
+    q_ext = actx.np.sqrt(q_ext)
```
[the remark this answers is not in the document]
MTCam: ([1138e69](https://github.com/illinois-ceesd/mirgecom/pull/648/commits/1138e690903e4e6612b8f29e57b3d2dbe7922ab6))

mirgecom/flux.py
```diff
@@ -35,179 +40,205 @@
 import numpy as np  # noqa
 
 
-def gradient_flux_central(u_tpair, normal):
-    r"""Compute a central flux for the gradient operator.
+def num_flux_lfr(f_minus_normal, f_plus_normal, q_minus, q_plus, lam, **kwargs):
```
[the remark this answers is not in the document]
MTCam: These are not the **kwargs you are looking for. :wave: 
([1138e69](https://github.com/illinois-ceesd/mirgecom/pull/648/commits/1138e690903e4e6612b8f29e57b3d2dbe7922ab6))

mirgecom/inviscid.py
```diff
@@ -84,20 +85,19 @@ def inviscid_flux(state):
                           momentum=mom_flux, species_mass=species_mass_flux)
 
 
-def inviscid_facial_flux(discr, state_tpair, local=False):
-    r"""Return the flux across a face given the solution on both sides *q_tpair*.
+def inviscid_flux_rusanov(state_pair, gas_model, normal, **kwargs):
```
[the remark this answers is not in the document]
MTCam: ([1138e69](https://github.com/illinois-ceesd/mirgecom/pull/648/commits/1138e690903e4e6612b8f29e57b3d2dbe7922ab6))

mirgecom/flux.py
```diff
@@ -35,179 +40,205 @@
...
+    r"""Compute Lax-Friedrichs/Rusanov flux after [Hesthaven_2008]_, Section 6.6.
 
-    The central gradient flux, $\mathbf{h}$, of a scalar quantity $u$ is calculated
-    as:
+    The Lax-Friedrichs/Rusanov flux is calculated as:
 
     .. math::
+        f_{\text{LFR}}=\frac{1}{2}\left(f^-+f^+-\lambda\left(q^--q^+\right)\right)
```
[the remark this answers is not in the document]
MTCam: Thanks! I had the a sign error in the formula, but the implementation was ok.  Fixed that, and also flipped the sign around on the implementation to make it match closely to the doc.
([1138e69](https://github.com/illinois-ceesd/mirgecom/pull/648/commits/1138e690903e4e6612b8f29e57b3d2dbe7922ab6))

mirgecom/flux.py
```diff
@@ -35,179 +40,205 @@
...
+        Normal component of physical flux interior to (left of) interface
+
+    f_plus_normal
+        Normal component of physical flux exterior to (right of) interface
+
+    q_minus
+        Physical state on interior of interface
+
+    q_plus
+        Physical state on exterior of interface
```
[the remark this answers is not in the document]
MTCam: Well we wanted this interface to be numerical quantities only with no grudgy data structures down here.

mirgecom/inviscid.py
MTCam: Fixed this up.

mirgecom/inviscid.py
```diff
@@ -128,32 +128,183 @@ def inviscid_facial_flux(discr, state_tpair, local=False):
...
+def inviscid_boundary_flux_for_divergence_operator(
+        discr, gas_model, boundaries, interior_boundary_states,
+        domain_boundary_states, quadrature_tag=None,
+        numerical_flux_func=inviscid_flux_rusanov, time=0.0):
+    """Compute the inviscid boundary fluxes for the divergence operator.
+
+    This routine encapsulates the computation of the inviscid contributions
+    to the boundary fluxes for use by the divergence operator. Its existence
+    is intended to allow multiple operators (e.g. Euler and Navier-Stokes) to
+    perform the computation without duplicating code.
```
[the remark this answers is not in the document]
MTCam: Please have another look @majosm. I tried to fix up the names I hope more to your liking, and maybe describing what they do more accurately. 

I do not disagree with your notion to rename... but just to explain my thinking here; this code in the fluid operators (and the supporting fluid infrastructure) is describing the treatment of a single DG element.  Boundary in that context is just the element boundary regardless of whether there is another element on the other side (local or remote) or the domain boundary.  That's what I mean by interior-boundary vs. domain-boundary.  

The interior boundaries (including those faces on the partition boundaries) are handled with `inviscid_facial_flux` as we have all the info we need (+) and (-) states and use the Riemann solver there.   Domain boundary handled by the BC routine and we have only the (-) side state available as input for those.  The BC is expected to return the numerical flux that will be used for that face.  How the BC gets the flux is the BC's business.  Users are free to implement time-dependent fluxing functions for those - thus the **kwargs so that the other flux routines don't holler when they get a `time` argument or whatever other parameters the user's flux routine might require.

mirgecom/inviscid.py
majosm: Heh, sorry -- I meant to comment on `inviscid_flux_rusanov`, but I guess I scrolled down too far. What I was getting at here is that `inviscid_flux_rusanov` has documentation for arguments that are no longer present (`discr`, `local`, etc.). Presumably they should be made to look more like the docs for `inviscid_flux_hll`.

mirgecom/inviscid.py
majosm: That's fine, I was just thinking we should try to stick to a consistent naming convention in the code if we can, i.e., `*_flux` for vectors, `*_facial_flux` for normal fluxes.

mirgecom/flux.py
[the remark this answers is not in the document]
majosm: How come? I don't see allowing `TracePair` as being too egregious... and then you get to use `tpair.avg`, `tpair.diff`, etc.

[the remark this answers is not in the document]
MTCam: At the time they were written, it allowed just sending numbers or numpy arrays to the low-level flux routines in such a way that we could just copy tests right out of the book.  The extra baggage of needing TracePairs data structure to call and test the numerical flux function.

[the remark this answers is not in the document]
majosm: Ah, OK. I guess that's fair.

mirgecom/inviscid.py
MTCam: ([72bc8a5](https://github.com/illinois-ceesd/mirgecom/pull/648/commits/72bc8a5f9210852630243c1d1d8eb03a1d9421d6))

mirgecom/inviscid.py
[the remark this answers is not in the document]
majosm: Still some doc issues to resolve, but thinking something along the lines of https://github.com/majosm/mirgecom/commit/91ab368462bca41c72d89910467345de7a77ba8a

mirgecom/inviscid.py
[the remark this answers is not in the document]
majosm: Do you have an example of an instance where this comes up?

mirgecom/inviscid.py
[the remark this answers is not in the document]
MTCam: I think I like that.  Still trying to grok.

mirgecom/inviscid.py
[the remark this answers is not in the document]
MTCam: No, and in light of your recent proposed changes, maybe it is unnecessary.

mirgecom/inviscid.py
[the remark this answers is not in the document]
majosm: Sounds good! I don't know if you happened to see #652, but that's the version with the issues fixed (and viscous stuff added, and up to date with `flux` -> `facial_flux` renaming.) The doc issue was fixed by unindenting the docstring inside `_as_facial_flux`. I also needed to add an explicit `name` parameter to `_as_facial_flux` to get the function naming set right. I think that's all for major differences. We can close that PR if you don't need it.

mirgecom/flux.py
```diff
@@ -35,179 +40,205 @@
...
+        HLL numerical flux.
     """
-    return flux_lfr(cv_tpair, f_tpair, normal, lam)@normal
+    actx = q_minus.array_context
+    f_star = (s_plus*f_minus_normal - s_minus*f_plus_normal
+              + s_plus*s_minus*(q_plus - q_minus))/(s_plus - s_minus)
 
+    # choose the correct f contribution based on the wave speeds
+    f_check_minus = \
+        actx.np.greater_equal(s_minus, 0*s_minus)*(0*f_minus_normal + 1.0)
```
inducer, line 167: `actx.zeros_like(s_minus)` or so?

Disclaimer: Doesn't totally work yet: https://github.com/inducer/arraycontext/issues/161

mirgecom/flux.py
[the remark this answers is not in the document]
inducer: Unresolving to say: Add a comment justifying the choice to supply different variables.

mirgecom/inviscid.py
```diff
@@ -84,34 +85,147 @@ def inviscid_flux(state):
                           momentum=mom_flux, species_mass=species_mass_flux)
 
 
-def inviscid_facial_flux(discr, state_tpair, local=False):
-    r"""Return the flux across a face given the solution on both sides *q_tpair*.
+def inviscid_facial_flux_rusanov(state_pair, gas_model, normal, **kwargs):
```
inducer, line 88: Remove kwargs?

mirgecom/inviscid.py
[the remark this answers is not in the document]
majosm: Is https://github.com/majosm/mirgecom/commit/9176e74aff79cc39d5f9d3cdb7f5d8cf59424c68 the way to go here? And if so, do we want to implement that in this PR or save it for later?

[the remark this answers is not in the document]
MTCam: Yes, and yes _this_ PR.  It is low-hanging fruit to just do it here.   I have tried to integrate it already, and am still cleaning up stuff here and there (like **kwargs, and some flux routine namings), but your eyes on the changes would be terrific when you get a chance @majosm.

mirgecom/inviscid.py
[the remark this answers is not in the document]
MTCam: Gone :)

mirgecom/flux.py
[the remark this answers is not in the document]
MTCam: Took a stab.  How's it looking now?

doc/discretization.rst
```diff
@@ -132,96 +146,413 @@ $$
...
 
-Solid walls
-^^^^^^^^^^^
 
-Inviscid boundary flux
-""""""""""""""""""""""
-$\b{h}^*_e$ is equal to the (interior; - side) pressure contribution of
-$\b{F}^I(\b{Q}_{bc})\cdot\b{n}$
-(since $\b{V}\cdot\b{n} = 0$).
+Adiabtic slip wall
```
inducer, line 278: Read to here

doc/discretization.rst
```diff
@@ -132,96 +146,413 @@ $$
...
-or by prescribing one or more components of the boundary fluxes $\b{h}^*_e$, $\b{h}^*_v$,
-and $\b{H}^*_s$.  Descriptions of particular boundary treatments follow in the next few
-sections.
+Boundary conditions ($\b{Q}_{bc}, \b{\Sigma}_{bc}$) are set by prescribing one or more
+components of the solution or its gradient on the (+) side of the boundary,
+($\b{Q}^+, \b{\Sigma}^+$), respectively, or by prescribing one or more components of the
+boundary fluxes $\b{h}^*_e$, $\b{h}^*_v$, and $\b{H}^*_s$.  Descriptions of particular
+boundary treatments follow in the next few sections.
 
+.. image:: figures/ElementBoundary.png
```
inducer, line 272: Track down the source file?

inducer: https://sphinxcontrib-tikz.readthedocs.io/en/latest/

mirgecom/viscous.py
```diff
@@ -275,63 +277,170 @@ def viscous_flux(state, grad_cv, grad_t):
             momentum=tau, species_mass=-j)
 
 
-def viscous_facial_flux(discr, state_tpair, grad_cv_tpair, grad_t_tpair,
-                        local=False):
-    """Return the viscous flux across a face given the solution on both sides.
+def viscous_divergence_flux(discr, state_pair, grad_cv_pair, grad_t_pair,
+                         beta=0., gamma=0, **kwargs):
```
inducer, line 281: kwargs?

mirgecom/viscous.py
```diff
@@ -275,63 +277,170 @@ def viscous_flux(state, grad_cv, grad_t):
...
+
+    return divergence_flux(trace_pair=f_pair, normal=normal,
+                           alpha=jump_term, beta=beta)
+
+
+def viscous_flux_on_element_boundary(
+        discr, gas_model, boundaries, interior_state_pairs,
+        domain_boundary_states, grad_cv, interior_grad_cv_pairs,
+        grad_t, interior_grad_t_pairs, quadrature_tag=None,
+        numerical_flux_func=viscous_divergence_flux, time=0.0):
```
inducer, line 343:
```suggestion
        numerical_flux_func=viscous_divergence_flux):
```

mirgecom/viscous.py
```diff
@@ -275,63 +277,170 @@ def viscous_flux(state, grad_cv, grad_t):
...
+    jump_term = -gamma*outer(q_pair.diff, normal)/2
+
+    return divergence_flux(trace_pair=f_pair, normal=normal,
+                           alpha=jump_term, beta=beta)
+
+
+def viscous_flux_on_element_boundary(
+        discr, gas_model, boundaries, interior_state_pairs,
+        domain_boundary_states, grad_cv, interior_grad_cv_pairs,
+        grad_t, interior_grad_t_pairs, quadrature_tag=None,
```
inducer, line 342:
```suggestion
        grad_t, interior_grad_t_pairs, discretization_tag=None,
```
(here and elsewhere)

inducer: Bump to issue? Document to be a discretization tag?

mirgecom/viscous.py
MTCam: I couldn't figure out how to pass time down through the operator to the boundary without this argument.

doc/discretization.rst
MTCam: > Track down the source file?

I have it, if it is useful.  PPTX

mirgecom/flux.py
```diff
@@ -35,179 +40,224 @@
...
+    dissipation to the otherwise central flux.
 
-        lambda parameter for Lax-Friedrichs/Rusanov flux
+    Parameters
+    ----------
+    trace_pair: :class:`~grudge.trace_pair.TracePair`
+        Trace pair for the face upon which flux calculation is to be performed
+    normal: numpy.ndarray
+        object array of :class:`~meshmode.dof_array.DOFArray` with outward-pointing
+        normals
```
majosm, line 254: Document `alpha` and `beta` parameters.

mirgecom/flux.py
```diff
@@ -35,179 +40,224 @@
...
+    The divergence flux, $h$, is calculated as:
 
-    normal: numpy.ndarray
+    .. math::
 
-        object array of :class:`~meshmode.dof_array.DOFArray` with outward-pointing
-        normals
+        h(\mathbf{v}^-, \mathbf{v}^+; \mathbf{n}) = \frac{1}{2}
+        \left(\mathbf{v}^{+}+\mathbf{v}^{-}+\beta(\mathbf{v}^{+}-\mathbf{v}^{-}
+        +\alpha\right) \cdot \hat{n}
```
majosm, line 241: Is there a use case for a nonzero `alpha` as defined here? Seems like it would add a directional bias to the flux. (_Edit: I see what it's doing now; see other comment below._)

mirgecom/flux.py
```diff
@@ -35,179 +40,224 @@
...
+        normals
 
     Returns
     -------
     numpy.ndarray
-
-        object array of :class:`~meshmode.dof_array.DOFArray` with the
-        Lax-Friedrichs/Rusanov flux.
+        object array of :class:`~meshmode.dof_array.DOFArray` with the flux for each
+        scalar component.
```
majosm, line 260: This doesn't look quite right. Maybe we should be intentionally vague here since the function is generic and the result type depends on what's inside the trace pair?

mirgecom/flux.py
```diff
@@ -35,179 +40,224 @@
...
+    """
+    from arraycontext import outer
+    return outer(u_tpair.avg + .5*beta*u_tpair.diff, normal)
 
-        Solution trace pair for faces for which numerical flux is to be calculated
 
-    f_tpair: :class:`~grudge.trace_pair.TracePair`
+def divergence_flux(trace_pair, normal, alpha=0, beta=0):
```
majosm, line 232: I'm not sure if I like these generic `gradient_flux`/`divergence_flux` functions overall. They seem both overly-general and at the same time not general enough (e.g., what if the quantity being penalized is not the same as the quantity used in the central part?). I wonder if it's better to keep the implementations simple (`gradient_flux_central`, `divergence_flux_central`, etc.), and just add more of them when we need different flux formulations.

mirgecom/viscous.py
```diff
@@ -275,63 +277,170 @@ def viscous_flux(state, grad_cv, grad_t):
             momentum=tau, species_mass=-j)
 
 
-def viscous_facial_flux(discr, state_tpair, grad_cv_tpair, grad_t_tpair,
-                        local=False):
-    """Return the viscous flux across a face given the solution on both sides.
+def viscous_divergence_flux(discr, state_pair, grad_cv_pair, grad_t_pair,
```
majosm, line 280:
```suggestion
def viscous_facial_flux(discr, state_pair, grad_cv_pair, grad_t_pair,
```
(for symmetry with inviscid.)

Also, maybe hedge and call it `viscous_facial_flux_central`? (I thought it was already called this at one point...?) Unless we're sure we're not going to add any more viscous flux types (in which we case we don't need the `numerical_flux_func` argument in `viscous_flux_on_element_boundary`).

mirgecom/viscous.py
```diff
@@ -275,63 +277,170 @@ def viscous_flux(state, grad_cv, grad_t):
...
+    f_ext = viscous_flux(state_pair.ext, grad_cv_pair.ext,
+                         grad_t_pair.ext)
+    f_pair = TracePair(state_pair.dd, interior=f_int, exterior=f_ext)
+    q_pair = TracePair(state_pair.dd, interior=state_pair.int.cv,
+                       exterior=state_pair.ext.cv)
+    from arraycontext import outer
+    jump_term = -gamma*outer(q_pair.diff, normal)/2
+
+    return divergence_flux(trace_pair=f_pair, normal=normal,
+                           alpha=jump_term, beta=beta)
```
majosm, line 336: Ah, now I see how `alpha` is being used. Not really a fan of having an extra catch-all parameter to `divergence_flux` like that. If the form of the flux is significantly different from a standard central flux, it should probably be its own separate flux function.

Also, if the intent of `gamma` is to support an interior penalty method, keep in mind that the form of the penalty coefficient may be complicated and not really something you want to expose in the public viscous flux interface. Might be best to leave this out for now and add a separate `viscous_facial_flux_penalty` or something later.

mirgecom/viscous.py
```diff
@@ -275,63 +277,170 @@ def viscous_flux(state, grad_cv, grad_t):
...
+        discr, gas_model, boundaries, interior_state_pairs,
+        domain_boundary_states, grad_cv, interior_grad_cv_pairs,
+        grad_t, interior_grad_t_pairs, quadrature_tag=None,
+        numerical_flux_func=viscous_divergence_flux, time=0.0):
+    """Compute the inviscid boundary fluxes for the divergence operator.
+
+    This routine encapsulates the computation of the inviscid contributions
+    to the boundary fluxes for use by the divergence operator. Its existence
+    is intended to allow multiple operators (e.g. Euler and Navier-Stokes) to
+    perform the computation without duplicating code.
```
majosm, line 349: Update for viscous.

majosm (commented):
Made another pass through and found some stuff (mostly doc fixes).

mirgecom/viscous.py
majosm: I think we can make it a little more obvious that this is just a passthrough to the BCs (e.g. by passing something like a `boundary_kwargs` argument instead of passing `time` etc. explicitly), but I suspect this is best saved for another PR.

doc/discretization.rst
```diff
@@ -132,96 +146,413 @@ $$
...
-We denote the *boundary fluxes* as $\b{h}^*_e(\b{Q}_{bc})$,
-$\b{h}^*_v(\b{Q}_{bc}$, $\b{\Sigma}_{bc})$, and
-$\b{H}^*_s(\b{Q}_{bc})$, where $\b{Q}_{bc}$, $\b{\Sigma}_{bc}$ denote
-boundary conditions imposed on the state, and the gradient of the state respectively.
+In DG, fluxes are responsible for handling the flow of information
+between adjacent cells, and for transferring the boundary conditions
+into the domain from the domain boundary.  In this sense, all of the
+boundary conditions are *weak*, in that they are weakly enforced
+through the flux, as opposed to directly setting the solution of the
+boundary elements (a *strong* enforcement).
```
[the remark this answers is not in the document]
MTCam: This point is covered by the paper that we ended up using to implement the BCs.  I think we are talking about the same thing - SAT is just another approach of weak boundary conditions.  I think they are weak by virtue of not prescribing the BC directly, but by providing the _flux_ that drives the soln to the BC.

mirgecom/viscous.py
MTCam: Can we stash time somewhere in the fluid state?  That'd be great? 🤔

mirgecom/viscous.py
MTCam: Made an issue, but eager to tackle it here if we can.

doc/discretization.rst
```diff
@@ -121,8 +123,20 @@ $$
...
+a single valued flux that both elements agree on.  That is, they must be functions
+of both $\pm$ states, and must produce a consistent flux. In general, numerical
+fluxes for the divergence operator must satisfy the consistency relations:
+
+.. math::
+   \b{h}(\b{Q}, \b{Q}; \b{n}) = \b{F}(\b{Q})\cdot\b{n}~~~~~~
+   \b{h}(\b{Q}^+,\b{Q}^-;\b{n}) = -\b{h}(\b{Q}^-, \b{Q}^+;-\b{n}),
+
+where $\b{h}$ is the numerical flux function, and $\b{F}$ is the flux function at
+the element boundary the numerical flux function is intended to replace.
```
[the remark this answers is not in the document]
MTCam: I'd be happy to if I could find them stated somewhere.  The only place I've seen these consistency relations stated, they are in ref. to the discussion of the Riemann solver based fluxes.

mirgecom/flux.py
MTCam: They are just equation (13) of Mengaldo.

mirgecom/viscous.py
MTCam: These were just based on the fluxes Mengaldo used.  With the default parameters, they end up being central like we have been using.

doc/discretization.rst
```diff
@@ -132,96 +146,413 @@ $$
 \b{H}_s(\b{Q}^+_h, \b{Q}_h^-; \b{n}) &\approx \b{Q}_h\b{n}.
 $$
 
-Choices of numerical fluxes corresponding to BR1
-------------------------------------------------
+The $\b{h}_e$, and $\b{h}_v$ numerical flux functions are responsible for
+calculating the numerical fluxes for the divergence of the inviscid and viscous
+transport fluxes, respectively.  The $\b{H}_s$ numerical flux function is
+responsible for calculating the numerical flux for the gradient operator, and
+so is employed, for example, in the auxiliary equation.
```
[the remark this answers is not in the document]
MTCam: Are you sure?  Can I reword it to make it better for you?  This is the only place where the actual notation for the numerical fluxes used in the rest of the document are actually connected with the physical flux terms they are intended to replace.

doc/discretization.rst
```diff
@@ -132,96 +146,413 @@ $$
...
+the boundary treatment in |mirgecom| is to prescribe the boundary solution $\b{Q}^+$ to be
+used in the numerical flux function to *induce* the desired boundary condition, $\b{Q}_{bc}$.
+
+The adiabatic slip wall boundary treatment is implemented by the
+:class:`~mirgecom.boundary.AdiabaticSlipBoundary`. The boundary solution
+is prescribed as follows:
+
+.. math::
+
+   \b{Q}^+ = \b{Q}^- - 2*\left(\rho\b{v}^-\cdot\hat{\b{n}}\right)\hat{\b{n}},
```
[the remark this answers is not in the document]
MTCam: Fixed.

mirgecom/viscous.py
MTCam: Yes we have planned on using different viscous fluxes - esp when we were examining the BCs closely and matching our implementation with Mengaldo.

mirgecom/boundary.py
```diff
@@ -100,29 +95,29 @@ def _dummy_temperature_func(self, temperature_minus, **kwargs):
...
 
-    def _boundary_quantity(self, discr, btag, quantity, **kwargs):
+    def _boundary_quantity(self, discr, btag, quantity, local=False, **kwargs):
         """Get a boundary quantity on local boundary, or projected to "all_faces"."""
-        if "local" in kwargs:
-            if kwargs["local"]:
-                return quantity
-        return discr.project(btag, btag.with_dtag("all_faces"), quantity)
+        return quantity if local else discr.project(
+            btag, btag.with_dtag("all_faces"), quantity)
```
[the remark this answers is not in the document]
MTCam: AV still uses it.

doc/discretization.rst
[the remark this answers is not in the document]
MTCam: I re-worded this.  Can you take another look?

doc/discretization.rst
```diff
@@ -132,96 +146,413 @@ $$
...
+   \b{h}^*_v(\b{Q}_{bc}, \b{\Sigma}_{bc}) = \b{h}_v\left(\b{Q}^-,\b{\Sigma}^-,\b{Q}_{bc},\b{\Sigma}_{bc}\right),
+
+where $\b{Q}_{bc}$ are the same values used to prescribe $\b{h}^*_e$.
+
+In *MIRGE-Com*, we use the central flux to transfer viscous BCs to the domain:
+
+.. math::
+
+     \b{h}^*_v(\b{Q}_{bc}, \b{\Sigma}_{bc}) = \frac{1}{2}\left(\mathbf{F}_v(\mathbf{Q}_{bc},\mathbf{\Sigma}_{bc}) + \mathbf{F}_v(\mathbf{Q}^{-},\mathbf{\Sigma}^{-})\right)
+
```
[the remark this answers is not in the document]
MTCam: You are correct.  We implemented _new_ boundary conditions to match Mengaldo.  This document describes what is done in the BCs that are in the code at this point (i.e., in _this_ PR).  I'll update the docs to the new BCs in a separate PR.

doc/discretization.rst
```diff
@@ -132,96 +146,413 @@ $$
...
+   
+  \b{Q}^{\pm} &\equiv \text{conserved quantities on the exterior/interior of the boundary face} \\
+  \b{Q}_{bc} &\equiv \text{boundary condition for the fluid conserved quantities} \\
+  \b{\Sigma}^{\pm} &\equiv \text{gradient of conserved quantities on ext/int of boundary face} \\
+  \b{\Sigma}_{bc} &\equiv \text{boundary condition for grad of soln } \\
+  \b{v}^{\pm} &\equiv \text{Flow velocity on the ext/int of boundary face} \\
+  \b{h}^*_e &\equiv \text{boundary flux for the divergence of inviscid flux} \\
+  \b{h}^*_v &\equiv \text{boundary flux for divergence of viscous flux} \\
+  \b{H}_s^* &\equiv \text{boundary flux for the gradient of the conserved quantities} \\
+  \hat{\b{n}} &\equiv \text{outward pointing normal for the boundary face}
```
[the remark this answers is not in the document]
majosm: Maybe something like:

```
$\b{Q}^{\pm}$:
    conserved quantities on the exterior/interior of the boundary face
$\b{Q}_{bc}$:
    boundary condition for the fluid conserved quantities
$\b{\Sigma}^{\pm}$:
    gradient of conserved quantities on ext/int of boundary face
$\b{\Sigma}_{bc}$:
    boundary condition for grad of soln
$\b{v}^{\pm}$:
    flow velocity on the ext/int of boundary face
$\b{h}^*_e$:
    boundary flux for the divergence of inviscid flux
$\b{h}^*_v$:
    boundary flux for divergence of viscous flux
$\b{H}_s^*$:
    boundary flux for the gradient of the conserved quantities
$\hat{\b{n}}$:
    outward pointing normal for the boundary face
```

Screen Shot 2022-05-10 at 10 52 39 AM
?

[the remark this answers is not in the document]
MTCam: I found `definition list` in RST format which rendered almost identical to above.  Whadda ya think?

doc/discretization.rst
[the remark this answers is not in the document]
majosm: Hmm, unfortunately I don't know them either. Do you have a link to where you found the ones for div? Maybe we can derive them analogously.

My guess is they should be the same for the gradient (with dot replaced by multiplication). AFAICT the first ensures that the numerical flux agrees with the physical flux when the two sides agree on the state, and the second ensures that the surface integrals computed by adjacent faces agree (up to a minus sign due to difference in orientation).

doc/discretization.rst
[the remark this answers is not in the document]
majosm: This paragraph in the previous section seems sufficient to me:

Screen Shot 2022-05-10 at 12 15 53 PM

doc/discretization.rst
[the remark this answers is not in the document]
majosm: 🤷‍♂️ I guess I can live with it. I'm not really convinced yet that what the Mengaldo paper describes as "Weak-Prescribed" is any different numerically from strong enforcement... but I would have to read the paper more carefully to say for sure.

mirgecom/flux.py
majosm: Almost -- I would get rid of the `numpy.ndarray` result type (since that's likely not what it will be) and the object array language in the description. Maybe just say something along the lines of `object of the same type as *trace_pair.int* containing the face-normal fluxes for each component`?

mirgecom/viscous.py
majosm: I think my main issues with these are:

1. The names are too general (LFR and HLL are also divergence fluxes, after all)
2. The `alpha` parameter in `divergence_flux` is too open-ended.

If we moved the `jump_term` stuff inside the divergence flux function and found some better names, I think I would be OK with them.

What do you think about integrating these functions into the lower-level flux interface so they're similar to `num_flux_hll`, etc.? Something like https://github.com/majosm/mirgecom/commit/22d1af3714b20f0a468c1d403169a649bbecfe50.

MTCam: Sure this is just fine with me.

doc/discretization.rst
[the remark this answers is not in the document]
MTCam: Those are the internal face numerical flux functions.  The fluxes under discussion are those used at the domain boundary.  They are not the same functions, in general, and are documented as separate flux treatments in both BR1 and Mengaldo references.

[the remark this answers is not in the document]
MTCam: Removed.

doc/discretization.rst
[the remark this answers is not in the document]
MTCam: Eqn 10 of the [BR1 reference](https://www.sciencedirect.com/science/article/pii/S0021999196955722?via%3Dihub)
There's some discussion of the underpinning of this in 6.6.2 of my HW, Revisiting Numerical Fluxes or some such.

mirgecom/viscous.py
MTCam: Actually, can we please keep the interface naming such that it is clear what operator (divergence) vs. (gradient) the flux is intended for?   For the former the return is a scalar for each eqn and for the latter it is a vector quantity for each eqn.

MTCam: ok, let's do that when we add new fluxes.  for now i just removed all the fluxes that you having a problem with.  can we push on without them for now?

doc/discretization.rst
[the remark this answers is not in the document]
majosm: Meh, I'm not sure if it's worth the trouble. I think I'll just mark this as resolved and possibly suggest a minor change in wording below.

doc/discretization.rst
```diff
@@ -121,128 +123,414 @@ $$
 Numerical fluxes
 ================
 
-To account for the discontinuities at element faces, DG employs numerical fluxes, which are
-enforced to be singled-valued, but are functions of both $\pm$ states:
+Numerical fluxes are responsible for calculating the fluxes at the DG element boundaries.  
+Numerical fluxes must account for the discontinuities at element faces, and calculate
+a single valued flux that both elements agree on.  That is, they must be functions
+of both $\pm$ states, and must produce a consistent flux. In general, numerical
+fluxes for the divergence operator must satisfy the consistency relations:
```
majosm, line 130:
```suggestion
of both $\pm$ states, and must produce a consistent flux.

For a conservation law $\frac{\partial \b{Q}}{\partial t} + \nabla \cdot \b{F}(\b{Q}) = \b{S}$,
the numerical flux $h$ must satisfy the consistency relations
```

doc/discretization.rst
```diff
@@ -121,128 +123,414 @@ $$
...
-$$
-\b{h}_e(\b{Q}_h^+, \b{Q}^-_h; \b{n}) &\approx \b{F}^I(\b{Q}_h)
-\cdot\b{n}, \\
-\b{h}_v(\b{Q}_h^+, \b{\Sigma}_h^+, \b{Q}_h^-, \b{\Sigma}_h^-;
-\b{n}) &\approx \b{F}^V(\b{Q}_h, \b{\Sigma}_h)\cdot\b{n}\\
-\b{H}_s(\b{Q}^+_h, \b{Q}_h^-; \b{n}) &\approx \b{Q}_h\b{n}.
-$$
+.. math::
+   h(\b{Q}, \b{Q}; \b{n}) = \b{F}(\b{Q})\cdot\b{n}~~~~~~
+   h(\b{Q}^+,\b{Q}^-;\b{n}) = -h(\b{Q}^-, \b{Q}^+;-\b{n}),
```
majosm, line 134:
```suggestion
   h(\b{Q}^+,\b{Q}^-;\b{n}) = -h(\b{Q}^-, \b{Q}^+;-\b{n}).
```

doc/discretization.rst
```diff
@@ -121,128 +123,414 @@ $$
...
 
-Choices of numerical fluxes corresponding to BR1
-------------------------------------------------
+where $h$ is the numerical flux function, and $\b{F}$ is the flux function at
+the element boundary the numerical flux function is intended to replace.
```
majosm, line 137:
```suggestion
```

majosm (commented):
Github made me split this into 3 suggestions, ugh sorry.

majosm (approved):
Just had another pass through; looks good on my end. Left a couple of optional thoughts.

doc/discretization.rst
```diff
@@ -121,128 +123,414 @@ $$
...
-or by prescribing one or more components of the boundary fluxes $\b{h}^*_e$, $\b{h}^*_v$,
-and $\b{H}^*_s$.  Descriptions of particular boundary treatments follow in the next few
-sections.
+Boundary conditions ($\b{Q}_{bc}, \b{\Sigma}_{bc}$) are set by prescribing one or more
+components of the solution or its gradient on the (+) side of the boundary,
+($\b{Q}^+, \b{\Sigma}^+$), respectively, or by prescribing one or more components of the
+boundary fluxes $h^*_e$, $h^*_v$, and $\b{H}^*_s$.  Descriptions of particular
+boundary treatments follow in the next few sections.
 
+.. image:: figures/ElementBoundary.png
```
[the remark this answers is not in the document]
MTCam: fwiw, we intend to replace this outright with tikz.

MTCam:
Anything preventing merge here from your perspective, @inducer?

doc/discretization.rst
inducer: #667

doc/discretization.rst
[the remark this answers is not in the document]
MTCam: #667

mirgecom/inviscid.py
```diff
@@ -128,32 +176,124 @@ def inviscid_facial_flux(discr, state_tpair, local=False):
...
+    # left (internal), and right (external) wave speed estimates
+    # can alternatively use the roe estimated states to find the wave speeds
+    s_minus = u_int - c_int*q_int_rt
+    s_plus = u_ext + c_ext*q_ext_rt
+
+    f_minus_normal = inviscid_flux(state_pair.int)@normal
+    f_plus_normal = inviscid_flux(state_pair.ext)@normal
+
+    q_minus = state_pair.int.cv
+    q_plus = state_pair.ext.cv
```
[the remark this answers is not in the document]
anderson2981: those expressions follow almost verbatim from Toro. I'm not sure we need extra comments here.

anderson2981 (approved):
This passes all the tests I setup initially to verify its implementation from a physics perspective. I don't see any changes that would impeded its proper operation. And the tests are still passing.

MTCam:
Resolved part of #509

PR #652, opened 2022-04-29
Use a decorator to handle inviscid/viscous flux boilerplate

majosm:
**Questions for the review**:

### DIFF
--- a/mirgecom/inviscid.py
+++ b/mirgecom/inviscid.py
@@ -6,7 +6,6 @@ Inviscid Flux Calculation
 .. autofunction:: inviscid_flux
 .. autofunction:: inviscid_facial_flux_rusanov
 .. autofunction:: inviscid_facial_flux_hll
-.. autofunction:: inviscid_facial_flux
 .. autofunction:: inviscid_flux_on_element_boundary
 
 Inviscid Time Step Computation
@@ -85,7 +84,7 @@ def inviscid_flux(state):
                           momentum=mom_flux, species_mass=species_mass_flux)
 
 
-def inviscid_facial_flux_rusanov(state_pair, gas_model, normal, **kwargs):
+def _rusanov_facial_flux(state_pair, gas_model, normal, **kwargs):
     r"""High-level interface for inviscid facial flux using Rusanov numerical flux.
 
     The Rusanov or Local Lax-Friedrichs (LLF) inviscid numerical flux is calculated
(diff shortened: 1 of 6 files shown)


mirgecom/inviscid.py
```diff
@@ -198,13 +172,22 @@ def inviscid_facial_flux_hll(state_pair, gas_model, normal, **kwargs):
...
-    r"""Return the numerical inviscid flux for the divergence operator.
+def _as_facial_flux(flux_snippet_func, name):
+    """Promote a facial flux snippet to a complete inviscid facial flux function."""
+    def wrapped_func(
+            discr, gas_model, state_pair, local=False, **kwargs):
+        normal = thaw(discr.normal(state_pair.dd), state_pair.int.array_context)
+        num_flux = flux_snippet_func(state_pair, gas_model, normal, **kwargs)
+        dd = state_pair.dd
+        dd_allfaces = dd.with_dtag("all_faces")
+        return num_flux if local else discr.project(dd, dd_allfaces, num_flux)
```
inducer, line 183: That `local` leaves a bad taste in my mouth. Should we deprecate the `local=False` version (and explicitly do the projection outside)?

majosm: @MTCam What do you think about this? If we made the facial flux functions always return the flux on the same discretization as the input, it would more or less eliminate the need for this extra processing/decorator stuff (all that would be left is retrieving the normal from `discr`, which I think would be OK to move into the flux implementation functions). We could do the `"all_faces"` projections inside `inviscid_flux_on_element_boundary` instead.

MTCam: The reason that the `local` argument exists is for use down in the boundary conditions where sometimes you want the flux coming back for the local boundary surface only, not projected to "all faces".  I'm not a giant fan of automatically always projecting to "all faces" when that's not where the thing we are computing actually lives.

inducer: So could we move the project to "all faces" out of these functions, if only some of its users seem to need it?

MTCam: Oh, sure.  I am not opposed to moving it or doing it another way - just so long as we can get it locally when we need.

majosm: How about something like https://github.com/majosm/mirgecom/commit/56b91bbafad82a8374ed865792ded385526f2702? (Just a sketch, not fully implemented.) Basically, let the interior flux functions and the BCs return fluxes on their respective `dd`, and then the calling function can project to `"all_faces"` if it wants.

MTCam: If it means it needs to be projected it to all faces down inside the operator (e.g., ns_operator), then I am not so keen on that.   If the _caller_ in this case means an intermediate layer where the details of projection to/from "all_faces" and quadrature domain, etc, then I think that'd be ok.

MTCam: In the boundary conditions I guess?  There's a _boundary_quantity utility there that handles it.

majosm: > In the boundary conditions I guess? There's a _boundary_quantity utility there that handles it.

(IIRC we talked about this on Slack after you posted this, but for completeness: I think the change in https://github.com/majosm/mirgecom/commit/9176e74aff79cc39d5f9d3cdb7f5d8cf59424c68 should do away with the need for the projection step inside the boundary conditions.)

I like that newer approach better than the one I originally proposed in this PR, so if there aren't any objections I think I'll close this and we can go back to #648.

PR #677, opened 2022-05-27
Pre-generated thermochem

MTCam:
This change set adds the capability to use pre-generated thermochemical mechanisms.   

- Allows mechanisms from development version of Cantera required for adding mixture transport model
- Allows drivers to evict Cantera from runtime requirements (key in improving / enabling production driver compiles and reasonable runtimes)
- Imports actual code for mechanisms - allowing easy inspection by users

**Questions for the review**:

### DIFF
--- a/.pylintrc.yml
+++ b/.pylintrc.yml
@@ -4,3 +4,5 @@
   val: all
 - arg: enable
   val: E
+- arg: ignore
+  val: mirgecom/mechanisms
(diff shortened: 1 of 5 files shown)


anderson2981 (approved):
This looks great. It allows us to skip cantera altogether, which can be a nuisance at best or a complete show stopper.

PR #678, opened 2022-05-28
Add Mengaldo BCs implementation

MTCam:
This is a draft change to add BCs from the [Mengaldo reference.](https://arc.aiaa.org/doi/10.2514/6.2014-2923)

- Fixes all solid-wall viscous boundary treatments per @anderson2981 bug fix
- BCs that implement (instead of inherit default) viscous flux routines that are consistent with Mengaldo
- Adds farfield and in/out flow boundaries from Mengaldo with @tulioricci bug fix for outflow

**Questions for @lukeolson for the review**:
  - [x] Discusssed. Sufficient for now.
  - [x] Create issue on symmetry boundary condition to remove or remove now to check this off
  	- Removed forever :)


mirgecom/boundary.py
```diff
@@ -602,3 +689,628 @@ def temperature_bc(self, state_minus, **kwargs):
...
+        boundary_vel = np.dot(state_minus.velocity, nhat)*nhat
+        boundary_speed = actx.np.sqrt(np.dot(boundary_vel, boundary_vel))
+        speed_of_sound = state_minus.speed_of_sound
+        kinetic_energy = gas_model.eos.kinetic_energy(state_minus.cv)
+        gamma = gas_model.eos.gamma(state_minus.cv, state_minus.temperature)
+        external_pressure = 2*self._pressure - state_minus.pressure
+        boundary_pressure = actx.np.where(actx.np.greater(boundary_speed,
+                                                          speed_of_sound),
+                                          state_minus.pressure, external_pressure)
+        internal_energy = boundary_pressure / (gamma - 1)
```
tulioricci, line 828: This line only works for IdealGas. For mixture, it is necessary to use "gas_model.eos.internal_energy"

mirgecom/boundary.py
```diff
@@ -602,3 +689,628 @@ def temperature_bc(self, state_minus, **kwargs):
...
+            state_minus.dim, mass=state_minus.mass_density,
+            energy=state_minus.energy_density, momentum=mom_plus,
+            species_mass=state_minus.species_mass_density
+        )
+        return make_fluid_state(cv=cv_plus, gas_model=gas_model,
+                                temperature_seed=state_minus.temperature)
+
+    def adiabatic_wall_state_for_diffusion(self, discr, btag, gas_model,
+                                           state_minus, **kwargs):
+        """Return state with 0 velocities and energy(Twall)."""
```
tulioricci, line 1220: """Return state with opposite normal momentum."""

MTCam: For the inviscid part of the boundary treatment we return the opposite normal momentum because we use a Riemann solver to to calculate the fluxes.   We set the BC directly for the viscous part because we just calculate the viscous flux using the BC state, not an average or Riemann solver.

mirgecom/boundary.py
MTCam: I have not made this change yet because the boundary routines seem inconsistent with the reference we are using.   The ideal vs. mixture part is OK; it's the part where we seem to prescribe a "pressure" for the outflow that's bugging me.

tulioricci: Fixed with https://github.com/illinois-ceesd/mirgecom/pull/705

mirgecom/boundary.py
tulioricci: Fixed with https://github.com/illinois-ceesd/mirgecom/pull/705

mirgecom/boundary.py
```diff
@@ -653,7 +677,790 @@ def isothermal_noslip_state(self, discr, btag, gas_model, state_minus, **kwargs)
...
+            momentum=free_stream_density*free_stream_velocity,
+            species_mass=free_stream_spec_mass
+        )
+
+        return make_fluid_state(cv=cv_infinity, gas_model=gas_model,
+                                temperature_seed=free_stream_temperature)
+
+    def temperature_bc(self, state_minus, **kwargs):
+        """Return farfield temperature for use in grad(temperature)."""
+        return 0*state_minus.temperature + self._temperature
```
tulioricci, line 797: I am unsure this will work when "free_stream_temperature" is initially None sine here (https://github.com/illinois-ceesd/mirgecom/pull/678/files#diff-af7fe103044e2f2c2fb594902042f0c99e7c8221c6e94d0310bfed735ab96823R763) you are attributing the computed value to "free_stream_temperature" and not "self._temperature"

mirgecom/boundary.py
```diff
@@ -653,7 +677,790 @@ def isothermal_noslip_state(self, discr, btag, gas_model, state_minus, **kwargs)
...
+        wall_state = make_fluid_state(cv=wall_cv, gas_model=gas_model,
+                                      temperature_seed=state_minus.temperature)
+        state_pair = TracePair(btag, interior=state_minus, exterior=wall_state)
+
+        normal = state_minus.array_context.thaw(discr.normal(btag))
+        return numerical_flux_func(state_pair, gas_model, normal)
+
+    def temperature_bc(self, state_minus, **kwargs):
+        """Get temperature value used in grad(T)."""
+        # return 2*self._wall_temp - state_minus.temperature
```
tulioricci, line 1093: Remove the commented line

mirgecom/boundary.py
MTCam: We cannot set `self._temperature` down in this routine.   That one must be set in `__init__` if anywhere.   We may need some conditional processing in this routine to resolve the issue - but that's going to be tricky because we don't have access to the `gas_model` here.

tulioricci: Can you just replace "self._temperature" here with "free_stream_temperature" ? It is the output from both "is None" and "is not None" checks.

MTCam: @tulioricci see what you think of the farfield now

tulioricci: Looks good to me.

mirgecom/boundary.py
```diff
@@ -666,7 +882,1095 @@ def isothermal_noslip_state(
     def temperature_bc(self, state_minus, **kwargs):
         r"""Get temperature value to weakly prescribe wall bc.
 
-        Returns $2*T_\text{wall} - T^-$ so that a central gradient flux
+        Returns $2T_\text{wall} - T^-$ so that a central gradient flux
         will get the correct $T_\text{wall}$ BC.
         """
         return 2*self._wall_temp - state_minus.temperature
```
MTCam, line 888: This might not be correct, or at least not consistent with (Mengaldo) [https://arc.aiaa.org/doi/10.2514/6.2014-2923], eqn. 50. and paragraph that follows.  I think the `temperature_bc` function used for the BC on `grad(T)` should send back `Tplus = Twall`, not `Tplus = 2Twall - Tminus` as it currently does.  It can result in negative `Tplus`.  Mengaldo's paper says to send `Twall*n` for the flux used in the grad(T) instead of using a numerical flux function, if I have the right read.

mirgecom/boundary.py
```diff
@@ -301,8 +373,8 @@ def __init__(self,
         if not self._bnd_temperature_func:
             self._bnd_temperature_func = self._temperature_for_prescribed_state
         if not self._grad_num_flux_func:
-            self._grad_num_flux_func = num_flux_central
-
+            # self._grad_num_flux_func = num_flux_central
+            self._grad_num_flux_func = _ldg_bnd_flux_for_grad
```
majosm, line 377: Should this be applied universally like this? Or only to the Mengaldo BCs?

mirgecom/boundary.py
```diff
@@ -666,7 +882,1095 @@ def isothermal_noslip_state(
...
+        wall_state = self.adiabatic_wall_state_for_advection(
+            dcoll, dd_bdry, gas_model, state_minus)
+        state_pair = TracePair(dd_bdry, interior=state_minus, exterior=wall_state)
+
+        normal = state_minus.array_context.thaw(dcoll.normal(dd_bdry))
+        return numerical_flux_func(state_pair, gas_model, normal)
+
+    def temperature_bc(self, state_minus, **kwargs):
+        """Get temperature value used in grad(T)."""
+        return state_minus.temperature
```
majosm, line 1547: Can be omitted I think.

mirgecom/boundary.py
```diff
@@ -502,67 +578,193 @@ def __init__(self):
...
+
+    def grad_temperature_bc(self, grad_t_minus, normal, **kwargs):
+        """
+        Compute temperature gradient on the plus state.
+
+        Impose the opposite normal component to enforce zero energy flux
+        from conduction.
+        """
+        return (grad_t_minus
+                - 2.*np.dot(grad_t_minus, normal)*normal)
```
majosm, line 686:
```suggestion
        return (grad_t_minus
                - np.dot(grad_t_minus, normal)*normal)
```
?

majosm (commented):
Added a few things I noticed while working on #835.

mirgecom/boundary.py
```diff
@@ -666,7 +882,1095 @@ def isothermal_noslip_state(
...
+        """Return state with zero-velocity."""
+        dd_bdry = as_dofdesc(dd_bdry)
+        mom_plus = 0.*state_minus.momentum_density
+        cv_plus = make_conserved(
+            state_minus.dim, mass=state_minus.mass_density,
+            energy=state_minus.energy_density, momentum=mom_plus,
+            species_mass=state_minus.species_mass_density
+        )
+        return make_fluid_state(cv=cv_plus, gas_model=gas_model,
+                                temperature_seed=state_minus.temperature)
```
majosm, line 1533: Should we be modifying the (kinetic) energy here, like we do for the isothermal no-slip? (If not, maybe add a comment explaining why for clarity.)

majosm (commented):
Forgot one.

mirgecom/boundary.py
MTCam: Yes, I think this is fixed now.

mirgecom/boundary.py
MTCam: Not sure.

mirgecom/boundary.py
MTCam: I think this is fixed.

majosm (approved):
Just noticed a couple of minor things, otherwise looks good to me!

mirgecom/boundary.py
```diff
@@ -492,181 +1051,810 @@ def av_flux(self, dcoll, dd_bdry, diffusion, **kwargs):
...
-class IsothermalNoSlipBoundary(PrescribedFluidBoundary):
-    r"""Isothermal no-slip viscous wall boundary.
+    def temperature_bc(self, state_minus, **kwargs):
+        """Return farfield temperature for use in grad(temperature)."""
+        return 0*state_minus.temperature + self._temperature
 
-    .. automethod:: isothermal_noslip_state
-    .. automethod:: temperature_bc
+
+class PressureOutflowBoundary(PrescribedFluidBoundary):
```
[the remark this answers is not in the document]
MTCam: We definitely should.  I just lazily did not mess with it this round.

mirgecom/boundary.py
```diff
@@ -162,13 +695,39 @@ def viscous_divergence_flux(self, dcoll, dd_bdry, gas_model, state_minus,
...
+        grad_cv_bc = self.grad_cv_bc(dcoll, dd_bdry, gas_model,
+                                     state_minus=state_minus,
+                                     grad_cv_minus=grad_cv_minus,
+                                     normal=normal, **kwargs)
+
+        grad_t_bc = self.grad_temperature_bc(grad_t_minus=grad_t_minus,
+                                             normal=normal, **kwargs)
+        # dcoll=dcoll, dd_bdry=dd_bdry, gas_model=gas_model,
+        # state_minus=state_minus, grad_cv_minus=grad_cv_minus,
+        # grad_t_minus=grad_t_minus)
```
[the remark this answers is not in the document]
MTCam: 👍

test/test_bc.py
```diff
@@ -57,6 +58,912 @@
...
+
+                assert actx.np.all(diff_wall_state.momentum_density[idim]
+                                   == expected_diff_momentum[idim])
+
+
+@pytest.mark.parametrize("dim", [1, 2, 3])
+@pytest.mark.parametrize("flux_func", [inviscid_facial_flux_rusanov,
+                                       inviscid_facial_flux_hll])
+def test_symmetry_wall_boundary(actx_factory, dim, flux_func):
+    """Check SymmetryBoundary boundary treatment."""
```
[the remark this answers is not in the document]
MTCam: I updated the comment.  We can merge this with AdiabaticSlipBoundary testing, but let's defer that change to an upcoming follow-on that updates the testing for the BCs.

PR #686, opened 2022-06-10
Fix `TracePair` imports now that `grudge.symbolic` is gone

inducer:
https://github.com/inducer/grudge/pull/259 deleted the symbolic execution path in grudge. Since we got rid of mirgecom downstream CI in grudge (https://github.com/inducer/grudge/pull/221/files#r819008736), I had no idea about these now-stale references to the symbolic stuff in mirgecom.

This PR mops those up.


MTCam (approved):
👍

PR #695, opened 2022-06-25
Check for valid file and file path in mirgecompare

poulNA:
**Questions for the review**:

### DIFF
--- a/bin/mirgecompare.py
+++ b/bin/mirgecompare.py
@@ -326,7 +326,6 @@ def compare_files_hdf5(first_file, second_file, tolerance=1e-12):
 if __name__ == "__main__":
     import argparse
     import os
-    import os.path
 
     # read in file and comparison info from command line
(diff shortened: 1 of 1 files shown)


matthiasdiener (commented):
This looks good, thanks Paul!
Two suggestions:
- Please move this code to the main script code below (around line 340), so that all file types handled by this script will have the same check.
- Please raise an `ArgumentError` instead of a `ValueError`

bin/mirgecompare.py
```diff
@@ -333,6 +326,7 @@ def compare_files_hdf5(first_file, second_file, tolerance=1e-12):
 if __name__ == "__main__":
     import argparse
     import os
+    import os.path
```
matthiasdiener, line 329: I think this line is not necessary

```suggestion
```

matthiasdiener:
This looks good, thanks!

matthiasdiener (approved):

PR #701, opened 2022-07-02
refactor mirgecompare.py into library + script

poulNA:
The intent of this change set is to move the functionality to simutils only.

**Questions for the review**:

@MTCam 
-----------------------
  - [x] ~~The PR should have a guide if needed (e.g., an ordering).~~
------------------------
------------------------
It would be great to have a few tests on this functionality.   Ones that demonstrate that the functionality is working as planned should be added.  Checking this box since this change set does not implement anything new - but we should still add tests for this functionality.  We use one of these comparison utils in the mirgecom testing suite, so that's at least something.

- [x] ~~Is everything that is implemented covered by tests?~~

------------------
Yes, there are design flaws, but this current work does not (meaningfully) affect the design. 


mirgecom/simutil.py
```diff
@@ -425,3 +432,391 @@ def create_parallel_grid(comm, generate_grid):
...
+                raise ValueError("Fidelity test failed: Mismatched data values "
+                                 "with given tolerance")
+
+    print("XDMF Fidelity test completed successfully with tolerance", tolerance)
+
+
+def compare_files_hdf5(first_file: str, second_file: str, tolerance: float = 1e-12):
+    """Compare files of hdf5 type.
+
+    .. note::
```
matthiasdiener, line 724: I think this line might not be necessary (same for the other functions)

```suggestion
    .. note::
```

mirgecom/simutil.py
```diff
@@ -21,6 +21,13 @@
 --------------
 
 .. autofunction:: generate_and_distribute_mesh
+
+File comparison utilities
```
matthiasdiener, line 25: @MTCam : Is this the right file for these functions?

MTCam: Good question.   I think that it will do no harm putting it here for now.  I'd also be ok with adding a new module like `solutions` or something like that for comparing solutions together.

Eventually, it would be nice to have this arranged into utils that worked like:

```python
black_file = "/path/to/black_file"
blue_file = "/path/to/blue_file"
black_soln = get_soln_from_file(black_file)
blue_file = get_soln_from_file(blue_file)
soln_errors = compare_solutions(black_soln, blue_soln)
```

That way the user (e.g., driver developer) could use these pieces to write their own comparison utilities, for example.

mirgecom/simutil.py
```diff
@@ -425,3 +432,391 @@ def create_parallel_grid(comm, generate_grid):
...
+    Parameters
+    ----------
+    first_file:
+        First file to compare
+    second_file:
+        Second file to compare
+    file_type:
+        Xdmf files
+    tolerance:
+        Max acceptable value
```
MTCam, line 602:
```suggestion
        Max acceptable absolute difference
```

mirgecom/simutil.py
```diff
@@ -425,3 +432,391 @@ def create_parallel_grid(comm, generate_grid):
...
+    Parameters
+    ----------
+    first_file:
+        First file to compare
+    second_file:
+        Second file to compare
+    file_type:
+        Hdf5 files
+    tolerance:
+        Max acceptable value
```
MTCam, line 735:
```suggestion
        Max acceptable absolute difference
```

mirgecom/simutil.py
```diff
@@ -425,3 +432,391 @@ def create_parallel_grid(comm, generate_grid):
...
+    Parameters
+    ----------
+    first_file:
+        First file to compare
+    second_file:
+        Second file to compare
+    file_type:
+        Vtu files
+    tolerance:
+        Max acceptable value
```
MTCam, line 456:
```suggestion
        Max acceptable absolute difference
```

MTCam (changes requested):
Thanks for working on this @poulNA.   

We just need a few things filled out to follow our software process.   It would be useful for me to know the scope of the current PR.   Is the goal just to move these functions inside mirgecom right now, or should we fix them up and make them better in this PR?   For example, we should not be using absolute differences here; relative error would be far superior.   Fix that now, or later?

If the intent here is just to move the functions first, then refactor them in ongoing work, then this is already looking good to me.

matthiasdiener:
> Thanks for working on this @poulNA.
> 
> We just need a few things filled out to follow our software process. It would be useful for me to know the scope of the current PR. Is the goal just to move these functions inside mirgecom right now, or should we fix them up and make them better in this PR? For example, we should not be using absolute differences here; relative error would be far superior. Fix that now, or later?
> 
> If the intent here is just to move the functions first, then refactor them in ongoing work, then this is already looking good to me.

Currently, the focus is just to refactor this in order to make future changes and additions. I think this PR is ready to, but I'm not sure how to resolve the production merge error.

MTCam:
> (...) but I'm not sure how to resolve the production merge error.

These are not the merge errors you are looking for.  :wave:  Move along.

MTCam (approved):
👍

PR #702, opened 2022-07-05
Adding local timestepping capability

tulioricci:
Adding local time stepping, where each cell has its own time step "dt" based on the local CFL number. This should speed-up convergence to steady state solution and it is NOT intended for time accurate simulations. The nomenclature follows the book from Blazek, chapter 9 (https://www.elsevier.com/books/computational-fluid-dynamics/blazek/978-0-08-099995-1).

The necessary functions were adapted from the existing "constant_cfl = True" infrastructure.

For verification of implementation (in addition to more complex cases I've been running but not including here)
- Added a Poisuille flow example to illustrate usage of local time stepping (local_dt = True).
- Added several tests in `test_time_integrators.py`
   - to assess time step error based on the expected order and 
   - to ensure the state-advancers in `steppers.py` properly step in multi-valued `local_dt` mode
- Added and sharpened documentation of fluid inviscid and viscous timestep calculation details (necessary to document *this* change set)

**Questions for the review**:

-----
@MTCam 

------
For using in `production` branch, I see none.   For getting this into main: the array context broadcasting thing, and the awkward construction in `get_sim_timestep` gives me pause. I am unsure about adding an additional example driver for what could be a simple parameter change (i.e., `local_dt=True`), but I don't have any good ideas about what to do instead.

I'll delay checking this box until we've had more discussion with @inducer and @lukeolson.  However, this will become a new production feeder branch because not having this capability in `production` is an impediment to getting center-relevant simulation work done. 

Update: use in production/prediction for about half a year has convinced me that this is ok.  There is no observable impact on performance and brings a nice capability.

------


MTCam:
I took off the production feeder label since this is reserved for the branches that merge together to make the production branch.

mirgecom/steppers.py
```diff
@@ -154,8 +173,13 @@ def _advance_state_stepper_func(rhs, timestepper, state, t_final, dt=0,
...
 
-        t += dt
-        istep += 1
+        istep = istep + 1
+        t = t + dt
+
+        if local_dt:
+            marching_loc = istep
+        else:
+            marching_loc = t
```
MTCam, line 182:
```suggestion
        marching_loc = istep if local_dt else t
```

MTCam (approved):
Looks good to me.

PR #709, opened 2022-07-11
Use `actx.freeze_thaw` instead of `force_evaluation`

majosm:
Now that `actx.freeze_thaw` exists, `force_evaluation` seems mostly redundant (except for the `None` handling for `advance_state`, but that can be worked around without too much trouble).

**Questions for the review**:


majosm:
@MTCam has persuaded me to keep `force_evaluation` around for user-facing clarity. Closing in favor of #708.

PR #712, opened 2022-07-14
[DO NOT MERGE] Sanity check tests by switching to L2 norm

majosm:
L2 norm avoids https://github.com/inducer/pyopencl/issues/583.

**Questions for the review**:

### DIFF
--- a/examples/heat-source-mpi.py
+++ b/examples/heat-source-mpi.py
@@ -180,7 +180,7 @@ def main(actx_class, ctx_factory=cl.create_some_context, use_logmgr=True,
         if logmgr:
             set_dt(logmgr, dt)
             logmgr.tick_after()
-    final_answer = actx.to_numpy(op.norm(discr, u, np.inf))
+    final_answer = actx.to_numpy(op.norm(discr, u, 2))
     resid = abs(final_answer - 0.00020620711665201585)
     if resid > 1e-15:
         raise ValueError(f"Run did not produce the expected result {resid=}")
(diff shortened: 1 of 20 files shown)


majosm:
@MTCam It looks like a few of the tests are producing NaNs (see CI). They aren't detected in main because pyopencl's `min`/`max` (used by `op.norm(..., np.inf)`) aren't handling NaNs correctly at the moment. We should look into these.

inducer:
Not sure I love this idea. I agree that we should catch NaNs, but I feel like we should fix the underlying issue in PyOpenCL instead of switching the norm to $L^2$.  $L^2$ is a weaker norm, meaning it can hide dimension-0 non-convergence up to third order (in 3D) and dimension-1 non-convergence up to second... so there's a lot of potential for non-convergence to go undetected.

majosm:
> Not sure I love this idea. I agree that we should catch NaNs, but I feel like we should fix the underlying issue in PyOpenCL instead of switching the norm to 
> L
> 2
> . 
> L
> 2
>  is a weaker norm, meaning it can hide dimension-0 non-convergence up to third order (in 3D) and dimension-1 non-convergence up to second... so there's a lot of potential for non-convergence to go undetected. Meanwhile, 
> L
> ∞
>  is sensitive to all these things.

Yeah this is a "do not merge" PR (sorry, forgot to mention that explicitly). Just wanted to run CI and see what happens.

MTCam:
> @MTCam It looks like a few of the tests are producing NaNs (see CI). They aren't detected in main because pyopencl's `min`/`max` (used by `op.norm(..., np.inf)`) aren't handling NaNs correctly at the moment. We should look into these.

Thanks! Scary!   Fixed by #713

PR #716, opened 2022-07-18
field-specific tolerance added by dictionary

poulNA:
Default field-specific tolerances organized by a tolerance dictionary

Closes https://github.com/annabellej/mirgecompare/issues/20

**Questions for the review**:

### DIFF
--- a/bin/mirgecompare.py
+++ b/bin/mirgecompare.py
@@ -29,8 +29,8 @@ if __name__ == "__main__":
     file_type = file_split[1:]  # remove dot
 
     user_tolerance = {
-    "u": 1e-12,
-    "v": 1e-12
+        "u": 1e-12,
+        "v": 1e-12
     }
 
     if args.tolerance:
--- a/bin/mirgecompare.py
+++ b/bin/mirgecompare.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 
+import json
 from mirgecom.simutil import\
     compare_files_vtu, compare_files_xdmf, compare_files_hdf5
 
@@ -14,6 +15,7 @@ if __name__ == "__main__":
         description="Process files to perform fidelity check")
     parser.add_argument("files", nargs=2, type=str)
     parser.add_argument("--tolerance", type=float)
+    parser.add_argument("-t")
     args = parser.parse_args()
 
     first_file = args.files[0]
@@ -32,7 +34,10 @@ if __name__ == "__main__":
     if args.tolerance:
         user_tolerance = args.tolerance
 
-    field_tolerance = {"u": 1e-12, "v": 1e-12}
+    field_tolerance = {}
+    if args.t:
+        field_tol_inp = json.loads(args.t)
+        field_tolerance = field_tol_inp
 
     # use appropriate comparison function for file type
     if file_type == "vtu" or file_type == "pvtu":
--- a/bin/mirgecompare.py
+++ b/bin/mirgecompare.py
@@ -32,9 +32,12 @@ if __name__ == "__main__":
     if args.tolerance:
         user_tolerance = args.tolerance
 
+    field_tolerance = {"u": 1e-12, "v": 1e-12}
+
     # use appropriate comparison function for file type
     if file_type == "vtu" or file_type == "pvtu":
-        compare_files_vtu(first_file, second_file, file_type, user_tolerance)
+        compare_files_vtu(first_file, second_file, file_type, user_tolerance,
+            field_tolerance)
     elif file_type == "xmf":
         compare_files_xdmf(first_file, second_file, user_tolerance)
     elif file_type == "h5":
--- a/bin/mirgecompare.py
+++ b/bin/mirgecompare.py
@@ -36,8 +36,7 @@ if __name__ == "__main__":
 
     field_tolerance = {}
     if args.t:
-        field_tol_inp = json.loads(args.t)
-        field_tolerance = field_tol_inp
+        field_tolerance = json.loads(args.t)
 
     # use appropriate comparison function for file type
     if file_type == "vtu" or file_type == "pvtu":
--- a/bin/mirgecompare.py
+++ b/bin/mirgecompare.py
@@ -28,11 +28,7 @@ if __name__ == "__main__":
     file_split = os.path.splitext(first_file)[1]
     file_type = file_split[1:]  # remove dot
 
-    user_tolerance = {
-        "u": 1e-12,
-        "v": 1e-12
-    }
-
+    user_tolerance = 1e-12
     if args.tolerance:
         user_tolerance = args.tolerance
 
--- a/bin/mirgecompare.py
+++ b/bin/mirgecompare.py
@@ -28,7 +28,11 @@ if __name__ == "__main__":
     file_split = os.path.splitext(first_file)[1]
     file_type = file_split[1:]  # remove dot
 
-    user_tolerance = 1e-12
+    user_tolerance = {
+    "u": 1e-12,
+    "v": 1e-12
+    }
+
     if args.tolerance:
         user_tolerance = args.tolerance
 
--- a/mirgecom/simutil.py
+++ b/mirgecom/simutil.py
@@ -438,7 +438,7 @@ def compare_files_vtu(
         first_file: str,
         second_file: str,
         file_type: str,
-        tolerance: float = 1e-12
+        tolerance
         ):
     """Compare files of vtu type.
 
@@ -507,14 +507,16 @@ def compare_files_vtu(
             raise ValueError("Fidelity test failed: Mismatched data array sizes")
 
         # verify individual values w/in given tolerance
-        print(f"Field: {point_data2.GetArrayName(i)}", end=" ")
+        fieldname = point_data1.GetArrayName(i)
+        print(f"Field: {fieldname}", end=" ")
         for j in range(arr1.GetSize()):
             test_err = abs(arr1.GetValue(j) - arr2.GetValue(j))
             if test_err > max_field_errors[i]:
                 max_field_errors[i] = test_err
         print(f"Max Error: {max_field_errors[i]}")
 
-    violation = any([max_field_errors[i] > tolerance for i in range(nfields)])
+    violation = any([max_field_errors[i] > tolerance[fieldname]
+        for i in range(nfields)])
     if violation:
         raise ValueError("Fidelity test failed: Mismatched data array "
                                  f"values {tolerance=}.")
--- a/mirgecom/simutil.py
+++ b/mirgecom/simutil.py
@@ -437,17 +437,18 @@ def create_parallel_grid(comm, generate_grid):
 def configurate(config_key, config_object=None, default_value=None):
     """Return a configured item from a configuration object."""
     if config_object is not None:
-        if config_key in config_object:
-            return config_object[config_key]
-
-    return default_value
+        d = config_object if isinstance(config_object, dict) else\
+            config_object.__dict__
+        if config_key in d:
+            return d[config_key]
+        return default_value
 
 
 def compare_files_vtu(
         first_file: str,
         second_file: str,
         file_type: str,
-        tolerance: float
+        tolerance: float = 1e-12
         ):
     """Compare files of vtu type.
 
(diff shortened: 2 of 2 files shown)


bin/mirgecompare.py
```diff
@@ -28,7 +28,11 @@
     file_split = os.path.splitext(first_file)[1]
     file_type = file_split[1:]  # remove dot
 
-    user_tolerance = 1e-12
+    user_tolerance = {
+    "u": 1e-12,
+    "v": 1e-12
+    }
+
```
matthiasdiener, line 35: What happens when a fieldname is not in the dict? Maybe add a function `get_tolerance(fieldname)` that returns either a value from the dict or another default value.

poulNA: Can this be a nested function or should it be defined outside of compare_files_vtu so it can be used with other compare functions?

MTCam: Sorry, it is here:

https://github.com/illinois-ceesd/mirgecom/blob/refactor-ux/mirgecom/simutil.py

matthiasdiener: hmm, since this is not in main, what do we do? should we just add the function as part of this PR? Also, I was not sure what we do regarding the config_object. Is there an example on how to use it?

MTCam: I just wanted to bounce this function off of you guys since you were discussing adding a function to do the same thing.   It would be my preference to add a function like this to `simutil` and then use it in this utility.   I have other (WIP) uses for this function.

Regarding `config_object` - it is totally users choice.   A dictionary works in this function (https://github.com/illinois-ceesd/mirgecom/blob/23d4d4bab885b891e3d59fd14006ad74092f1e03/mirgecom/simutil.py#L465), but also other sorts of objects too, since `object.__dict__` is more general.   This API allows us to change what we use as the `config_object` without changing the function with which we extract data from it.

mirgecom/simutil.py
```diff
@@ -434,11 +434,20 @@ def create_parallel_grid(comm, generate_grid):
     return generate_and_distribute_mesh(comm=comm, generate_mesh=generate_grid)
 
 
+def configurate(config_key, config_object=None, default_value=None):
+    """Return a configured item from a configuration object."""
+    if config_object is not None:
+        if config_key in config_object:
+            return config_object[config_key]
```
MTCam, line 441: If it is like this, then it will continue to work for python dictionaries, but also for more general objects.

```suggestion
         if config_key in config_object.__dict__:
            return config_object.__dict__[config_key]
```

matthiasdiener: I'm just wondering how that would work with a directly passed `dict`. `{}.__dict__` does not work.

matthiasdiener, line 441:
```suggestion
def configurate(config_key, config_object=None, default_value=None):
    """Return a configured item from a configuration object."""
    if config_object is not None:
        if isinstance(config_object, dict):
            d = config_object
        else:
            d = config_object.__dict__
        if config_key in d:
            return d[config_key]
```

matthiasdiener: See my comment below https://github.com/illinois-ceesd/mirgecom/pull/716/files#r927776964

MTCam: `d = config_object if isinstance(config_object, dict) else config_object.__dict__`?

mirgecom/simutil.py
```diff
@@ -434,11 +434,20 @@ def create_parallel_grid(comm, generate_grid):
...
+
+    return default_value
+
+
 def compare_files_vtu(
         first_file: str,
         second_file: str,
         file_type: str,
-        tolerance: float = 1e-12
+        tolerance: float
```
matthiasdiener, line 450: I think you can keep the default tolerance as 1e-12 now.

poulNA: I thought default tolerance was set in mirgecompare.py? Either as 1e-12 or whatever --tolerance takes in from the command line.

mirgecom/simutil.py
poulNA:
```violation = any([max_field_errors[i] > configurate(fieldname, field_tol,
TypeError: '>' not supported between instances of 'int' and 'NoneType'
```
Those changes cause this typeerror.

mirgecom/simutil.py
matthiasdiener: Right, but I think there is no harm in providing another default here

mirgecom/simutil.py
matthiasdiener: What did you provide as the `tolerance` argument to `configurate`?

mirgecom/simutil.py
MTCam: The way it currently is, the _caller_ must deal with it if the return is  `None`.   In this case, we need to make sure that a default value is passed so that None is _never_ returned.

mirgecom/simutil.py
poulNA: The function argument `tolerance`

matthiasdiener: Yes, that return is necessary

mirgecom/simutil.py
```diff
@@ -507,14 +517,18 @@ def compare_files_vtu(
...
         for j in range(arr1.GetSize()):
             test_err = abs(arr1.GetValue(j) - arr2.GetValue(j))
             if test_err > max_field_errors[i]:
                 max_field_errors[i] = test_err
-        print(f"Max Error: {max_field_errors[i]}")
+        print(f"Max Error: {max_field_errors[i]}", end=" ")
+        print(f"Tolerance: {configurate(fieldname, field_tol, tolerance)}")
+
+    violation = any([max_field_errors[i] > configurate(fieldname, field_tol,
+                                        tolerance) for i in range(nfields)])
```
MTCam, line 530: I think `fieldname` should be `point_data1.GetArrayName(i)` in this violation check.

MTCam, line 530: I think `fieldname` should be `point_data1.GetArrayName(i)` in this violation check.

MTCam: Come to think of it, perhaps something like:
```
field_tols = [configurate(point_data1.GetArrayName(i), field_tol, tolerance) for i in range(nfields)]
(...)
       print(f"Tolerance: {field_tols[i]}")
 violation = any([max_field_errors[i] > field_tols[i] for i in range(nfields)])
```

poulNA: Ooh, that makes a lot of sense. 
I have this right now but it can't compare between max_field_errors[I] and field_tol[I]... fieldname is point_data1.GetArrayName(i) here:
```
   field_tol[i] = [configurate(fieldname, field_tol, tolerance) for i in range(nfields)]
          print(f"Tolerance: {field_tol[i]}")

violation = any([max_field_errors[i] > field_tol[i] for i in range(nfields)])
```
Output:
```
 violation = any([max_field_errors[i] > field_tol[i] for i in range(nfields)])
TypeError: '>' not supported between instances of 'int' and 'list'
```

mirgecom/simutil.py
```diff
@@ -434,6 +434,16 @@ def create_parallel_grid(comm, generate_grid):
...
 
 
+def configurate(config_key, config_object=None, default_value=None):
+    """Return a configured item from a configuration object."""
+    if config_object is not None:
+        d = config_object if isinstance(config_object, dict) else\
+            config_object.__dict__
+        if config_key in d:
+            return d[config_key]
+        return default_value
```
MTCam, line 444: This needs to be at the function level:
```suggestion
    return default_value
```

mirgecom/simutil.py
MTCam: Woops!  Note that above I have a new variable `field_tols` which is a new _list_ variable containing all the field-specific tolerances (or the default if they did not exist).  Then I use `field_tols` in the violation check *not* the old `field_tol` which was (and should remain) a dictionary.

mirgecom/simutil.py
MTCam: Just make sure to put it back to the function level as it was before.  See the [comment below](https://github.com/illinois-ceesd/mirgecom/pull/716#discussion_r927839180).

mirgecom/simutil.py
poulNA: Ah my bad I just assumed they were the same list. The second block of code you mentioned throws error that the violation bool is not iterable but the first block of code works!

MTCam: Sorry about that!  Ha, well I'm not great at Python.  I should try things out before suggesting them!  Thanks for trying it out!

mirgecom/simutil.py
```diff
@@ -507,14 +518,20 @@ def compare_files_vtu(
...
+        fieldname = point_data1.GetArrayName(i)
+        print(f"Field: {fieldname}", end=" ")
         for j in range(arr1.GetSize()):
             test_err = abs(arr1.GetValue(j) - arr2.GetValue(j))
             if test_err > max_field_errors[i]:
                 max_field_errors[i] = test_err
-        print(f"Max Error: {max_field_errors[i]}")
+        print(f"Max Error: {max_field_errors[i]}", end=" ")
+        field_specific_tols = [configurate(point_data1.GetArrayName(i), field_tol,
+            tolerance) for i in range(nfields)]
```
MTCam, line 529: This line should be outside the i-loop (i.e. up above line 503)

mirgecom/simutil.py
```diff
@@ -490,6 +500,9 @@ def compare_files_vtu(
 
     nfields = point_data1.GetNumberOfArrays()
     max_field_errors = [0 for _ in range(nfields)]
+    field_tol = {}  # tolerance dict here for now
```
MTCam, line 503: Is there a plan for exposing this to the user so he can specify it?

poulNA: I was under the impression that the user can have the choice to specify individual fields instead of providing just one default value for all fields.

poulNA: We were thinking maybe a field specific tolerance file.

MTCam: I think that sounds like a reasonable option.  Any ideas about format? 

A command-line interface might be handy, too.   It looks like it might be possible (although perhaps ill-advised; @matthiasdiener ?), to do something like:

```python
python mirgecompare.py -t '{"u":  1e-5, "v": 1e-6}' <red_file> <blue_file>
```

Just looking at:

https://stackoverflow.com/questions/18006161/how-to-pass-dictionary-as-command-line-argument-to-python-script

poulNA: I think that format for the command line looks optimal. I'll try to implement that.

mirgecom/simutil.py
```diff
@@ -434,11 +434,22 @@ def create_parallel_grid(comm, generate_grid):
...
+    return default_value
+
+
 def compare_files_vtu(
         first_file: str,
         second_file: str,
         file_type: str,
-        tolerance: float = 1e-12
+        tolerance: float = 1e-12,
+        field_tolerance: [float] = None
```
matthiasdiener, line 452: This should probably de `Dict[float]`

bin/mirgecompare.py
```diff
@@ -32,9 +34,15 @@
     if args.tolerance:
         user_tolerance = args.tolerance
 
+    field_tolerance = {}
+    if args.t:
+        field_tol_inp = json.loads(args.t)
+        field_tolerance = field_tol_inp
```
matthiasdiener, line 40: maybe just:

```suggestion
        field_tolerance = json.loads(args.t)
```

poulNA:
Here is what the command line input vs. output looks like, showcasing all three ways to set the tolerance.

### Default

**Input:**
`bin/mirgecompare.py examples/fld-wave-0000.vtu examples/fld-wave-0000.vtu`

**Output:**
```
Field: u Max Error: 0 Tolerance: 1e-12
Field: v Max Error: 0 Tolerance: 1e-12
VTU Fidelity test completed successfully
```

### Manual

**Input:**
`bin/mirgecompare.py --tolerance 1e-13 examples/fld-wave-0000.vtu examples/fld-wave-0000.vtu`

**Output:**
```
Field: u Max Error: 0 Tolerance: 1e-13
Field: v Max Error: 0 Tolerance: 1e-13
VTU Fidelity test completed successfully
```

### Dictionary

**Input:**
`bin/mirgecompare.py -t '{"u":  1e-5, "v": 1e-6}' examples/fld-wave-0000.vtu examples/fld-wave-0000.vtu`

**Output:**
```
Field: u Max Error: 0 Tolerance: 1e-05
Field: v Max Error: 0 Tolerance: 1e-06
VTU Fidelity test completed successfully
```

matthiasdiener (approved):
LGTM, thanks!

MTCam:
> LGTM, thanks!

Great - lemme see if we can get the production test to pass, and we'll be gtg.

PR #717, opened 2022-07-20
Physical Artificial Viscosity

anderson2981:
Add a new transport class that implements artificial viscosity as an additive term to the physical viscosity.

This requires an additional parameter included with the fluid state, the smoothness. The smoothness is derived from the discretization and needs to be available to the transport class to compute the new viscosity term. We have added the smoothness field to dv.

Adds a new example to illustrate how to setup and use the new transport class.

**Questions for the review**:


majosm:
Since we currently define `dv` as something that contains variables derived from the EOS, I wonder if we should avoid putting `smoothness` in there. I experimented with adding an extra `*Vars` class to contain stuff like this (https://github.com/majosm/mirgecom/commit/b381f2ebd2a346de0a9623810436c2ec174002a0), but I'm not sure this is a good idea either (mainly because it becomes unclear what can depend on what -- e.g., `tv` can depend on something in `xv`, but can `xv` depend on something in `tv`? etc.)

I wonder if we could just make `smoothness` an optional member of `FluidState`? Then make `smoothness` an optional argument to the transport model functions?

anderson2981:
Right now, dv is defined in gas model as:
`Fluid state-dependent quantities corresponding to the chosen equation of state.`

I struggled with where to put the smoothness. I would like to move away from using cv_mass to define the smoothness, switching to either an enthalpy or entropy based formulation. In these cases, the smoothness will depend directly on the chosen eos as a function of P,T and would fit into dv I think. Ideally I would have put it there directly, but smoothness needs to know about the discretization to be calculated, and neither dv or tv have that knowledge to compute it. Nor should they IMO. So the solution was to compute smoothness for volume discr on the outside and pass it into the fluid state when it's constructed and let the projection mechanisms put it on the correct discr depending on which piece of the code needed the information

I'm currently testing an example with boundaries

MTCam:
> Since we currently define `dv` as something that contains variables derived from the EOS, I wonder if we should avoid putting `smoothness` in there. I experimented with adding an extra `*Vars` class to contain stuff like this ([majosm@b381f2e](https://github.com/majosm/mirgecom/commit/b381f2ebd2a346de0a9623810436c2ec174002a0)), but I'm not sure this is a good idea either (mainly because it becomes unclear what can depend on what -- e.g., `tv` can depend on something in `xv`, but can `xv` depend on something in `tv`? etc.)
> 
> I wonder if we could just make `smoothness` an optional member of `FluidState`? Then make `smoothness` an optional argument to the transport model functions?

I agree with the approach suggested by @majosm, mostly.   Like Matt, I don't think extending DV is what we should do here.  What we really need is a mechanism by which a user can add arbitrary computed quantities to the thing that gets passed around as "the data" (including the CV, DV, *V).  In this case, then the customized transport class can fail if the smoothness data is not available in the data structure.

I do not think we should cloud CV, or DV with the addition of a semi-permanent, or even optional "smoothness" field.  The smoothness field, imo, is just an example of the type of arbitrary data that the users will want/need to add to mirgecom data structures.  Computing CV, and DV does not depend on having the smoothness.  Instead, smoothness is a function of CV and DV.

All of that said, I think what @anderson2981 has done here is a good/reasonable approach for adding the capability using the existing data structures.   I would not spend much time trying to make it library-worthy however (e.g., adding a specialized EOS, etc) until we can converge on a longer-term approach.  Just my .02.

mirgecom/transport.py
```diff
@@ -255,3 +265,161 @@ def species_diffusivity(self, cv: ConservedVars, dv: GasDependentVars,
...
+        av_kappa = (dv.smoothness*mu
+                    * eos.heat_capacity_cp(cv, dv.temperature)/self._av_prandtl)
+        return av_kappa + self._physical_transport.thermal_conductivity(
+            cv, dv, eos)
+
+    def species_diffusivity(self, cv: ConservedVars,
+                            dv: Optional[GasDependentVars] = None,
+                            eos: Optional[GasEOS] = None) -> DOFArray:
+        r"""Get the vector of species diffusivities, ${d}_{\alpha}$."""
+        return self._physical_transport.species_diffusivity(cv, dv, eos)
```
tulioricci, line 425: Mike, please add:
        return self._physical_transport.species_diffusivity(cv, dv, eos) \
                + dv.smoothness*self._av_species_diffusivity

PR #718, opened 2022-07-28
Bound-preserving limiter

tulioricci:
Adding bound-preserving limiter based on #591 and https://royalsocietypublishing.org/doi/10.1098/rspa.2011.0153

The functions have been exercised by 1D flame propagation (MixtureEOS) and McKenna burner simulation with IdealGas and MixtureEOS. An example on how to call the limiter inside the driver is presented by the autoignition example and a simple test was added to verify if negative species are properly modified to be within physical values.

**Questions for the review**:
-------------------
@MTCam


mirgecom/limiter.py
```diff
@@ -0,0 +1,78 @@
...
+
+def cell_characteristic_size(actx, dcoll: DiscretizationCollection):
+    """Evaluate cell area or volume."""
+    zeros = actx.thaw(actx.freeze(dcoll.nodes()))[0]
+    return op.elementwise_integral(dcoll, zeros + 1.0)
+
+
+def positivity_preserving_limiter(dcoll: DiscretizationCollection, volume, field,
+                                  mmin=0.0, mmax=1.0):
+    """Implement the positivity-preserving limiter of Liu and Osher (1996)."""
```
inducer, line 49: Document the roles of `volume` and `field`, as well as the return value. Ideally also type-annotate them.

inducer, line 49: Make this a proper citation.

mirgecom/limiter.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # If the averaged is not bounded, some numerical exercise of this function
+    # failed since we work with a posteriori correction of the values. This
+    # operation is not described in the paper but greatly increased the robustness.
+    # This will not make the limiter conservative but it is better than having
+    # negative species. This should only be necessary for coarse grids or
+    # underresolved regions... If it is knowingly underresolved, then I think
+    # we can abstain to ensure "exact" conservation.
+    cell_avgs = actx.np.where(actx.np.greater(cell_avgs, mmin), cell_avgs, mmin)
+    cell_avgs = actx.np.where(actx.np.greater(cell_avgs, mmax), mmax, cell_avgs)
```
inducer, line 63: This belongs in the docstring, since what you are implementing is *not* the limiter of Liu and Osher once you include this modification.

mirgecom/limiter.py
```diff
@@ -0,0 +1,78 @@
+""":mod:`mirgecom.limiter` is for limiters and limiter-related constructs.
```
inducer, line 1: This needs to be hooked into the docs.

mirgecom/limiter.py
```diff
@@ -0,0 +1,78 @@
...
+    # Compute cell averages of the state
+    cell_avgs = 1.0/volume*op.elementwise_integral(dcoll, field)
+
+    # If the averaged is not bounded, some numerical exercise of this function
+    # failed since we work with a posteriori correction of the values. This
+    # operation is not described in the paper but greatly increased the robustness.
+    # This will not make the limiter conservative but it is better than having
+    # negative species. This should only be necessary for coarse grids or
+    # underresolved regions... If it is knowingly underresolved, then I think
+    # we can abstain to ensure "exact" conservation.
```
inducer, line 61: I'm not sure I understand this description. What exactly is the goal? What are you doing here? And what about fields that are not bounded above by 1?

inducer, line 61: Perhaps there should be a way to access *just* the Liu/Osher scheme, without your modifications.

tulioricci:
@inducer , thanks for the comments. Just pushed a revised version that covers all your remarks. Now, one argument controls if there is an upper bound and a boolean flag modifies or not the average, recovering the same implementation as the reference.

In my numerical exercises, I could only enforce the negative values to vanish when poking at the average too. If it is not forced back to zero, the "theta" evaluation drops to zero and all point at the cell will have the same negative value as the average. If no action is performed, the average will remain negative and also spreads to neighbor cells. I kept this as the default option. If the user wants to bypass this, the input argument should be False.

mirgecom/limiter.py
```diff
@@ -0,0 +1,108 @@
...
+THE SOFTWARE.
+"""
+
+from grudge.discretization import DiscretizationCollection
+import grudge.op as op
+
+
+def cell_characteristic_size(actx, dcoll: DiscretizationCollection):
+    r"""Evaluate cell area or volume."""
+    zeros = actx.thaw(actx.freeze(dcoll.nodes()))[0]
```
MTCam, line 40:
```suggestion
    zeros = dcoll.zeros(actx)
```

mirgecom/limiter.py
```diff
@@ -0,0 +1,108 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+"""
+
+from grudge.discretization import DiscretizationCollection
+import grudge.op as op
+
+
+def cell_characteristic_size(actx, dcoll: DiscretizationCollection):
```
MTCam, line 38: Any reason not to call this "cell_volume"?

mirgecom/limiter.py
```diff
@@ -0,0 +1,108 @@
...
+    the original reference but greatly increased the robustness.
+
+    An additional argument can be added to bound the upper limit.
+
+    Parameters
+    ----------
+    dcoll: :class:`grudge.discretization.DiscretizationCollection`
+        Grudge discretization with boundaries object
+    cell_size: meshmode.dof_array.DOFArray or numpy.ndarray
+        A field or collection of scalar fields to limit
```
MTCam, line 64: I'm having a little trouble understanding this parameter's doc. Could the doc for this just be something like: "The cell volumes" ? 

Is there a particular reason that the cell volumes should be passed in instead of computed down inside this function?

test/test_limiter.py
```diff
@@ -0,0 +1,95 @@
...
+    eps = 0.1
+    field = nodes[0]*eps
+
+    cell_area = cell_characteristic_size(actx, discr)
+
+    limited_field = bound_preserving_limiter(discr, cell_area, field, mmin=0.0)
+
+    error = actx.np.linalg.norm(op.elementwise_min(discr, limited_field), np.inf)
+
+    assert error > -1.0e-13
```
MTCam, line 65: This is just a nit-pick, but this looks funny to me.  How about:
```suggestion
    fld_min = actx.np.linalg.norm(op.elementwise_min(discr, limited_field), np.inf)
    assert fld_min > -1e-13
```

mirgecom/limiter.py
tulioricci: "Volume" reminds me of 3D.. So I wanted to use a more generic name for 2D as well.

mirgecom/limiter.py
tulioricci: My bad, just copied-and-paste and didn't change. Thanks for pointing that out.

I was computing the cell size/volume at the driver, since it doesn't change in time, and passing to the limiter function.. However, if that shouldn't matter, we can compute down inside the function.

mirgecom/limiter.py
MTCam: I am not fixing to hold merging up over this, but fwiw, I find `volume` to be sufficiently descriptive regardless of dimension, where `characteristic_size` leaves me wondering what this actually is; so I have to go look.   Luckily, in this case, it is an easily-understood one-liner.

MTCam (approved):
Looks good to me.

examples/autoignition-mpi.py
```diff
@@ -500,9 +502,25 @@ def my_get_timestep(t, dt, state):
...
+            + kin_energy
+        )
+
+        return make_conserved(dim=dim, mass=cv.mass, energy=energy_lim,
+                       momentum=cv.momentum, species_mass=cv.mass*spec_lim)
+
     def my_pre_step(step, t, dt, state):
         cv, tseed = state
-        fluid_state = construct_fluid_state(cv, tseed)
+        fluid_state = construct_fluid_state(limiter(cv, temp=tseed), tseed)
```
majosm, line 523: (I know this is merged already, but I just noticed it.) Should `limiter` be getting compiled somewhere?

tulioricci: Good idea. Let me work on that.

MTCam: Note that limiter as called here in this example is in the `pre-step` and does not limit the actual state used by the simulation. Typically I think the limiter is called inside the RHS (`my_rhs`, or whatever), and will be compiled when the RHS is compiled.

inducer: IIUC, the limiter should operate on the state estimate, not the RHS.

MTCam: Correct.   What I mean is that the limiter should be in the user's RHS function, operating on the state that is passed into the RHS.

inducer: I'm not sure I agree: I think the limiter should apply to the post-increment state in the time stepper, not the input of the RHS.

@lukeolson, do you agree?

tulioricci: @inducer , I agree that it is better to use the limiter to the post-increment of the time stepper (i.e., every sub-iteration of the RK). I am still seen negative values (although not as significant as before) and makes sense that applying only the RHS input (in the pre or post-step) may not be sufficient.

I will update the example this week. Thanks for pointing that out!

MTCam: > @inducer , I agree that it is better to use the limiter to the post-increment of the time stepper (i.e., every sub-iteration of the RK). 

Doing the limiting in the pre/post step callbacks precludes limiting every substep of multi-step timesteppers.  The callbacks are only called once per timestep.

I am not sure that it would matter whether one limits pre, or post-increment.   If the limiting is not performed on the state input to the RHS operator(s), then you have the possibility of feeding a state into the RHS that already has negative values.

PR #723, opened 2022-08-12
Add zero level option to thermochem interface.

MTCam:
Squash concentrations below `zero_level` to zero, preventing spurious reaction rates for insignificant concentrations of reactants.

**Questions for the review**:

### DIFF
--- a/mirgecom/thermochemistry.py
+++ b/mirgecom/thermochemistry.py
@@ -29,7 +29,7 @@ THE SOFTWARE.
 """
 
 
-def get_pyrometheus_wrapper_class(pyro_class, temperature_niter=5):
+def get_pyrometheus_wrapper_class(pyro_class, temperature_niter=5, zero_level=0.):
     """Return a MIRGE-compatible wrapper for a :mod:`pyrometheus` mechanism class.
 
     Dynamically creates a class that inherits from a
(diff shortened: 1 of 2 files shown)


mirgecom/thermochemistry.py
```diff
@@ -49,6 +49,8 @@ def get_pyrometheus_wrapper_class(pyro_class, temperature_niter=5):
         Pyro thermochemical mechanism to wrap
     temperature_niter: int
         Number of Newton iterations in `get_temperature` (default=5)
+    zero_level: float
+        Squash concentrations below this level to 0. (default=0.)
```
tulioricci, line 53: Just add some extra comments why this is necessary. The motivation was to avoid that tiny fractions of fuel drive it to be negative because of significant reaction rates. Unfortunately, the limiter could not bring it back to zero if the entire cell is negative.

Maybe adding a description as a header (like this https://github.com/illinois-ceesd/mirgecom/blob/813586f10eb61f580acd2ced58f1fef840d2feb8/mirgecom/euler.py#L1) would be easier rather than copying and paste the same piece of text everywhere.

PS: this header could also be used to explain that 5 iterations are more than enough for the temperature (and we can/should reduce it)

tulioricci (commented):
Pretty straightforward changes. I am okay with this. Please, just clarify why this is necessary.

mirgecom/thermochemistry.py
MTCam: How's this?

[e72722c](https://github.com/illinois-ceesd/mirgecom/pull/723/commits/e72722c00fe136357260e2a9c52e2d9abb0df088)

PR #727, opened 2022-08-17
Mirgecompare: check that rank numbers match

poulNA:
Rank number feature by XML parsing through pvtu files
**Questions for the review**:

### DIFF
--- a/mirgecom/simutil.py
+++ b/mirgecom/simutil.py
@@ -512,7 +512,7 @@ def compare_files_vtu(
     output1 = reader1.GetOutput()
 
     # Check rank number
-    def numranks(filename):
+    def numranks(filename: str) -> int:
         tree = Et.parse(filename)
         root = tree.getroot()
         return len(root.findall(".//Piece"))
--- a/mirgecom/simutil.py
+++ b/mirgecom/simutil.py
@@ -520,7 +520,7 @@ def compare_files_vtu(
     if file_type == "pvtu":
         rank1 = numranks(first_file)
         rank2 = numranks(second_file)
-        if rank1 is not rank2:
+        if rank1 != rank2:
             raise ValueError(f"File '{first_file}' has {rank1} ranks, "
                 f"but file '{second_file}' has {rank2} ranks.")
 
--- a/mirgecom/simutil.py
+++ b/mirgecom/simutil.py
@@ -66,7 +66,6 @@ from meshmode.dof_array import DOFArray
 from typing import List
 from grudge.discretization import DiscretizationCollection
 
-
 logger = logging.getLogger(__name__)
 
 
@@ -461,6 +460,8 @@ def compare_files_vtu(
         If it fails the files contain data outside the given tolerance.
     """
     import vtk
+    import xml.etree.ElementTree as Et
+    from warnings import warn
 
     # read files:
     if file_type == "vtu":
@@ -473,6 +474,19 @@ def compare_files_vtu(
     reader1.SetFileName(first_file)
     reader1.Update()
     output1 = reader1.GetOutput()
+
+    # Check rank number
+    def numranks(filename):
+        tree = Et.parse(filename)
+        root = tree.getroot()
+        return len(root.findall(".//Piece"))
+
+    if file_type == "pvtu":
+        rank1 = numranks(first_file)
+        rank2 = numranks(second_file)
+        if rank1 is not rank2:
+            warn(f"File {first_file} has {rank1} rank, "
+                f"but File {second_file} has {rank2} rank.")
 
     reader2.SetFileName(second_file)
     reader2.Update()
--- a/mirgecom/simutil.py
+++ b/mirgecom/simutil.py
@@ -498,7 +498,6 @@ def compare_files_vtu(
     """
     import vtk
     import xml.etree.ElementTree as Et
-    from warnings import warn
 
     # read files:
     if file_type == "vtu":
@@ -522,8 +521,8 @@ def compare_files_vtu(
         rank1 = numranks(first_file)
         rank2 = numranks(second_file)
         if rank1 is not rank2:
-            warn(f"File '{first_file}' has {rank1} ranks, "
-                f"but File {second_file} has {rank2} rank.")
+            raise ValueError(f"File '{first_file}' has {rank1} ranks, "
+                f"but File {second_file} has {rank2} ranks.")
 
     reader2.SetFileName(second_file)
     reader2.Update()
--- a/mirgecom/simutil.py
+++ b/mirgecom/simutil.py
@@ -522,7 +522,7 @@ def compare_files_vtu(
         rank2 = numranks(second_file)
         if rank1 is not rank2:
             raise ValueError(f"File '{first_file}' has {rank1} ranks, "
-                f"but File {second_file} has {rank2} ranks.")
+                f"but file '{second_file}' has {rank2} ranks.")
 
     reader2.SetFileName(second_file)
     reader2.Update()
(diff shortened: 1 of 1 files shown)


poulNA:
### Example Output

**Comparing file from a single rank and a file from a double rank** 
Now warns if compared files were run with different ranks. Otherwise no warning

Command: `bin/mirgecompare.py 1rank/fld-wave-mpi-000-0000.pvtu 2rank/fld-wave-mpi-000-0000.pvtu`

Output: 
```
File 1rank/fld-wave-mpi-000-0000.pvtu has 1 rank, but File 2rank/fld-wave-mpi-000-0000.pvtu has 2 rank.
  warn(f"File {first_file} has {rank1} rank, "
Field: u Max Error: 0.96754678098049
Field: v Max Error: 0
  File "/home/paulj3/emirge/mirgecom/bin/mirgecompare.py", line 37, in <module>
    compare_files_vtu(first_file, second_file, file_type, user_tolerance)
  File "/home/paulj3/emirge/mirgecom/mirgecom/simutil.py", line 533, in compare_files_vtu
    raise ValueError("Fidelity test failed: Mismatched data array "
ValueError: Fidelity test failed: Mismatched data array values tolerance=1e-12.
```

MTCam:
> ### Example Output
> **Comparing file from a single rank and a file from a double rank** Now warns if compared files were run with different ranks. Otherwise no warning

Hi @poulNA and @matthiasdiener.  I am a little confused why it should be a warning and not just an error.  @matthiasdiener said it is a warning because you guys have to parse the XML yourselves to determine `nranks`.   Why does that preclude an error?

poulNA:
> > ### Example Output
> > **Comparing file from a single rank and a file from a double rank** Now warns if compared files were run with different ranks. Otherwise no warning
> 
> Hi @poulNA and @matthiasdiener. I am a little confused why it should be a warning and not just an error. @matthiasdiener said it is a warning because you guys have to parse the XML yourselves to determine `nranks`. Why does that preclude an error?

I believe it’s because we decided not to halt the execution in the case of different nranks because there might be some useful information. And the code executes fully without running into any other major problems.  Hopefully @matthiasdiener can provide more reason.

MTCam:
> > Hi @poulNA and @matthiasdiener. I am a little confused why it should be a warning and not just an error. @matthiasdiener said it is a warning because you guys have to parse the XML yourselves to determine `nranks`. Why does that preclude an error?
> 
> I believe it’s because we decided not to halt the execution in the case of different nranks because there might be some useful information. And the code executes fully without running into any other major problems. Hopefully @matthiasdiener can provide more reason.

Comparing two VTU sets run on different number of ranks has no value unless we can stitch the meshes and data together from both sets back into a fully assembled mesh.  There is just no way to correlate the VTU files.   Trying to do it should result in an error, imo.

matthiasdiener:
> > ### Example Output
> > **Comparing file from a single rank and a file from a double rank** Now warns if compared files were run with different ranks. Otherwise no warning
> 
> Hi @poulNA and @matthiasdiener. I am a little confused why it should be a warning and not just an error. @matthiasdiener said it is a warning because you guys have to parse the XML yourselves to determine `nranks`. Why does that preclude an error?

I wasn't sure whether there are other circumstances in which a pvtu file could have multiple pieces, e.g. due to exceeding some file size limits or so. If we are reasonably sure that there aren't such circumstances, we might as well convert this to an error.

mirgecom/simutil.py
```diff
@@ -511,6 +512,19 @@ def compare_files_vtu(
...
+    def numranks(filename):
+        tree = Et.parse(filename)
+        root = tree.getroot()
+        return len(root.findall(".//Piece"))
+
+    if file_type == "pvtu":
+        rank1 = numranks(first_file)
+        rank2 = numranks(second_file)
+        if rank1 is not rank2:
+            warn(f"File {first_file} has {rank1} rank, "
```
matthiasdiener, line 525:
```suggestion
            warn(f"File '{first_file}' has {rank1} ranks, "
```

MTCam:
> I wasn't sure whether there are other circumstances in which a pvtu file could have multiple pieces, e.g. due to exceeding some file size limits or so. If we are reasonably sure that there aren't such circumstances, we might as well convert this to an error.

afaik the only way to get multiple VTU files at a single time level in a single run is to write a partitioned dataset, and there is also not a built-in way through VTK to assemble I think.    If we are ever to be able to compare two runs of differing Nranks (which we do really need!), we'll need to assemble both datasets back to their serial representation (ugh).  If we switch to HDF5, or another format which can do single file (serializes on-the-fly), then we would not have this limitation.  

This basic problem is just true for formats and i/o strategies where we write one file per partition (rank).

poulNA:
> > I wasn't sure whether there are other circumstances in which a pvtu file could have multiple pieces, e.g. due to exceeding some file size limits or so. If we are reasonably sure that there aren't such circumstances, we might as well convert this to an error.
> 
> afaik the only way to get multiple VTU files at a single time level in a single run is to write a partitioned dataset, and there is also not a built-in way through VTK to assemble I think. If we are ever to be able to compare two runs of differing Nranks (which we do really need!), we'll need to assemble both datasets back to their serial representation (ugh). If we switch to HDF5, or another format which can do single file (serializes on-the-fly), then we would not have this limitation.
> 
> This basic problem is just true for formats and i/o strategies where we write one file per partition (rank).

Should I then go ahead and change the warning to an error?

matthiasdiener:
> > > I wasn't sure whether there are other circumstances in which a pvtu file could have multiple pieces, e.g. due to exceeding some file size limits or so. If we are reasonably sure that there aren't such circumstances, we might as well convert this to an error.
> > 
> > 
> > afaik the only way to get multiple VTU files at a single time level in a single run is to write a partitioned dataset, and there is also not a built-in way through VTK to assemble I think. If we are ever to be able to compare two runs of differing Nranks (which we do really need!), we'll need to assemble both datasets back to their serial representation (ugh). If we switch to HDF5, or another format which can do single file (serializes on-the-fly), then we would not have this limitation.
> > This basic problem is just true for formats and i/o strategies where we write one file per partition (rank).
> 
> Should I then go ahead and change the warning to an error?

Yes please.

mirgecom/simutil.py
```diff
@@ -511,6 +511,19 @@ def compare_files_vtu(
...
+        tree = Et.parse(filename)
+        root = tree.getroot()
+        return len(root.findall(".//Piece"))
+
+    if file_type == "pvtu":
+        rank1 = numranks(first_file)
+        rank2 = numranks(second_file)
+        if rank1 is not rank2:
+            raise ValueError(f"File '{first_file}' has {rank1} ranks, "
+                f"but File {second_file} has {rank2} ranks.")
```
matthiasdiener, line 525:
```suggestion
                f"but file ‘{second_file}’ has {rank2} ranks.")
```

mirgecom/simutil.py
```diff
@@ -511,6 +511,19 @@ def compare_files_vtu(
     reader1.Update()
     output1 = reader1.GetOutput()
 
+    # Check rank number
+    def numranks(filename):
```
matthiasdiener, line 515:
```suggestion
    def numranks(filename: str) -> int:
```

mirgecom/simutil.py
```diff
@@ -511,6 +511,19 @@ def compare_files_vtu(
...
+        tree = Et.parse(filename)
+        root = tree.getroot()
+        return len(root.findall(".//Piece"))
+
+    if file_type == "pvtu":
+        rank1 = numranks(first_file)
+        rank2 = numranks(second_file)
+        if rank1 is not rank2:
```
matthiasdiener, line 523:
```suggestion
        if rank1 != rank2:
```

matthiasdiener (commented):
Apart from these minor comments, LGTM!

matthiasdiener (approved):
LGTM, thanks!

MTCam (approved):
This looks great!  Thanks, Paul!

PR #735, opened 2022-08-25
Physical av

anderson2981:
Production feeder for physical artificial viscosity.

Add a new transport class that implements artificial viscosity as an additive term to the physical viscosity.

This requires an additional parameter included with the fluid state, the smoothness. The smoothness is derived from the discretization and needs to be available to the transport class to compute the new viscosity term. We have added the smoothness field to dv.

Adds a new example to illustrate how to setup and use the new transport class.

based on work in #717 

**Questions for the review**:

### DIFF
--- a/mirgecom/eos.py
+++ b/mirgecom/eos.py
@@ -74,11 +74,14 @@ class GasDependentVars:
 
     .. attribute:: temperature
     .. attribute:: pressure
+    .. attribute:: speed_of_sound
+    .. attribute:: smoothness
     """
 
     temperature: DOFArray
     pressure: DOFArray
     speed_of_sound: DOFArray
+    smoothness: DOFArray
 
 
 @dataclass_array_container
@@ -167,7 +170,8 @@ class GasEOS(metaclass=ABCMeta):
 
     def dependent_vars(
             self, cv: ConservedVars,
-            temperature_seed: Optional[DOFArray] = None) -> GasDependentVars:
+            temperature_seed: Optional[DOFArray] = None,
+            smoothness: Optional[DOFArray] = None) -> GasDependentVars:
         """Get an agglomerated array of the dependent variables.
 
         Certain implementations of :class:`GasEOS` (e.g. :class:`MixtureEOS`)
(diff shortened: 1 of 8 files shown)


mirgecom/eos.py
```diff
@@ -167,18 +174,33 @@ def get_internal_energy(self, temperature, species_mass_fractions=None):
...
 
         Certain implementations of :class:`GasEOS` (e.g. :class:`MixtureEOS`)
         may raise :exc:`TemperatureSeedMissingError` if *temperature_seed* is not
         given.
         """
         temperature = self.temperature(cv, temperature_seed)
+        # MJA, it doesn't appear that we can have a None field embedded inside DV,
+        # make a dummy smoothness in this case
+        if smoothness_mu is None:
+            smoothness_mu = 0. * cv.mass
```
lukeolson, line 191: Are there any (compile) risks with initializing `smoothness_mu` from `cv.mass`?  Should we use `zeros` or `zeros_like` directly from the array context?

MTCam: This pattern is used pervasively.  While we prefer the `zeros_like` facility (which was created exactly for this type of thing), this pattern is not (or has not presented itself as) a compilation issue.

MTCam: What do we think of this? [1425245](https://github.com/illinois-ceesd/mirgecom/pull/735/commits/1425245a48ae4b4de54e19fe365944b4eb8d39f2)

lukeolson (approved):

mirgecom/eos.py
lukeolson: Looks better -- there are several places where this is used with `smoothness_*`

MTCam: I only found a couple places in eos.   I suppose where they are not passed (i.e. for inviscid-only and runs with AV off), that setting them to the same zeros is OK. [e1ad33f](https://github.com/illinois-ceesd/mirgecom/pull/735/commits/e1ad33fe6c95a551bd40cacaf867baf60c2e015b)

PR #726, opened 2022-08-15
Multiple volumes

majosm:
Extracted out of #630. Depends on inducer/grudge#246 and inducer/meshmode#342.

**Questions for the review** @lukeolson:
- [x] Is the scope and purpose of the PR clear?
  - [x] The PR should have a description.
    - Common updates include updating `dcoll`, `BoundaryDomainTag` (from grudge multi-volume), adding a dof descriptor `dd`, promoting the boundary tags.
  - [x] The PR should have a guide if needed (e.g., an ordering).
- [x] Is every top-level method and class documented? Are things that should be documented actually so?
  - [x] add a more complete description in the docstring of `multiple-volumes-mpi.py` to check this off.
- [x] Is the interface understandable? (I.e. can someone figure out what stuff does?) Is it well-defined?
- [x] Does the implementation do what the docstring claims?
- [x] Is everything that is implemented covered by tests?
  - [x] yes, although tests are only covering independent volumes.  There is no coupling.
  - [x] Add a note that `u=x` and `u=y` on vol 1 and 2, resp, to close this box: https://github.com/illinois-ceesd/mirgecom/pull/726/files#diff-a711e4a672e5ea17608d8536895c9a666557a0872b0b23b2e3f3809874898dfbR91
- [x] Do you see any immediate risks or performance disadvantages with the design? Example: what do interface normals attach to?

PR #741, opened 2022-09-02
Add optional user-specified `comm_tag` to operators

majosm:
...for disambiguating multiple operator calls within a single compiled function.

**Questions for the review**:

### DIFF
--- a/mirgecom/artificial_viscosity.py
+++ b/mirgecom/artificial_viscosity.py
@@ -149,10 +149,6 @@ from grudge.dof_desc import (
 import grudge.op as op
 
 
-class _AVCVTag:
-    pass
-
-
 class _AVRTag:
     pass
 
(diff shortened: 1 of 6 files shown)


MTCam (approved):
Neat solution.

majosm:
(Verified in the `multiple-volumes-mpi.py` example of #726.)

@matthiasdiener @inducer Does this look OK to you guys?

inducer:
LGTM. Two thoughts I had while scrolling:

- There's a potential disaster here if someone were to forget to pass the tags through a level. We should think of a way to give the user a chance to examine the set of tags to make sure it matches expectations. https://github.com/inducer/pytato/issues/357 https://github.com/inducer/pytato/issues/358
- Given the number of kwargs in some of those routines, you want to start kwarg-only-ing some of those. (I saw that done in one place already :tada:)

PR #745, opened 2022-09-08
use pocl_dev channel

matthiasdiener:
**Questions for the review**:

### DIFF
--- a/conda-env.yml
+++ b/conda-env.yml
@@ -4,6 +4,7 @@
 #
 name: ceesd
 channels:
+- conda-forge/label/pocl_dev
 - conda-forge
 - nodefaults
 dependencies:


inducer:
I wonder if we should just incorporate https://github.com/pocl/pocl/pull/1069 into the feedstock instead of patching all the downstreams. It's clearly a fix... (though I think I should address the latest comments there.

matthiasdiener:
> I wonder if we should just incorporate [pocl/pocl#1069](https://github.com/pocl/pocl/pull/1069) into the feedstock instead of patching all the downstreams. It's clearly a fix... 

Yeah, I think that might be a good idea; Now that we have a new pyopencl release, new emirge installations (or people running conda update in an existing installation) will always run with an SVM memory pool and hit the broken SVM free implementation in pocl... 

> (though I think I should address the latest comments there.

If you want, I can go ahead and address them today.

inducer:
I'm not opposed to also adding pocl-cuda-svm. AFAICT, there should not be any aspects of that that affect existing users of pocl, so I don't see why not.

> If you want, I can go ahead and address them today.

That'd be helpful. I've invited you to my pocl fork. Do separate commits so I can easily see what you've changed.

PR #751, opened 2022-09-15
add vmprof profiling for time stepper

matthiasdiener:
**Questions for the review**:


mirgecom/steppers.py
```diff
@@ -136,6 +144,15 @@ def _advance_state_stepper_func(rhs, timestepper,
...
     compiled_rhs = _compile_rhs(actx, rhs)
 
+    if vmprof_filename:
+        import vmprof
+        import os
+
+        logger.info(f"Logging vmprof profile to '{vmprof_filename}'.")
+
+        outfd = os.open(vmprof_filename, os.O_RDWR | os.O_CREAT | os.O_TRUNC)
+        vmprof.enable(outfd)
```
MTCam, line 154: I think this would be more powerful if left in the hands of the driver/sim developer. 

The file can be opened in the driver, and the call(s) to `vmprof.{en,dis}able` could be performed by the `{pre,post}_step_callback` functions.

That'd obviate direct changes to the stepper, and allow the user to turn vmprof on-and-off at set timesteps.

PR #766, opened 2022-09-21
Use fusion actx env.

MTCam:
This change set will point `mirgecom@main` at the fusion array context, and upstream packages required for our prediction runs.  It renders our GPU testing more effective in that it speeds it up by more than an order of magnitude, and allows us to test lazy multi-species runs from main (which does not work with the default array context).

**Questions for the review** @lukeolson:
- [x] ~~Is every top-level method and class documented? Are things that should be documented actually so?~~
- [x] ~~Is the interface understandable? (I.e. can someone figure out what stuff does?) Is it well-defined?~~
- [x] ~~Does the implementation do what the docstring claims?~~
- [x] ~~Do you see any immediate risks or performance disadvantages with the design? Example: what do interface normals attach to?~~

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,17 +8,19 @@ pymetis
 importlib-resources
 psutil
 pyyaml
+git+https://github.com/pythological/kanren.git#egg=miniKanren
 
 # The following packages will be git cloned by emirge:
 --editable git+https://github.com/inducer/pymbolic.git#egg=pymbolic
 #--editable git+https://github.com/inducer/pyopencl.git#egg=pyopencl
---editable git+https://github.com/inducer/loopy.git#egg=loopy
+--editable git+https://github.com/kaushikcfd/loopy.git#egg=loopy
 --editable git+https://github.com/inducer/dagrt.git#egg=dagrt
 --editable git+https://github.com/inducer/leap.git#egg=leap
 --editable git+https://github.com/inducer/modepy.git#egg=modepy
---editable git+https://github.com/inducer/arraycontext.git#egg=arraycontext
---editable git+https://github.com/inducer/meshmode.git#egg=meshmode
+--editable git+https://github.com/kaushikcfd/arraycontext.git#egg=arraycontext
+--editable git+https://github.com/kaushikcfd/meshmode.git#egg=meshmode
 --editable git+https://github.com/majosm/grudge.git@multi-main#egg=grudge
---editable git+https://github.com/inducer/pytato.git#egg=pytato
+--editable git+https://github.com/kaushikcfd/pytato.git#egg=pytato
 --editable git+https://github.com/ecisneros8/pyrometheus.git#egg=pyrometheus
 --editable git+https://github.com/illinois-ceesd/logpyle.git#egg=logpyle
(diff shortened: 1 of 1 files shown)


matthiasdiener (approved):
21 mins for examples on porter, not bad.

MTCam:
> 21 mins for examples on porter, not bad.

The main things for me is that it speeds the actual running portion substantially:
lazy accuracy tests:  Before(7.5m)   After(1.5m)
examples tests: Before(121m)  After(10m)

... and it also allows us to now run _all_ the examples in lazy mode, instead of only the ones that work under the default array context.

requirements.txt
```diff
@@ -8,17 +8,19 @@ pymetis
...
---editable git+https://github.com/inducer/arraycontext.git#egg=arraycontext
---editable git+https://github.com/inducer/meshmode.git#egg=meshmode
+--editable git+https://github.com/kaushikcfd/arraycontext.git#egg=arraycontext
+--editable git+https://github.com/kaushikcfd/meshmode.git#egg=meshmode
 --editable git+https://github.com/majosm/grudge.git@multi-main#egg=grudge
---editable git+https://github.com/inducer/pytato.git#egg=pytato
+--editable git+https://github.com/kaushikcfd/pytato.git#egg=pytato
 --editable git+https://github.com/ecisneros8/pyrometheus.git#egg=pyrometheus
 --editable git+https://github.com/illinois-ceesd/logpyle.git#egg=logpyle
+--editable git+https://github.com/kaushikcfd/feinsum.git#egg=feinsum
```
lukeolson, line 26: Why is this needed?

requirements.txt
```diff
@@ -8,17 +8,19 @@ pymetis
 importlib-resources
 psutil
 pyyaml
+git+https://github.com/pythological/kanren.git#egg=miniKanren
```
lukeolson, line 11: Why is this needed?

lukeolson (approved):

requirements.txt
MTCam: @kaushikcfd can you let us know if these packages are required?

MTCam:
@lukeolson - thanks for your review on this.  I do not know exactly *why* these packages are required, but I do see that both are used in @kaushikcfd 's versions of `meshmode` and miniKanren is also used in `loopy`, so I am inclined to go ahead and merge this.  We can always drop them out easily if we want.

lukeolson:
Right.  I guess my question is why do we need to repeat the dep requirement [1]

[1] https://github.com/kaushikcfd/meshmode/blob/main/requirements.txt#L32

MTCam:
> Right. I guess my question is why do we need to repeat the dep requirement [1]
> 
> [1] [kaushikcfd/meshmode@`main`/requirements.txt#L32](https://github.com/kaushikcfd/meshmode/blob/main/requirements.txt?rgh-link-date=2022-09-24T16%3A25%3A34Z#L32)

Oh I see. I admit I do not know.  

I do not know why we need to repeat _any_ packages in the `mirgecom/requirements.txt` that already appear in those for the subpackages.   At this point, and after a cursory discussion of it with @matthiasdiener ,  I am confident that their appearance here:

- may or may not be required
- is doing no apparent harm

I lean toward not fixing what isn't broken.

PR #767, opened 2022-09-21
Wall coupling

majosm:
Adds functionality for simulations containing thermally-coupled gas and wall subdomains. 

**Questions for the review** @MTCam :


mirgecom/diffusion.py
```diff
@@ -175,44 +214,67 @@ def __init__(self, value):
...
 
 class _DiffusionGradTag:
     pass
 
 
+class _DiffusionLengthscalesTag:
+    pass
+
+
 def grad_operator(
```
tulioricci, line 275: @majosm , this function name struggles with this one: https://github.com/illinois-ceesd/mirgecom/blob/main/mirgecom/operators.py#L36

Maybe it would be better to rename it?

majosm: You can alias function names when you import them, like this: https://github.com/illinois-ceesd/mirgecom/blob/24be73a00a0085996e404a74e7fb6dfc408dbcdc/mirgecom/multiphysics/thermally_coupled_fluid_wall.py#L69

majosm:
```
************* Module mirgecom.multiphysics.thermally_coupled_fluid_wall
[24](https://github.com/illinois-ceesd/mirgecom/actions/runs/4078592674/jobs/7028999559#step:5:25)
mirgecom/multiphysics/thermally_coupled_fluid_wall.py:257:8: E0611: No name '_get_rotation_matrix' in module 'mirgecom.boundary' (no-name-in-module)
```

Hmm, looks like this feeder depends on #678 now after [b944c8d](https://github.com/illinois-ceesd/mirgecom/pull/767/commits/b944c8dfa0722750b706b7bda1e212668427f19f). @MTCam What do we need to do to set that up?

MTCam:
> Hmm, looks like this feeder depends on #678 now after [b944c8d](https://github.com/illinois-ceesd/mirgecom/pull/767/commits/b944c8dfa0722750b706b7bda1e212668427f19f). @MTCam What do we need to do to set that up?

I have been wondering if this would happen.  Any chance you could just merge this change set on top of #735 so that the production feeders would go `mengaldo-bcs` -> `physical-av` -> `wall-coupling` = `build-production`?

tulioricci:
@majosm , I started working on a documentation for the wall-model/coupling in https://github.com/illinois-ceesd/mirgecom/pull/855/files. Maybe you can copy-and-paste then finish what I started since it makes more sense in this PR.. Adding the rest of the documentation in the different functions would be great too.

Also, this chunk of code (https://github.com/illinois-ceesd/mirgecom/pull/767/files#diff-a0739e93cfcad9332b45a7073fea3d7e4b00bee8b17a57cf7d7bee6b804b5042R760-R862) is confusing due to the lack of comments and it seems to be the core of the coupling.

Ping me again when you are done with the changes due to Mengaldo and I can review.

mirgecom/diffusion.py
```diff
@@ -175,44 +214,67 @@ def __init__(self, value):
...
+            kappa_tpair, u_tpair, grad_u_tpair, lengthscales_tpair, normal,
+            penalty_amount=penalty_amount)
+
+
+class _DiffusionKappa1Tag:
+    pass
+
+
+class _DiffusionKappa2Tag:
+    pass
```
MTCam, line 256: I wonder if it would be cleaner to collapse some stuff like this.  The tag classes seem awkward.  What about sthn like:
```suggestion
class _DiffusionKappaCommTag:
     pass
```

Then using `(_DiffusionKappaCommTag, 1)`, `(_DiffusionKappaCommTag, 2)`.   This strategy could help reduce some of the awkward looking creation of many similarly named tags?

mirgecom/multiphysics/thermally_coupled_fluid_wall.py
```diff
@@ -0,0 +1,1041 @@
...
+        for i in range(dim):
+            comps.append(axis[i])
+
+    from pytools.obj_array import make_obj_array
+    comps = make_obj_array(comps)
+    return comps.reshape(dim, dim)
+
+
+class InterfaceFluidSlipBoundary(PrescribedFluidBoundary):
+    """Interface boundary condition for the fluid side."""
```
MTCam, line 170: Looking through this, it could possibly directly inherit from the AdiabaticSlip to pick up the fluid boundary stuff.   If not then maybe at least MengaldoBoundaryCondition.   I hope we can avoid using PrescribedFluidBoundary.

mirgecom/viscous.py
```diff
@@ -342,6 +344,88 @@ def viscous_facial_flux_central(dcoll, state_pair, grad_cv_pair, grad_t_pair,
     return num_flux_central(f_int, f_ext)@normal
 
 
+def viscous_facial_flux_harmonic(dcoll, state_pair, grad_cv_pair, grad_t_pair,
+                                 gas_model=None):
```
MTCam, line 348: This inadvertently leaked into #868 , i'll move it out of there and let you add it here.

test/test_diffusion.py
```diff
@@ -377,7 +377,7 @@ def get_rhs(t, u):
 
 
 @pytest.mark.parametrize("order", [1, 2, 3, 4])
-def test_diffusion_discontinuous_kappa(actx_factory, order, visualize=False):
+def test_diffusion_discontinuous_kappa(actx_factory, order, visualize=True):
```
MTCam, line 380: Maybe revert to `False` for merge?

test/test_multiphysics.py
```diff
@@ -40,6 +62,17 @@
 logger = logging.getLogger(__name__)
 
 
+def get_box_mesh(dim, a, b, n):
```
MTCam, line 65: We really need a unified box mesh generator for mirgecom. This appears re-implemented in many places.

MTCam (approved):
👍 Nothing but minor comments from me.  Nothing that needs to be addressed before merge as far as I can tell.

examples/run_examples.sh
```diff
@@ -72,7 +72,8 @@ do
         echo "*** Example $example failed."
         failed_examples="$failed_examples $example"
     fi
-    rm -rf *vtu *sqlite *pkl *-journal restart_data
+    # FIXME: This could delete data from other runs
+    # rm -rf *vtu *sqlite *pkl *-journal restart_data
```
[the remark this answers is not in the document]
majosm: This bit me once when I had a bunch of output files from a longer run of one of the examples (they got removed in addition to the ones produced by `run_examples.sh`). Maybe it would be better for the `rm` part of this to be in the CI script somewhere?

mirgecom/diffusion.py
majosm: I ended up removing the duplicated communication, so there's only one of each `DiffusionKappaCommTag`/`DiffusionStateCommTag` now.

test/test_multiphysics.py
majosm: Yeah, agreed. https://github.com/illinois-ceesd/mirgecom/issues/887

MTCam (approved):

PR #768, opened 2022-09-22
Make wall coupling a feeder

majosm:
**Questions for the review**:


majosm:
OK if I merge this into production @MTCam (minus the testing env stuff)?

mirgecom/viscous.py
```diff
@@ -332,10 +333,29 @@ def viscous_facial_flux_central(dcoll, state_pair, grad_cv_pair, grad_t_pair,
...
+
+    state_pair_with_harmonic_mean_coefs = TracePair(
+        state_pair.dd,
+        interior=replace_coefs(state_pair.int, kappa=kappa_harmonic_mean),
+        exterior=replace_coefs(state_pair.ext, kappa=kappa_harmonic_mean))
+
+    f_int = viscous_flux(
+        state_pair_with_harmonic_mean_coefs.int, grad_cv_pair.int, grad_t_pair.int)
+    f_ext = viscous_flux(
+        state_pair_with_harmonic_mean_coefs.ext, grad_cv_pair.ext, grad_t_pair.ext)
```
MTCam, line 358: I would like it a lot better if this would be a separate function instead of changing `viscous_facial_flux_central`.  Why not add `viscous_facial_flux_harmonic`?

tulioricci: I agree with Mike. With two functions, I could exercise it with the 1D flame or the standing burner and check if there is any impact/downside when the diffusion coefficients change.

MTCam: My issue with it the way that it is it does something other than straight "central" when the thermal conductivity is tweaked before flux computation.  We have machinery where the user can pass in the numerical flux function they want to use, so adding a new flux function as a possible choice seems like what we'd want to do here.

PR #772, opened 2022-09-26
Adding variable spec. diff to PowerLawTransp

tulioricci:
- Adding Lewis number to evaluate species diffusivity as a function of temperature within the PowerLawTransport model. User-specified Lewis number (as an array with "nspecies" shape) may be used if necessary.

- Adding PowerLawTransport to the ns-mix.py example.

**Questions for the review**:
- [x] Is the scope and purpose of the PR clear?
  - [x] The PR should have a description.
  - [x] The PR should have a guide if needed (e.g., an ordering).
- [x] Is every top-level method and class documented? Are things that should be documented actually so?
- [x] Is the interface understandable? (I.e. can someone figure out what stuff does?) Is it well-defined?
- [x] Does the implementation do what the docstring claims?
- [ ] Is everything that is implemented covered by tests?
    - Powerlaw transport has MMS tests that are not checked-in yet, that's a separate effort
- [x] Do you see any immediate risks or performance disadvantages with the design? Example: what do interface normals attach to?
	- Users could accidentally use the powerlaw transport wrong by sending the wrong thing (scalar or mis-shaped) lewis array.

PR #776, opened 2022-09-27
Change evaluation of cell average data inside limiter

tulioricci:
- Change cell average evalution from "op.elementwise_integral" to "nodal-to-modal, zero-out polynomials, convert back from modal-to-nodal."
- Added function in "simutil" to avoid recomputing temperature for the limited state. Temperature is kept constant during the limiting operation. Thus, there is no need to go over the expensive temperature evaluation if it doesn't change. Instead, all other DVs and TVs (only for N-S eqs.) are re-evaluated for speed-up.
- Add compiled functions to autoignition example.

Work in progress toward applying the limiter during the state creation (https://github.com/illinois-ceesd/mirgecom/pull/743) which will replace the current implementation. However, the current changes are necessary to speed-up stable calculations with multi-species.

**Questions for the review**:


mirgecom/limiter.py
```diff
@@ -88,14 +92,30 @@ def bound_preserving_limiter(dcoll: DiscretizationCollection, field,
...
-    cell_avgs = 1.0/cell_size*op.elementwise_integral(dcoll, dd, field)
+    def cancel_polynomials(grp):
+        return actx.from_numpy(np.asarray([1 if sum(mode_id) == 0
+                                           else 0 for mode_id in grp.mode_ids()]))
+
+    # map from nodal to modal
+    discr = dcoll.discr_from_dd(DD_VOLUME_ALL)
+    modal_map = dcoll.connection_from_dds(DD_VOLUME_ALL, DD_VOLUME_ALL_MODAL)
+    nodal_map = dcoll.connection_from_dds(DD_VOLUME_ALL_MODAL, DD_VOLUME_ALL)
+    modal_field = modal_map(field)
```
majosm, line 104:
```suggestion
    dd_nodal = dd
    dd_modal = dd_nodal.with_discr_tag(DISCR_TAG_MODAL)

    discr = dcoll.discr_from_dd(dd_nodal)
    modal_map = dcoll.connection_from_dds(dd_nodal, dd_modal)
    nodal_map = dcoll.connection_from_dds(dd_modal, dd_nodal)
    modal_field = modal_map(field)
```

mirgecom/limiter.py
```diff
@@ -88,14 +92,30 @@ def bound_preserving_limiter(dcoll: DiscretizationCollection, field,
...
+
+    # cancel the ``high-order'' polynomials p > 0, and only the average remains
+    filtered_modal_field = DOFArray(
+        actx,
+        tuple(actx.einsum("ej,j->ej",
+                          vec_i,
+                          cancel_polynomials(grp),
+                          arg_names=("vec", "filter"),
+                          tagged=(FirstAxisIsElementsTag(),))
+              for grp, vec_i in zip(discr.groups, modal_field))
```
majosm, line 114: Should be a modal `discr`, right?

examples/autoignition-mpi.py
```diff
@@ -499,25 +500,75 @@ def my_get_timestep(t, dt, state):
...
+        fluid_state = construct_fluid_state(cv, tseed)
+
+        # apply limiter and reevaluate CV
+        limited_cv = apply_limiter(
+            fluid_state.cv, fluid_state.pressure, fluid_state.temperature)
+
+        # since temperature is not changed during the limiting process, it is
+        # possible to avoid recomputing it by NOT calling make_fluid_state.
+        # Thus, evaluate ONLY other DV and TV variables and reassemble the state
+        new_dv = update_dv(limited_cv, fluid_state.temperature)
```
majosm, line 563: @MTCam will be able to provide more specific guidance on this, but I wonder if it would be better to modify `eos.dependent_vars` to take in an optional `temperature` argument instead of adding a separate `update_dependent_vars` function.

examples/autoignition-mpi.py
```diff
@@ -499,25 +500,75 @@ def my_get_timestep(t, dt, state):
...
+
+        if viscous_terms_on:
+            new_tv = update_tv(limited_cv, new_dv)
+            fluid_state = update_fluid_state(limited_cv, new_dv, new_tv)
+        else:
+            fluid_state = update_fluid_state(limited_cv, new_dv)
+
+        cv = fluid_state.cv
```
majosm, line 571: Maybe condense this and the `dv` update into a single compiled function, say `construct_fluid_state_with_temperature`? Or even the limiter step too and call it e.g. `limit_fluid_state`?

examples/autoignition-mpi.py
```diff
@@ -586,9 +637,20 @@ def my_post_step(step, t, dt, state):
...
+
+        # same description as in pre-step
+        limited_cv = limiter(fluid_state.cv, fluid_state.pressure,
+                             fluid_state.temperature)
+        new_dv = update_dependent_vars(limited_cv, fluid_state.temperature, eos)
+        if viscous_terms_on:
+            new_tv = gas_model.transport.transport_vars(limited_cv, new_dv, eos)
+            fluid_state = ViscousFluidState(limited_cv, new_dv, new_tv)
+        else:
+            fluid_state = FluidState(limited_cv, new_dv)
```
majosm, line 652: Same here (though I guess you would call the non-compiled version).

majosm (commented):
Added a few suggestions/thoughts.

tulioricci:
Thanks for your suggestions @majosm . I implemented all points you raised. I also removed the function from "simutil" and placed it in the driver itself, inside the "limit_fluid_state".

majosm (approved):
Looks good on my end. Let's wait until @MTCam has a chance to look before merging though, as I'm not super familiar with the details of how the limiter works.

examples/autoignition-mpi.py
```diff
@@ -499,25 +499,65 @@ def my_get_timestep(t, dt, state):
...
+            momentum=mass_lim*cv.velocity, species_mass=mass_lim*spec_lim)
+
+        # since temperature nor pressure are changed during the limiting process,
+        # it is possible to avoid recomputing it by NOT calling make_fluid_state.
+        # Thus, evaluate ONLY other DV and TV variables for the limited state
+        new_dv = MixtureDependentVars(
+            temperature=temperature,
+            pressure=pressure,
+            speed_of_sound=eos.sound_speed(limited_cv, temperature),
+            species_enthalpies=eos.species_enthalpies(limited_cv, temperature))
```
MTCam, line 540: Is there a performance reason not to use this:
```suggestion
        new_dv = fluid_state.dv.replace(
               speed_of_sound=eos.sound_speed(limited_cv, temperature)
               species_enthalpies=eos.species_enthalpies(limited_cv, temperature)
        )
```

examples/autoignition-mpi.py
```diff
@@ -499,25 +499,65 @@ def my_get_timestep(t, dt, state):
...
+
+        # update transport vars, if necessary
+        if gas_model.transport is not None:
+            new_tv = gas_model.transport.transport_vars(limited_cv, new_dv, eos)
```
MTCam, line 544: what about `fluid_state.tv.replace` here too?

tulioricci: In another PR, I am not recomputing the viscosity every time (https://github.com/illinois-ceesd/mirgecom/pull/772/files#diff-cc316056955cf6f5921e330e4134bbbb83ca53f817f90ab48548b3ad4453c028R128). So I think this way is better with that case in mind.

examples/autoignition-mpi.py
tulioricci: Not that I am aware. Changing to your suggestion.

PR #782, opened 2022-10-14
add pyrasite to dev packages

matthiasdiener:
[pyrasite](https://github.com/lmacken/pyrasite) can be used to inject python code into *running* python processes, useful e.g. for getting a stack trace at runtime without stopping the application:

```bash
$ pyrasite-shell <PID>
```

```python
import sys, traceback
for thread_id, frame in sys._current_frames().items():
    print("Stack for thread {}".format(thread_id))
    traceback.print_stack(frame)
    print("")
```

This even works with MPI.

---

**Questions for the review**:

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -13,3 +13,5 @@ pep8
 flake8-docstrings
 
 git+https://github.com/vmprof/vmprof-python.git
+
+pyrasite


MTCam (approved):
Yes, please.

inducer (approved):

PR #786, opened 2022-10-25
Python 3.10

matthiasdiener:
https://www.python.org/downloads/release/python-3100/

- match / case statement
- Much better error messages
- zip length checking


matthiasdiener:
Closing in favor of #788.

PR #789, opened 2022-10-28
Replace nozzle production test with prediction test.

MTCam:
This change set:

- drops `nozzle` out of production testing
- adds prediction-driver smoke test
- run the production tests in distributed-lazy mode
- adds script support for running production tests on Lassen and Porter

**Questions for the review**:


matthiasdiener (approved):
LGTM, thanks! Should we also run this on porter (instead of the examples)?

MTCam:
> LGTM, thanks! Should we also run this on porter (instead of the examples)?

How about in addition to?  Hope that is ok.  I think it is useful to do them both, and the porter tests still finish before _pytest_ (I think).

PR #798, opened 2022-11-14
Use production-compatible subpkgs in main.

MTCam:
When main and production have different `requirements.txt`, it frustrates effective testing of production/prediction drivers for upstream packages.  

**Questions for the review**:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,6 @@ mpi4py
 numpy
 pytest
 pytest-cov
-pyvisfile
 pymetis
 importlib-resources
 psutil
@@ -18,9 +17,10 @@ git+https://github.com/pythological/kanren.git#egg=miniKanren
 --editable git+https://github.com/inducer/leap.git#egg=leap
 --editable git+https://github.com/inducer/modepy.git#egg=modepy
 --editable git+https://github.com/inducer/arraycontext.git#egg=arraycontext
---editable git+https://github.com/kaushikcfd/meshmode.git#egg=meshmode
---editable git+https://github.com/inducer/grudge.git#egg=grudge
+--editable git+https://github.com/majosm/meshmode.git@production#egg=meshmode
+--editable git+https://github.com/majosm/grudge.git@coupled-volumes#egg=grudge
 --editable git+https://github.com/kaushikcfd/pytato.git#egg=pytato
 --editable git+https://github.com/pyrometheus/pyrometheus.git#egg=pyrometheus
 --editable git+https://github.com/illinois-ceesd/logpyle.git#egg=logpyle
 --editable git+https://github.com/kaushikcfd/feinsum.git#egg=feinsum
+--editable git+https://github.com/inducer/pyvisfile.git#egg=pyvisfile


MTCam:
> > When main and production have different `requirements.txt`, it frustrates effective testing of production/prediction drivers for upstream packages.
> 
> Can you explain this more? Not sure I'm understanding.

Upstream packages (e.g. `grudge`) should test `production` drivers in their tests of downstream packages.  Doing this will plug a hole in our current testing that has resulted in broken production many times over.   The upstream packages currently test only the mirgecom examples found in `mirgecom@main`.

The downstream testing done in `grudge` (for example) installs `mirgecom` by taking in and editing the `mirgecom/requirements.txt@main` so that the version of `grudge` being tested against is not overwritten by the `mirgecom` install process.   When there are changes to `requirements.txt` between `mirgecom@main` and `mirgecom@production`, the upstream package CI needs to be aware and re-install `mirgecom@production` in the same way (i.e. including editing `requirements.txt`) if it wants to run the "production tests".  

Alternatively: `mirgecom@main` can always use a prediction-compatible `requirements.txt` and then the upstream CI's get the correct `mirgecom/requirements.txt` file from the beginning, and no "insider" information into `mirgecom@production` is required.

inducer:
I mean, maybe downstream testing of mirgecom doesn't make sense? If there's a hard dependency on features not merged into `main` of the base packages, then that would seem to be the case. My objective with providing downstream testing is to help avoid breakage, not to inflict pain. So I'm just as happy to stop doing this.

MTCam:
> I mean, maybe downstream testing of mirgecom doesn't make sense? If there's a hard dependency on features not merged into `main` of the base packages, then that would seem to be the case. My objective with providing downstream testing is to help avoid breakage, not to inflict pain. So I'm just as happy to stop doing this.

I don't think I understand how this issue would suggest, or warrant the disabling of downstream testing.  That said, I do not disagree with the notion that downstream testing doesn't really make sense - but I am loath to eliminate any checks we have that catch breakage before it impacts our prediction simulation capability.

Surely we need some testing to ensure incoming changes in the upstream packages don't break mirgecom@production.  Currently upstream packages are testing examples@main only - which does prevent _some_ breakage - so I think there is value in keeping those tests.   We frequently experience breakage to production cases due to changes in upstream pkgs still, so I am hoping to plug that testing hole by bringing production testing into the suite of things tested by the upstream pkgs.

If all downstream testing of mirgecom is disabled, then we need to pick up the slack on the mirgecom side.  To do that effectively, I think we'd need to pin specific hashes in our "working/production" `requirements.txt` and track a separate `testing` branch or so that points at development HEADs.

majosm:
> The downstream testing done in `grudge` (for example) installs `mirgecom` by taking in and editing the `mirgecom/requirements.txt@main` so that the version of `grudge` being tested against is not overwritten by the `mirgecom` install process.

Question: what happens in a grudge PR when the mirgecom requirements.txt entry for grudge isn't main? Does that branch get replaced with the PR branch, or does it attempt to merge the PR changes in?

> When there are changes to `requirements.txt` between `mirgecom@main` and `mirgecom@production`, the upstream package CI needs to be aware and re-install `mirgecom@production` in the same way (i.e. including editing `requirements.txt`) if it wants to run the "production tests".
> 
> Alternatively: `mirgecom@main` can always use a prediction-compatible `requirements.txt` and then the upstream CI's get the correct `mirgecom/requirements.txt` file from the beginning, and no "insider" information into `mirgecom@production` is required.

I think this would prevent us from ever pointing to a different branch in production vs. main, wouldn't it? Not sure if we want to commit to that?

majosm:
> AFAIK (@inducer?): It does not merge; it just replaces the "grudge" line with the current development to be tested. Here are some of relevant constructs:
> 
> * Grudge CI where the `requirements` tweaking begins: https://github.com/inducer/grudge/blob/8df43f44ca9d2900d26fd22e183917f91538247d/.github/workflows/ci.yml#L117
> * Utility that edits the `requirements` file:
>   https://gitlab.tiker.net/inducer/ci-support/-/blob/main/ci-support.sh#L679

That sounds pretty fragile... if mirgecom requires changes from a non-main branch in order to work, and the PR being tested doesn't include those changes, stuff could break. To me this seems like an argument that we should strive not to modify requirements.txt in mirgecom's main branch.

> Yes, that agrees with my understanding of it. It essentially amounts to the rule that no PR in the production feeder chain may change `requirements.txt`. Meaning... if there is a change of upstream packages required to do our prediction, we _require_ that this change is made in `mirgecom@main` _not_ down in the feeder PRs. I think this would actually be a positive change in our process, and reduce some of the tangle.

I think this is about as realistic as requiring that we merge all of our mirgecom feeder PRs into main before running the prediction (i.e. not very). 🙂 IMO production exists for stuff like this, if we attempt to constrain it too much we'll just end up with unofficial production branches floating around elsewhere.

MTCam:
> > AFAIK (@inducer?): It does not merge; it just replaces the "grudge" line with the current development to be tested.
> > (...)
> 
> That sounds pretty fragile... if mirgecom requires changes from a non-main branch in order to work, and the PR being tested doesn't include those changes, stuff could break. To me this seems like an argument that we should strive not to modify requirements.txt in mirgecom's main branch.

It seems like it is not fragile per se; it is just insufficient, or has this one foible that it doesn't account for the subproject being tested maybe requiring changes to the local package being tested.  @inducer has been reliably using this to test downstream projects effectively and has kind of "standardized" it across his own projects (which is what leads me to say it isn't fragile).      

> 
> I think this is about as realistic as requiring that we merge all of our mirgecom feeder PRs into main before running the prediction (i.e. not very). 🙂 IMO production exists for stuff like this, if we attempt to constrain it too much we'll just end up with unofficial production branches floating around elsewhere.

I see your thinking, but I don't see it that way.  I am saying that the `requirements@main` should always be that which our production/prediction requires.  Other PRs can continue to change their `requirements.txt` as they require, but if the PR is to become a "feeder" branch for the CEESD prediction, then it must work on `requirements.txt@main`.   

This policy would give us much better visibility on what sub-packages are required for prediction, and on more eyes on  when and why those `requirements.txt` change. We are still allowed to change it, and change it in PRs, but changes to the one used for production/prediction must go through the gate of review before they are used because they must be in  `requirements.txt@main`.   

This is the same ultimate policy we are pursuing with the code; to run the prediction out of main.  Eliminating the difference between `requirements@main` vs. `requirements@production` seems like a low-hanging way to permanently reduce one chunk of the difference between those two "versions" of the code.  I agree it is more difficult to think we can ever eliminate all the differences between main and production branch, at least wrt the code.  Unrealistic? yeah, that's fair - but still an ideal situation, and useful as a goal.

majosm:
> It seems like it is not fragile per se; it is just insufficient, or has this one foible that it doesn't account for the subproject being tested maybe requiring changes to the local package being tested. @inducer has been reliably using this to test downstream projects effectively and has kind of "standardized" it across his own projects (which is what leads me to say it isn't fragile).

The fragile part is allowing requirements.txt of main to point at non-main branches of subpackages. This is not something we've been doing for long, and I think it's only working with the CI more or less by coincidence right now.

> I see your thinking, but I don't see it that way. I am saying that the `requirements@main` should always be that which our production/prediction requires. Other PRs can continue to change their `requirements.txt` as they require, but if the PR is to become a "feeder" branch for the CEESD prediction, then it must work on `requirements.txt@main`.
> 
> This policy would give us much better visibility on what sub-packages are required for prediction, and on more eyes on when and why those `requirements.txt` change. We are still allowed to change it, and change it in PRs, but changes to the one used for production/prediction must go through the gate of review before they are used because they must be in `requirements.txt@main`.
> 
> This is the same ultimate policy we are pursuing with the code; to run the prediction out of main. Eliminating the difference between `requirements@main` vs. `requirements@production` seems like a low-hanging way to permanently reduce one chunk of the difference between those two "versions" of the code. I agree it is more difficult to think we can ever eliminate all the differences between main and production branch, at least wrt the code. Unrealistic? yeah, that's fair - but still an ideal situation, and useful as a goal.

I really don't think we should continue messing with requirement.txt in main. Fusion should be treated as a one-off situation and after that gets resolved we can go back to having requirements.txt point to the stable/tested subpackage versions (i.e. main) for everything.

Continually pushing towards running out of main certainly makes sense, but trying to rigidly enforce it as a policy for subpackages is arbitrary and not helpful IMO.

PR #805, opened 2022-12-07
Update Mengaldo BCs

tulioricci:
Add no-shear stress condition for Symmetry/slip wall
Add RiemannInflow,
Add linearized Euler Eqs. for outflow
Rename OutflowBoundary to PressureOutflowBoundary

**Questions for the review**:

### DIFF
--- a/mirgecom/boundary.py
+++ b/mirgecom/boundary.py
@@ -988,7 +988,7 @@ class RiemannInflowBoundary(PrescribedFluidBoundary):
     .. automethod:: inflow_state
     """
 
-    def __init__(self, dim, free_stream_state_func):
+    def __init__(self, free_stream_state_func):
         """Initialize the boundary condition object."""
         self.free_stream_state_func = free_stream_state_func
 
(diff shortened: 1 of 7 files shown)


mirgecom/boundary.py
```diff
@@ -1416,8 +1529,112 @@ def adiabatic_slip_grad_av(self, dcoll, dd_bdry, grad_av_minus, **kwargs):
...
+                 free_stream_pressure=None,
+                 free_stream_temperature=None,
+                 free_stream_species_mass_fractions=None):
+        """Initialize the boundary condition object."""
+        self._ref_state = free_stream_state
+        self._mass = free_stream_density
+        self._velocity = free_stream_velocity
+        self._pressure = free_stream_pressure
+        self._temperature = free_stream_temperature
+        self._Y = free_stream_species_mass_fractions
```
MTCam, line 1577: How's come this?   Why not s'th like:
```suggestion
if free_stream_state is not None:
   self._mass = free_stream_state.mass
   (...) etc
```
... and then eliminate the if/else blocks that check for existence of `self._ref_state`?

PR #828, opened 2023-01-27
warn when cache dirs are not rank-specific on a node

matthiasdiener:
**Questions for the review**:

### DIFF
--- a/mirgecom/mpi.py
+++ b/mirgecom/mpi.py
@@ -83,7 +83,7 @@ def _check_cache_dirs() -> None:
                 assert all_paths
                 if len(all_paths) != len(set(all_paths)):
                     hostname = MPI.Get_processor_name()
-                    dup = [path for path in all_paths if all_paths.count(path) > 1]
+                    dup = [path for path in set(all_paths) if all_paths.count(path) > 1]
 
                     from warnings import warn
                     warn(f"Multiple ranks are sharing '{var}' on node '{hostname}'. "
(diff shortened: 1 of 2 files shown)


matthiasdiener:
This is ready for review. Here is how it looks like in CI:

`
/home/runner/work/mirgecom/mirgecom/.miniforge3/envs/testing/lib/python3.11/site-packages/mirgecom-2020.1-py3.11.egg/mirgecom/mpi.py:89: UserWarning: Multiple ranks are sharing 'XDG_CACHE_HOME' on node 'fv-az447-593'. Duplicate 'XDG_CACHE_HOME's: ['/tmp', '/tmp'].
`

mirgecom/mpi.py
```diff
@@ -49,7 +52,52 @@ def shared_split_comm_world():
...
+                # Create a fake path so there will not be a second warning below.
+                my_path = f"no/such/path/rank{node_rank}"
+
+            all_paths = node_comm.gather(my_path, root=0)
+
+            if node_rank == 0:
+                assert all_paths
+                if len(all_paths) != len(set(all_paths)):
+                    hostname = MPI.Get_processor_name()
+                    dup = [path for path in all_paths if all_paths.count(path) > 1]
```
majosm, line 86:
```suggestion
                    dup = [path for path in set(all_paths) if all_paths.count(path) > 1]
```

majosm (approved):
👍 Looks good to me, apart from the one suggestion.

PR #837, opened 2023-02-15
Update filter testing

MTCam:
This change set updates the testing done for filtering.  The old broken test is removed and a new test that checks the polynomial expansion coefficients against the exact known result of filtering is added.   

A commented out section in the testing allows for monomial fields for hand-testing and making spectral plots that are easier to analyze for correctness because they have more intuitive spectra.   We may want to revive these in the near future, so they are left commented for the time being.

**Questions for the review**:


test/test_filter.py
```diff
@@ -120,136 +115,197 @@ def test_filter_coeff(actx_factory, filter_order, order, dim):
...
+    # print(f"{unfiltered_spectra=}")
+    # print(f"{spectra_numbers.shape=}")
+    # print(f"{spectra_numbers=}")
+
+    element_spectra = make_obj_array([
+        [np.array([0. for i in range(element_order+1)])
+         for j in range(numfields)] for k in range(numelem)])
+
+    total_power = [np.array([0. for _ in range(element_order+1)])
+                   for e in range(numfields)]
```
majosm, line 243: It looks like this and the code below could be simplified by using multidimensional numpy arrays instead of these nested lists of arrays. At a glance I can't tell if there's anything getting in the way of doing that or not. For example, if you could find a way to compute `element_spectra` with numpy operations, then I think `total_power` would just be something like `sum(element_spectra, axis=0)`.

majosm (approved):
Makes sense to me. Just added a few suggestions that might clean things up a bit.

test/test_filter.py
```diff
@@ -68,19 +65,17 @@ def test_filter_coeff(actx_factory, filter_order, order, dim):
...
-    # JSH/TW Nodal DG Methods, Section 10.1
-    # DOI: 10.1007/978-0-387-72067-8
-    nmodes = 1
-    for d in range(1, dim+1):
-        nmodes *= (order + d)
-    nmodes /= math.factorial(int(dim))
-    nmodes = int(nmodes)
+
+    from mirgecom.simutil import get_number_of_tetrahedron_nodes
+    nmodes = get_number_of_tetrahedron_nodes(dim, order)
```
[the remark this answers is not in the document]
MTCam: Not sure we should do that here.  Let's defer and discuss.

test/test_filter.py
```diff
@@ -120,136 +115,197 @@ def test_filter_coeff(actx_factory, filter_order, order, dim):
...
+
+    element_spectra = make_obj_array([
+        [np.array([0. for i in range(element_order+1)])
+         for j in range(numfields)] for k in range(numelem)])
+
+    total_power = [np.array([0. for _ in range(element_order+1)])
+                   for e in range(numfields)]
+
+    for fldi in range(numfields):
+        field_spectra = spectra_numbers[fldi][0][0]
```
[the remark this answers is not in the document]
MTCam: Correct. It is done.

test/test_filter.py
majosm: I took a stab at this in [8a2bb98](https://github.com/illinois-ceesd/mirgecom/pull/837/commits/8a2bb98759b4e2338e3b97f7d45e373ef1e9e2a6). Feel free to revert/edit as desired.

MTCam: Outstanding!  Thanks!

MTCam: Please see what you think of this - just moving your changes off into a utility that we can use outside of testing.
([71f1708](https://github.com/illinois-ceesd/mirgecom/pull/837/commits/71f17082ab63973782b097c5a93a9b007c8e162d))

mirgecom/filter.py
```diff
@@ -61,6 +65,52 @@
...
+    accumulated_spectra = np.zeros(
+        (numfields, numelem, element_order+1), dtype=np.float64)
+
+    for i in range(nummodes):
+        accumulated_spectra[:, :, emodes_to_pmodes[i]] += np.abs(
+            modal_spectra[:, :, i])
+
+    return accumulated_spectra
+
+
```
majosm: Question: how is @anderson2981 planning to use this? During timestepping? If so, maybe we should rewrite this again to use `actx` arrays/lazy.

MTCam: Just as a debugging/diagnostic tool for now.  This is an improvement on what I gave him earlier.

PR #843, opened 2023-02-23
add py.typed marker to "export" type info

matthiasdiener:
With this PR, other packages using mirgecom can use mypy to check for type conformance with the mirgecom API.
See https://peps.python.org/pep-0561/ for context.


matthiasdiener:
Not sure what's happening in the production CI, but this is ready to go.

MTCam (approved):

PR #849, opened 2023-03-08
Make RHS compile optional in `advance_state`

majosm:
Needed for https://github.com/illinois-ceesd/drivers_y3-prediction/pull/6.

**Questions for the review**:

### DIFF
--- a/mirgecom/steppers.py
+++ b/mirgecom/steppers.py
@@ -76,7 +76,7 @@ def _is_unevaluated(actx, ary):
 def _advance_state_stepper_func(rhs, timestepper, state, t_final, dt=0,
                                 t=0.0, istep=0, pre_step_callback=None,
                                 post_step_callback=None, force_eval=None,
-                                local_dt=False, max_steps=None, compile_rhs=None):
+                                local_dt=False, max_steps=None, compile_rhs=True):
     """Advance state from some time (t) to some time (t_final).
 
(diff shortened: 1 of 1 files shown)


mirgecom/steppers.py
```diff
@@ -76,7 +76,7 @@ def _is_unevaluated(actx, ary):
 def _advance_state_stepper_func(rhs, timestepper, state, t_final, dt=0,
                                 t=0.0, istep=0, pre_step_callback=None,
                                 post_step_callback=None, force_eval=None,
-                                local_dt=False, max_steps=None):
+                                local_dt=False, max_steps=None, compile_rhs=None):
```
matthiasdiener, line 79: Should this be `False` by default (same for leap)?

majosm: Strictly speaking the default value doesn't really matter for these, since they're only called from `advance_state` and it always passes its own value for `compile_rhs`. I couldn't tell if the convention was to default to `None` (like `force_eval`) or `True`/`False` (like `local_dt`) so I just picked one. I can switch to the other if that would be preferable.

matthiasdiener: mypy would likely complain if you add a `:bool` annotation, so I'd just go ahead and make it an actual bool instead.

matthiasdiener (approved):

PR #870, opened 2023-04-08
Modify MxTransp to handle single species case

tulioricci:
**Questions for the review**:

### DIFF
--- a/mirgecom/transport.py
+++ b/mirgecom/transport.py
@@ -476,9 +476,9 @@ class MixtureAveragedTransport(TransportModel):
         actx = cv.mass.array_context
 
         diffusivity = self._pyro_mech.get_species_mass_diffusivities_mixavg(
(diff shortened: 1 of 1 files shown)


mirgecom/transport.py
```diff
@@ -489,10 +467,30 @@ def species_diffusivity(self, cv: ConservedVars,  # type: ignore[override]
...
+        diffusivity = self._pyro_mech.get_species_mass_diffusivities_mixavg(
+            dv.pressure, dv.temperature, cv.species_mass_fractions)
+
+        # floor to avoid single-species case breaking the mixture rule
+        epsilon = 1e-4
+
+        # if one species dominate, uses a small diffusivity.
+        # The actual number should't matter since, in the single-species case,
+        # diffusion is proportional to a nearly zero-gradient.
+        dummy_diffusivity = 1e-6
```
MTCam, line 482: Can we make these class parameters so that the user can set them if desired?

PR #255, opened 2021-02-18
logging: add GPU memory consumption on Nvidia

matthiasdiener:

PR #874, opened 2023-04-12
warn when memory usage is high on the GPU

matthiasdiener:
**Questions for the review**:

### DIFF
--- a/mirgecom/logging_quantities.py
+++ b/mirgecom/logging_quantities.py
@@ -404,8 +404,6 @@ class DeviceMemoryUsage(PostLogQuantity):
         super().__init__(name, "MByte", description="Memory usage (GPU)")
 
         import ctypes
-        self.free = ctypes.c_size_t()
-        self.total = ctypes.c_size_t()
 
         try:
             # See https://gist.github.com/f0k/63a664160d016a491b2cbea15913d549#gistcomment-3654335  # noqa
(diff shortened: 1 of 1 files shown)


matthiasdiener:
This is a relatively crude way of checking (in particular, it doesn't capture if there are temporary allocations within the time step that exceed the memory size), but I couldn't come up with a better way to do this.

mirgecom/logging_quantities.py
```diff
@@ -428,6 +428,13 @@ def __call__(self) -> Optional[float]:
             warn(f"cudaMemGetInfo failed with error {ret}.")
             return None
         else:
+            if self.free.value / self.total.value < 0.1:
```
inducer, line 431: It doesn't belong here, but :shrug: IDK.

inducer, line 431: While I'm at it, why are `self.free` and `self.total` instance attributes? I don't see a need for them to be long-lived.

MTCam (approved):
In general, I like this warning.   Not intending to subvert @inducer's concerns.  I don't see anything that should prevent merge.

mirgecom/logging_quantities.py
matthiasdiener: That's a good catch, thanks! Unfortunately, I can't remember why I wrote the code this way in #255 (possibly to reduce overhead from the `ctypes.c_size_t()` calls?). I can't see a performance difference with [844db39](https://github.com/illinois-ceesd/mirgecom/pull/874/commits/844db3950e53f4a04fdd24b6363b2d447de2d348), so it should be ok to change these to local variables.

inducer:
Don't let me hold up the merge here, I think this is fair enough.

PR #878, opened 2023-04-18
Changes to examples in production

MTCam:
These are the various changes to the examples that were made in production branch.   Most changes were made downstream of main before main was able to run the example with the change/new feature in-place.  This change-set will update the examples to their "production" forms.

- combozzle (y1+-prediction-like scalable test case) is added - lazy enabled
- lazy is enabled for inviscid mixture case
- scalar adv/diff case is added
- poiseuille multispecies case is added
- nsmix is updated to enable different boundaries, overintegration

**Questions for the review**:


matthiasdiener (approved):

examples/mixture-mpi.py
```diff
@@ -378,12 +404,16 @@ def my_post_step(step, t, dt, state):
             logmgr.tick_after()
         return make_obj_array([fluid_state.cv, tseed]), dt
 
+    def dummy_rhs(t, state):
+        return make_obj_array([0*state[0], 0*state[1]])
+
```
majosm, line 409: Looks like the `dummy_*` functions are unused here? (I see the ones in `combozzle-mpi` being used but not these.)

examples/scalar-advdiff-mpi.py
```diff
@@ -0,0 +1,441 @@
...
+        periodic = (True, )*dim
+        from meshmode.mesh.generation import generate_regular_rect_mesh
+        generate_mesh = partial(generate_regular_rect_mesh, a=(box_ll,)*dim,
+                                b=(box_ur,) * dim, nelements_per_axis=(nel_1d,)*dim,
+                                periodic=periodic)
+        local_mesh, global_nelements = generate_and_distribute_mesh(comm,
+                                                                    generate_mesh)
+        local_nelements = local_mesh.nelements
+
+    dcoll = EagerDGDiscretization(
```
majosm, line 154: Update?

PR #882, opened 2023-04-21
update AV2 transport class 

anderson2981:
Don't use negative values of the smoothness indicator.

This can cause negative results for transport properties.

**Questions for the review** @MTCam :


mirgecom/transport.py
```diff
@@ -640,14 +640,20 @@ def bulk_viscosity(self, cv: ConservedVars,  # type: ignore[override]
                        dv: GasDependentVars,
                        eos: GasEOS) -> DOFArray:
         r"""Get the bulk viscosity for the gas, $\mu_{B}$."""
-        return (dv.smoothness_beta*self.av_beta(cv, dv, eos)
+        actx = cv.array_context
+        smoothness_beta = actx.np.where(
+            actx.np.greater(dv.smoothness_beta, 0.), dv.smoothness_beta, 0.)
+        return (smoothness_beta*self.av_beta(cv, dv, eos)
```
MTCam, line 646: notice any performance difference?

MTCam (approved):

PR #891, opened 2023-04-27
Add limiter_func to interface and pass to make_operator_fluid_states

MTCam:
Currently, multiphysics runs with mixtures are sending unlimited fluid states to the fluid operators.   This change set fixes that issue.

**Questions for the review**:

### DIFF
--- a/mirgecom/multiphysics/thermally_coupled_fluid_wall.py
+++ b/mirgecom/multiphysics/thermally_coupled_fluid_wall.py
@@ -1165,7 +1165,7 @@ def coupled_ns_heat_operator(
     limiter_func:
 
         Callable function to be passed to
-        :func:`~mirgecom.gas_model.make_fluid_operator_states`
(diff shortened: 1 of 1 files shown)


anderson2981 (approved):
Looks good to me.

majosm (approved):
Good catch. Thanks!

mirgecom/multiphysics/thermally_coupled_fluid_wall.py
```diff
@@ -1165,7 +1165,7 @@ def coupled_ns_heat_operator(
     limiter_func:
 
         Callable function to be passed to
-        :func:`~mirgecom.gas_model.make_fluid_operator_states`
```
MTCam: I like the misspelled name better.

tulioricci:
Should we also include the limiter here?
https://github.com/illinois-ceesd/mirgecom/blob/dc19f7c98fa6003e870737d4582472eee9c4f56f/mirgecom/navierstokes.py#L436

PR #892, opened 2023-04-28
Deprecate duplicate `force_evaluation`

majosm:
Turns out we have two versions of `force_evaluation`, one in `utils` and one in `simutil`. I think the one in `utils` is the preferred implementation (it uses `freeze_thaw` instead of `thaw(freeze(...))`), so let's keep that one.

**Questions for the review**:


MTCam:
Yes, the implementation of the one in utils is preferred.  imo, we should move it to `simutil`.

MTCam (approved):

PR #893, opened 2023-04-28
Add radiation

majosm:
Adds radiation sink term to fluid-wall coupling.

**Questions for the review**:


tulioricci:
@majosm , two comments:

1) should we standardize the nomenclature of BCs? Right now, this is what we have:
IsothermalSlipWallBoundary
IsothermalWallBoundary
AdiabaticSlipBoundary
AdiabaticNoslipWallBoundary

2) replace "wall_epsilon" with "wall_emissivity"

majosm:
> 2. replace "wall_epsilon" with "wall_emissivity"

Done (forgot to push 🙂).

MTCam:
> @majosm , two comments:
> 
> 1. should we standardize the nomenclature of BCs? Right now, this is what we have:
>    IsothermalSlipWallBoundary
>    IsothermalWallBoundary
>    AdiabaticSlipBoundary
>    AdiabaticNoslipWallBoundary

I dislike the disparity in naming, too.   I also dislike when every BC has "WallBoundary" attached to it, if they _all_ have it, then why have it at all?  For me, these could easily be `IsothermalSlip`, `IsothermalNoslip` , etc.   It's already known as a BC because we `from mirgecom.boundary import IsothermalNoslip`.

Now that we have multiple domains and materials, we could re-think some of the code structuring too, like renaming mirgecom.boundary to mirgecom.fluid_boundary.   But again, these are all non-essential changes that can easily distract us and seem unimportant unless we are already set up to really nail the prediction.

mirgecom/multiphysics/thermally_coupled_fluid_wall.py
```diff
@@ -1381,6 +1608,10 @@ def coupled_ns_heat_operator(
             fluid_grad_temperature, wall_grad_temperature,
             fluid_boundaries, wall_boundaries,
             interface_noslip=interface_noslip,
+            interface_radiation=interface_radiation,
+            wall_emissivity=wall_emissivity,
+            sigma=sigma,
+            ambient_temperature=ambient_temperature,
```
tulioricci, line 1614: If the function is deprecated, I wouldn't updated it with radiation.

tulioricci:
And we don't need to get the gradient here (https://github.com/illinois-ceesd/mirgecom/blob/d512ebd2a0690a9ad1c5b1581ac59c360bee34bf/mirgecom/multiphysics/thermally_coupled_fluid_wall.py#L1839)
because it is already computed above

mirgecom/multiphysics/thermally_coupled_fluid_wall.py
```diff
@@ -1421,6 +1652,10 @@ def basic_coupled_ns_heat_operator(
         *,
         time=0.,
         interface_noslip=True,
+        interface_radiation=False,
+        wall_emissivity=None,
+        sigma=None,
```
tulioricci, line 1657: Should we default it to its actual value 5.67e-8?

mirgecom/multiphysics/thermally_coupled_fluid_wall.py
majosm: https://github.com/illinois-ceesd/mirgecom/pull/893/commits/31ce3239f59a815d21a079a87c047e9a48e37431

majosm:
> And we don't need to get the gradient here (
> 
> https://github.com/illinois-ceesd/mirgecom/blob/d512ebd2a0690a9ad1c5b1581ac59c360bee34bf/mirgecom/multiphysics/thermally_coupled_fluid_wall.py#L1839
> 
> )
> because it is already computed above

It's not actually being recomputed there, since we're passing it in as `grad_u` (and similarly for the fluid side). So basically it's just cosmetic. I tried removing it (code [here](https://github.com/majosm/mirgecom/commit/8a2000f44c07a7cdbf404b09764b381a7c4270c1)), but I don't know if it's any cleaner. 🤷‍♂️ Thoughts?

mirgecom/multiphysics/thermally_coupled_fluid_wall.py
majosm: I worry about specifying default values for constants, because it invites the possibility of omitting them by accident, leading to silently incorrect results if the value is supposed to be something other than the default. Better to produce an error I think.

tulioricci:
> > And we don't need to get the gradient here (
> > https://github.com/illinois-ceesd/mirgecom/blob/d512ebd2a0690a9ad1c5b1581ac59c360bee34bf/mirgecom/multiphysics/thermally_coupled_fluid_wall.py#L1839
> > 
> > )
> > because it is already computed above
> 
> It's not actually being recomputed there, since we're passing it in as `grad_u` (and similarly for the fluid side). So basically it's just cosmetic. I tried removing it (code [here](https://github.com/majosm/mirgecom/commit/8a2000f44c07a7cdbf404b09764b381a7c4270c1)), but I don't know if it's any cleaner. man_shrugging Thoughts?

I prefer the new version, but it is not a big deal since, in practice, we may end up having this in the driver itself... man_shrugging

tulioricci:
LGTM, but I will let @MTCam give the final word.

MTCam (approved):
👍

PR #894, opened 2023-05-01
Pass through inviscid numerical flux func.

MTCam:
~~Fixes~~ Mitigates a bug in the multiphysics interface that prevents callers from setting `inviscid_numerical_flux_func`.  

Defaulting this causes a pretty harmful thing.   If nothing is passed as `inviscid_numerical_flux_func` to `ns_operator`, then it defaults to `inviscid_facial_flux_rusanov`.    But if nothing is passed  as `inviscid_numerical_flux_func` to `entropy_stable_ns_operator`, it defaults to `entropy_stable_inviscid_flux_rusanov`.    

By internally defaulting this parameter, the multiphysics operator actually changes the expected default behavior.  If nothing is passed as `inviscid_numerical_flux_func` to the coupled operator, then it passes `invsicid_numerical_flux_func=inviscid_facial_flux_rusanov` regardless of whether we use the entropy stable or the standard NS operator.  This will fail for entropy stable.

On the other hand, not fixing this bug renders the driver setting of `inviscid_numerical_flux_func` ineffective.  For the prediction driver, we _always_ pass the `inviscid_numerical_flux_func` argument.  We do not default this setting, including for entropy stable, so fixing it like this will make it work as-intended for prediction. 

One thing we _could_ do is make `*_numerical_flux_func` required arguments of the coupled operator.

@majosm , @anderson2981 

**Questions for the review**:

### DIFF
--- a/mirgecom/multiphysics/thermally_coupled_fluid_wall.py
+++ b/mirgecom/multiphysics/thermally_coupled_fluid_wall.py
@@ -1285,6 +1285,7 @@ def coupled_ns_heat_operator(
     ns_result = ns_operator(
         dcoll, gas_model, fluid_state, fluid_all_boundaries,
         time=time, quadrature_tag=quadrature_tag, dd=fluid_dd,
+        inviscid_numerical_flux_func=inviscid_numerical_flux_func,
         viscous_numerical_flux_func=viscous_numerical_flux_func,
         return_gradients=return_gradients,
         operator_states_quad=fluid_operator_states_quad,


MTCam:
> ```python
> if inviscid_numerical_flux_func is not None:
>     ns_result = ns_operator(
>         ..., inviscid_numerical_flux_func=inviscid_numerical_flux_func)
> else:
>     ns_result = ns_operator(...)
> ```

I kind of dislike this (in-particular the switching between calling forms based on user arguments).  There is another user-settable flux function, too; `viscous_numerical_flux_func`, which, if set, faces the same sorts of issues and complicates the `if this/then that/else other` logic and code structure required of the coupled operator in order to deal with this.

The more I think about it, the more I would lean on making them required arguments for using the coupled operator, and forcing the user to decide what to use (and pass) for them.  Without that, I think we'd need logic down in the coupled operator to decide what to pass, and this is less attractive from my perspective. 

Looking at this has made me super paranoid that in our code we have places where `something_something=None` gets sent as an argument to a function, overriding a default argument that would-have-been valid had we not passed actual `None`.

PR #895, opened 2023-05-02
Allow `kappa` to be either a scalar or a vector in vectorized `diffusion_operator`

majosm:
Tweaks `diffusion_operator` so that in the case of `u` being an object array, `kappa` can either be a scalar (existing behavior) or an object array.

**Questions for the review**:


tulioricci:
Looks good to me.

tulioricci (approved):
LGTM.

PR #902, opened 2023-05-16
Attempt to simplify/generalize coupling

majosm:
Deprecates operator definitions in thermal coupling code and refocuses the API on setting up the interface boundaries. This allows for more flexibility on the driver side (e.g., more than 2 volumes, easier access to gradients, etc.).

cc @tulioricci

**Questions for the review**:


majosm:
@tulioricci, @MTCam I think this is ready for a look through. (I'll do the production merge after.)

examples/thermally-coupled-mpi.py
```diff
@@ -85,6 +79,82 @@ class MyRuntimeError(RuntimeError):
...
+    if return_gradients:
+        fluid_rhs, fluid_grad_cv, fluid_grad_temperature = ns_result
+        wall_rhs, wall_grad_temperature = diffusion_result
+        return (
+            fluid_rhs, wall_rhs, fluid_grad_cv, fluid_grad_temperature,
+            wall_grad_temperature)
+    else:
+        return ns_result, diffusion_result
+
+
```
MTCam, line 157: I like this spirit of this change set, but I'm slightly concerned that this causes most of what is highlighted here to be repeated in every coupled driver, adding to an already onerous driver maintenance situation.

Can any of this be offered as a canned interface?

tulioricci: I was thinking the exact same thing. I would keep the operator instead of deprecating it.

mirgecom/multiphysics/thermally_coupled_fluid_wall.py
```diff
@@ -783,112 +894,149 @@ def get_interface_boundaries(
...
+
+    comm_tag: Hashable
+        Tag for distributed communication
     """
-    if interface_noslip:
-        fluid_bc_class = InterfaceFluidNoslipBoundary
-    else:
-        fluid_bc_class = InterfaceFluidSlipBoundary
+    fluid_interface_boundaries_no_grad, wall_interface_boundaries_no_grad = \
+        _get_interface_boundaries_no_grad(
```
tulioricci, line 902: Is it necessary that this tiny function calls a bigger function? Why not have one function only?

mirgecom/multiphysics/thermally_coupled_fluid_wall.py
```diff
@@ -783,112 +894,149 @@ def get_interface_boundaries(
...
+    quadrature_tag
+
+        An identifier denoting a particular quadrature discretization to use during
+        operator evaluations.
+
+    comm_tag: Hashable
+        Tag for distributed communication
+    """
+    fluid_interface_boundaries, wall_interface_boundaries = \
+        _get_interface_boundaries(
```
tulioricci, line 1018: Same as above. Why two functions?

mirgecom/multiphysics/thermally_coupled_fluid_wall.py
MTCam: I rather like this encapsulation for the interface getters.

mirgecom/multiphysics/thermally_coupled_fluid_wall.py
```diff
@@ -1174,6 +1316,19 @@ def coupled_ns_heat_operator(
...
         The tuple `(fluid_rhs, wall_rhs)`.
     """
+    from warnings import warn
+    warn(
+        "coupled_ns_heat_operator is deprecated and will disappear in Q3 2023. "
+        "Set up interface boundaries explicitly via"
+        ":func:`add_interface_boundaries` and include them when calling the "
+        "individual operators instead.", DeprecationWarning, stacklevel=2)
+
+    if not use_kappa_weighted_grad_flux_in_fluid:
```
tulioricci, line 1326: Why don't we remove it right now and make the warning says that it is being ignored?

MTCam (approved):
Overall this is great.  Only one slight concern from my perspective, but nothing preventing merge I think.

examples/thermally-coupled-mpi.py
majosm: Hmm. Well, the primary motivation for removing it was that it was proving difficult to come up with a one-size-fits-all operator definition (because simulations may have different numbers of volumes, or they may want to precompute things like the operator states and/or gradients). Seemed the most reasonable to just have the driver writer decide how they want to call the operators. I could leave this in as a default/example, but I think I would want to rename it to emphasize that (`basic_ns_heat_operator` or something?). Thoughts?

mirgecom/multiphysics/thermally_coupled_fluid_wall.py
majosm: I don't have a strong opinion about this. I suspect at one point the `_get_*` functions were called in multiple places, but now that they're only called from the corresponding `add_*` it looks like they could be safely removed. You guys decide what I should do and I'll do it. 🙂

examples/thermally-coupled-mpi.py
MTCam: I suspected this would be the case.   The `basic` operator idea is fine for me; on the other hand the example also helps with this.  Aren't the stand-burner case and the prediction case identical in this respect?   (er couldn't they be?).  If so, I would actually prefer for this function to be centralized for both those drivers, and only re-implemented in each driver that needs to re-implement them.

Also, these are just suggestions - nothing show-stopping in my opinion.

tulioricci: For the prediction/burner we also need the NS coupling, not only the heat equation coupling. Having the triple coupling is what motivated this PR. So I wouldn't worry in making this particular coupling too focused on the more complex drivers we have.

MTCam: It seems any way we slice it, we're going to have many versions of this coupled operator.  The more i read it, the shorter it gets :)

mirgecom/multiphysics/thermally_coupled_fluid_wall.py
tulioricci: We decided that you can leave it as it is. Moreover, we decided that you have to buy as a whisky.

MTCam: > we decided that you have to buy us a whisky.
+10 👍 

I also don't feel strongly about this.  If it is safe to remove, and @tulioricci strongly dislikes then it should be inlined.

mirgecom/multiphysics/thermally_coupled_fluid_wall.py
majosm: Done in 6cd4494.

PR #907, opened 2023-05-25
add a check for ISL version

matthiasdiener:
Code based on https://github.com/inducer/islpy/issues/28#issuecomment-702408541

---

**Questions for the review**:

### DIFF
--- a/mirgecom/mpi.py
+++ b/mirgecom/mpi.py
@@ -167,7 +167,7 @@ def _check_isl_version() -> None:
     library to determine if we are running with imath-32.
     """
     import ctypes
-    import islpy
+    import islpy  # type: ignore[import]
 
     try:
         ctypes.cdll.LoadLibrary(islpy._isl.__file__).isl_val_get_num_gmp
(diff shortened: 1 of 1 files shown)


inducer:
It might be worth having this in loopy instead?

MTCam:
> It might be worth having this in loopy instead?

Could someone update kaushikcfd's fork if it does go into loopy?  Would be cool if we could just take that over and make it illinois-ceesd/loopy, then we could keep it updated for ourselves.

matthiasdiener:
> It might be worth having this in loopy instead?

I'm not sure - this code only works since we use the `isl` package from conda-forge, which only provides the `gmp` and `imath32` variants, but not `imath`. I have not found a way to differentiate between `imath` and `imath32` via this method, so in a general installation, we can not be sure if isl is the "very fast" or "very slow" variant...

inducer:
That's a fair point.

MTCam:
Do we want to go ahead and merge this here, then?

MTCam (approved):

PR #908, opened 2023-05-26
Radiation for production

tulioricci:
No merge, only production feeder

**Questions for the review**:


majosm:
> Took a quick pass, looks great to me. I just have a little concern about how we're handling `fluid_bc_class`, not sure it'd be critical at this point.

Thanks! (FYI, the "real" radiation PR is #893. This one is that + ESDG. Strictly speaking I guess that means ESDG should merge before radiation? But I think if we merge #893 into main, then merge this into #922, and finally merge that up with the new main, it should work out in the end.)

PR #914, opened 2023-06-16
expose coupled gradient computation 

anderson2981:
This encapsulates the part that updates the boundaries and computes the gradients for the coupled system. Basically everything that gets done before calling ns_operator and diffusion_operator. This is required when the driver needs the gradients to compute some other mesh properties, e.g. artificial viscosity.

Updates coupled_ns_heat_operator to use the new implementation.

**Questions for the review**:


majosm (commented):
I wonder if we'd be better off just switching over to the driver-defined coupling as implemented in https://github.com/illinois-ceesd/mirgecom/pull/902? I think @tulioricci is already using it.

majosm:
Is there anything in this PR that we still want to add? Or did #902 mostly cover it?

anderson2981:
no longer needed.

PR #917, opened 2023-06-22
Fix issues with `==` in tests and update to work with numpy 1.25

majosm:
We seem to have some bugs in our tests that were uncovered by the changes in object array behavior in numpy 1.25. Specifically, in a few places we assert on the equality of `ConservedVars` objects:

```python
            assert ff_bndry_state.cv == exp_ff_cv
```

which looks reasonable enough, but upon inspection:

```python
ff_bndry_state.cv == exp_ff_cv=ConservedVars(mass=DOFArray((cl.TaggableCLArray([[1],
       [1]], dtype=int8),)), energy=DOFArray((cl.TaggableCLArray([[1],
       [1]], dtype=int8),)), momentum=False, species_mass=array([], dtype=bool))
bool(ff_bndry_state.cv == exp_ff_cv)=True
```

i.e., `==` here is doing some strange things, and CV's boolean conversion is questionable. The discretization here is the boundary of a 1D mesh, so it looks like most of the components are doing elementwise equality, except for momentum.

If I understand correctly, `ConservedVars` gets its `__eq__` operator from arraycontext's `with_container_arithmetic` decorator. @inducer Is the equality operator resulting from this decorator intended to do elementwise comparison? If so, is there a way we can make it work for array containers containing object arrays? Or do we just need to disable `==` for `ConservedVars`, as I've attempted to do here?

**Questions for the review**:

### DIFF
--- a/mirgecom/eos.py
+++ b/mirgecom/eos.py
@@ -65,7 +65,7 @@ class MixtureEOSNeededError(Exception):
 
 
 @dataclass_array_container
-@dataclass(frozen=True)
+@dataclass(frozen=True, eq=False)
 class GasDependentVars:
     """State-dependent quantities for :class:`GasEOS`.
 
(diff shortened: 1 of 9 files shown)


inducer:
[This](https://github.com/inducer/arraycontext/pull/234) is related, and I think the upshot is that `==` for array containers with object arrays is no longer a thing, with `actx.np.equal(..., ...)` the most plausible replacement.

matthiasdiener (approved):

PR #918, opened 2023-06-26
factor out actx init

matthiasdiener:
**Questions for the review**:

### DIFF
--- a/mirgecom/simutil.py
+++ b/mirgecom/simutil.py
@@ -1138,7 +1138,8 @@ def configurate(config_key, config_object=None, default_value=None):
 
 
 def get_reasonable_array_context_class(lazy: bool = False, distributed: bool = True,
-                              profiling: bool = False):
+                              profiling: bool = False) -> Type[ArrayContext]:
+    """Return a :class:`ArrayContext` that satisfies the given constraints."""
     if lazy and profiling:
         raise ValueError("Can't specify both lazy and profiling")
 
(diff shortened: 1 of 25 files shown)


MTCam (commented):
I love this.  Maybe we don't need the `use_profiling` or the interface to query whether it is a profiling actx.   Maybe we should consider sticking actx utils into a separate module like array_context.py so that we import from mirgecom.array_context.

mirgecom/simutil.py
```diff
@@ -1129,6 +1137,70 @@ def configurate(config_key, config_object=None, default_value=None):
...
+    from arraycontext import PyOpenCLArrayContext
+    return issubclass(actx_class, PyOpenCLArrayContext)
+
+
+def actx_class_is_profiling(actx_class) -> bool:
+    from mirgecom.profiling import PyOpenCLProfilingArrayContext
+    return issubclass(actx_class, PyOpenCLProfilingArrayContext)
+
+
+def initialize_actx(actx_class: Type[ArrayContext], comm: Optional["Comm"]) -> Tuple[ArrayContext, cl.Context, cl.CommandQueue, cl.tools.AllocatorBase]:
```
majosm, line 1170: Maybe only return `actx`, so it's compatible with non-CL array contexts (numpy?)? It looks like `context`, `queue`, and `allocator` are all attributes of `actx` for pyopencl/pytato, so presumably they could be retrieved in the driver that way if needed.

majosm (commented):
Just one thought, otherwise I like it. 👍

mirgecom/simutil.py
matthiasdiener: Thanks, I think that makes sense. I changed it in [6578353](https://github.com/illinois-ceesd/mirgecom/pull/918/commits/65783536c02c3a01bb3d184474d11c48b85e3e87)

MTCam (approved):
This is great.

PR #920, opened 2023-06-27
Use tmp cache dirs (by default) on porter, dont override user settings

MTCam:
Slight change to parallel spawning on porter to use tmp cache dirs, trying to stop `CUDA_FILE_NOT_FOUND` and `Failed to generate PTX` errors on Porter.
Also changes all of parallel spawning not to override the users settings if they exist.

**Questions for the review**:


scripts/run-gpus-generic.sh
```diff
@@ -11,6 +11,13 @@
 #
 # Run it like this:
 #   mpiexec -n 2 bash run-gpus-generic.sh python -m mpi4py pulse-mpi.py --lazy
+# unset CUDA_CACHE_DISABLE
+POCL_CACHE_ROOT=${POCL_CACHE_ROOT:-"/tmp/$USER/pocl-scratch"}
+XDG_CACHE_ROOT=${XDG_CACHE_HOME:-"/tmp/$USER/xdg-scratch"}
+POCL_CACHE_DIR=${POCL_CACHE_DIR:-"${POCL_CACHE_ROOT}/rank$OMPI_COMM_WORLD_RANK"}
+XDG_CACHE_HOME=${XDG_CACHE_HOME:-"${XDG_CACHE_ROOT}/rank$OMPI_COMM_WORLD_RANK"}
```
matthiasdiener, line 18: This won't work with mpich, can you put this in the `if` below and use `$MPI_LOCALRANKID` in the mpich case?

matthiasdiener (approved):

PR #924, opened 2023-06-29
Update get_box_mesh.py

aakankshbhat:
Updated get_box_mesh function

**Questions for the review**:

### DIFF
--- a/get_box_mesh.py
+++ b/get_box_mesh.py
@@ -1,0 +1,14 @@
+def get_box_mesh(dim,a,b,n, t=None, periodic=None):
+    if periodic is None:
+        periodic = (False)*dim
+        
+    dim_names = ["x","y","z"]
+    bttf = {}
+    for i in range(dim):
+        bttf["-"+str(i+1)] = ["-"+dim_names[i]]
+        bttf["+"+str(i+1)] = ["+"+dim_names[i]]
+    from meshmode.mesh.generation import generate_regular_rect_mesh as gen
+    if type(a) == tuple:
+        return gen(a=a, b=b, n=n, boundary_tag_to_face=bttf, mesh_type=t,periodic=periodic)
+    else:
+        return gen(a=(a,)*dim, b=(b,)*dim,n=(n,)*dim, boundary_tag_to_face=bttf, mesh_type=t, periodic=periodic)


get_box_mesh.py
```diff
@@ -0,0 +1,14 @@
+def get_box_mesh(dim,a,b,n, t=None, periodic=None):
+    if periodic is None:
+        periodic = (False)*dim
```
majosm, line 3:
```suggestion
        periodic = (False,)*dim
```
(Otherwise I think it doesn't become a tuple.)

get_box_mesh.py
```diff
@@ -0,0 +1,14 @@
...
+
```
majosm, line 4:
```suggestion
    if np.isscalar(a):
        a = (a,)*dim
    if np.isscalar(b):
        b = (b,)*dim
    if np.isscalar(n):
        n = (n,)*dim

```
Promoting the scalars to tuples is probably simpler than having the conditional call to `gen` below.

majosm (commented):
Thanks! This looks like a good start. I added a couple of minor suggestions. The next steps should probably be to: 1) put this into an existing mirgecom module (I think `mirgecom/simutil.py` would be a good spot), and 2) update the examples and tests to import `get_box_mesh` from the module, rather than defining it locally in each one.

mirgecom/simutil.py
```diff
@@ -92,6 +92,33 @@ def get_number_of_tetrahedron_nodes(dim, order, include_faces=False):
         nnodes = nnodes + (dim+1)*get_number_of_tetrahedron_nodes(dim-1, order)
     return nnodes
 
+def get_box_mesh(dim,a,b,n, t=None, periodic=None):
+    """
+    Creates a rectangular "box" like mesh using function generate_regular_rect_mesh
+
+    Ensures parameters needed for generate_regular_rect_mesh are satisfied
+    a,b,n,periodic are tuples of dimension = dim and are converted if not in this format
```
majosm, line 100: If you could format the arguments like [this](https://github.com/illinois-ceesd/mirgecom/blob/27130a33b32f6a08f1af02dcbb5c34f1410b8b9b/mirgecom/simutil.py#L160-L185), then they'll render better in the built documentation ([like this](https://mirgecom.readthedocs.io/en/latest/support/tools.html)).

get_box_mesh.py
```diff
@@ -0,0 +1,20 @@
+def get_box_mesh(dim,a,b,n, t=None, periodic=None):
```
majosm, line 1: Let's remove this file and `get_box_mesh_v2.py`, and just keep the one in `mirgecom/simutil.py`.

majosm (commented):
Looks good! Just added a few suggestions. Once those are addressed and the CI errors are fixed up, it should be ready to merge.

tulioricci:
Hi @aakankshbhat , before you commit any changes, load the mirgecom environment and run "flake8 *.py" and "pydocstyle *.py" so you can catch any errors with flake8 or docs without having to go through CI. That will make your life easier.

MTCam:
> Hi @aakankshbhat , before you commit any changes, load the mirgecom environment and run "flake8 *.py" and "pydocstyle *.py" so you can catch any errors with flake8 or docs without having to go through CI. That will make your life easier. You can also run the examples to make sure there are no issues before committing and pushing.

Actually, if you save this bit of code as `pre-commit` in mirgecom/.git/hooks, then it will basically do what Tulio suggests above, and make sure everything is up-to-snuff at commit time automatically.  It can get annoying every time, but can disable it on a case-by-case basis by specifying the `-n` option when committing if you don't want to run the hooks.

https://gist.github.com/MTCam/c9b7b58bdcd0cc72c1501b537cf3bb66

PR #941, opened 2023-07-28
Add MixtureEOS tests and more mechanisms

tulioricci:
- Mechanisms:
    - Update 7sp UIUC and rename past 7sp mechanism
    - Add mechanism for Y3 prediction
    - Add mechanism for graphite oxidation

- Add MixtureEOS test.
    - If I am blind and these tests are done somewhere else, I will remove `test_eos`.
    - If indeed they are not done anywhere else, any other suggestions for tests?

**Questions for the review**:


mirgecom/mechanisms/__init__.py
```diff
@@ -87,6 +87,11 @@ def get_mechanism_cti(mechanism_name: str) -> str:
 
 def get_mechanism_input(mechanism_name: str) -> str:
     """Get the contents of a mechanism YAML input file."""
+    if mechanism_name == "uiuc":
+        from warnings import warn
+        warn("The uiuc mechanism was updated in Q3 2023. "
+             "The previous mechanism is now called uiuc_xpacc.yaml.",
+             stacklevel=2)
```
MTCam, line 94: Let's delete the old mechanism and the symbolic link, and just reset the mech name to use the right one.  The warning should be enough to spur folks (all 3 or 4 of us) into getting it changed. 

```suggestion
    if mechanism_name == "uiuc":
        from warnings import warn
        warn("The uiuc mechanism was updated in Q3 2023. "
             "The previous mechanism is now called uiuc_xpacc.yaml.",
             stacklevel=2)
        mechanism_name="uiuc_xpacc.yaml"
```

MTCam (approved):
👍  If we can, we should add the new mechanisms to the existing tests.

PR #951, opened 2023-08-11
Update examples runscript and tweak lazy/numpy/eager comparisons

MTCam:
Some changes to bounce off dev team.
cc: @matthiasdiener 

**Questions for the review**:

### DIFF
--- a/examples/run_examples.sh
+++ b/examples/run_examples.sh
@@ -1,12 +1,35 @@
 #!/bin/bash
 
-set -ex
+# This option is undesired here.  We want to make sure to try all the examples,
+# even if some fail.
+# set -e
+# I'd prefer to set this judiciously to have clean-looking output in the CI logs
+# set -x
 set -o nounset
-
-rm -f *.vtu *.pvtu
 
 origin=$(pwd)
 examples_dir=${1-$origin}
+shift
+
+# This bit will let the user specify which
+# examples to run, default to run them all.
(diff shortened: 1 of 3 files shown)


examples/run_examples.sh
```diff
@@ -18,49 +44,200 @@ fi
...
+    #    TOL_NUMPY=1
+    # fi
 
     date
     printf "***\n***\n"
 
-    ${mpi_exec} -n 2 $mpi_launcher python -m mpi4py $example --casename ${example}-eager
-    ${mpi_exec} -n 2 $mpi_launcher python -m mpi4py $example --casename ${example}-lazy --lazy
-    ${mpi_exec} -n 2 $mpi_launcher python -m mpi4py $example --casename ${example}-numpy --numpy
+    declare -A test_results
```
matthiasdiener, line 82: Does this work on MacOS? Associative arrays are a bash 4 feature I think, while macOS has version 3

MTCam: Not out of the box, but it works in CI@github.  It also works on Lassen.  I brew installed bash and you get bash 5, then the script works on my Mac when I change the shbang to /opt/homebrew/bin/bash.

matthiasdiener: removed this requirement in [1cc955a](https://github.com/illinois-ceesd/mirgecom/pull/951/commits/1cc955ad740f6b1a2540da223ddd31bc92428cfa)

matthiasdiener (approved):

PR #950, opened 2023-08-09
Restructure examples

matthiasdiener:
- No more `-lazy` symlinks
- No more `-mpi` in name (the only non-MPI example, `wave.py`, has been renamed to `wave-nompi.py`).
- Test accuracy of all examples, comparing eager/lazy/numpy, as part of `run_examples.sh`

---

**Questions for the review**:
- [ ] Is the scope and purpose of the PR clear?
  - [ ] The PR should have a description.
  - [ ] The PR should have a guide if needed (e.g., an ordering).
- [ ] Is every top-level method and class documented? Are things that should be documented actually so?
- [ ] Is the interface understandable? (I.e. can someone figure out what stuff does?) Is it well-defined?
- [ ] Does the implementation do what the docstring claims?
- [ ] Is everything that is implemented covered by tests?
- [ ] Do you see any immediate risks or performance disadvantages with the design? Example: what do interface normals attach to?

PR #954, opened 2023-08-17
restructure examples v2

matthiasdiener:
Continuation of #950, without the more contentious dependency on #948.

- No more `-lazy` symlinks
- No more `-mpi` in name (the only non-MPI example, `wave.py` has been renamed to `wave-nompi.py`).
- Test accuracy of all examples, comparing eager/lazy/numpy, as part of `run_examples.sh`

**Questions for the review**:


MTCam (approved):
+100

tulioricci:
All examples will be running with lazy? I am afraid that the compilation time with ESDG and NS coupling will make CI take a few extra hours.

matthiasdiener:
> All examples will be running with lazy? I am afraid that the compilation time with ESDG and NS coupling will make CI take a few extra hours.

If it turns out to be an issue, we could manually blacklist those long-running examples inside (e.g.) run_examples.sh.

MTCam:
> All examples will be running with lazy? I am afraid that the compilation time with ESDG and NS coupling will make CI take a few extra hours.

We might need a mechanism to allow black-listing of certain examples from being run in certain ways.  Not sure.   Currently, the CI time isn't terrible.  `pytest` still takes longer than the examples.  Running all the examples in lazy isn't really new, but having no way to control which ones are run in lazy is new.

PR #957, opened 2023-08-19
run_examples.sh: fix running from different dir, add GitHub actions log grouping

matthiasdiener:
This is how the grouping looks like:

image

**Questions for the review**:

### DIFF
--- a/examples/run_examples.sh
+++ b/examples/run_examples.sh
@@ -2,7 +2,7 @@
 
 set -o nounset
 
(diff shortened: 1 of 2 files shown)


MTCam (approved):
👍 This is great.  If you can stand by (#958), that'd be great.  Otherwise we do need to fix the line where it sources the `mirge-testing-env.sh`.  It needs the top level mirgecom directory to be the first argument.

PR #968, opened 2023-09-07
Node-local domain decomp

MTCam:
This change set brings changes from @majosm's fork (https://github.com/majosm/mirgecom/commit/a444cdd7c8618caa79b14e93adf752e0c4f86c3d), which makes mesh partitioning a node-local affair.   Each physical compute node gets the mesh data (read from gmsh or generated) and locally partitions and distributes the mesh partitions among the ranks local to the physical compute node.   

**Questions for the review**:

### DIFF
--- a/mirgecom/mpi.py
+++ b/mirgecom/mpi.py
@@ -36,6 +36,10 @@ import sys
 from contextlib import contextmanager
 from typing import Callable, Any, Generator, TYPE_CHECKING
 
+import logging
+logger = logging.getLogger(__name__)
+
+
 if TYPE_CHECKING:
     from mpi4py.MPI import Comm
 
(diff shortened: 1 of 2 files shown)


mirgecom/simutil.py
```diff
@@ -888,6 +890,94 @@ def generate_and_distribute_mesh(comm, generate_mesh, **kwargs):
...
+
+    part_id_to_elements = {
+        PartID(volumes[vol_idx], rank):
+            np.where(
+                (volume_index_per_element == vol_idx)
+                & (rank_per_element == rank))[0]
+        for vol_idx in range(len(volumes))
+        for rank in range(num_ranks)}
+
+    # FIXME: Find a better way to do this
```
majosm, line 933:
```suggestion
    # TODO: Add a public meshmode function to accomplish this? So we're
    # not depending on meshmode internals
```

majosm (commented):
I took a quick look through the code to see if the node-local code is up to date. Looks like it is, apart from one comment. This version also has the correct indentation for `part_id_to_tag_to_elements`.

Issue #955, opened 2023-08-18
Update (a few) initializers to consider mixtures


tulioricci:
Adding this as a reminder: many initializers only work with `IdealSingleGas` but not mixtures.

@MTCam

Issue #576, opened 2021-12-15
Refactor boundary treatments


MTCam:
The current *MIRGE-Com* fluid boundary condition infrastructure is messy and hard to follow.   We want to refactor it.

The general strategy may be to use inheritance properly and eliminate the need for the derived classes (bc's) to specify a collection of functions to customize the boundary treatment.  Instead, function overrides can be used to clean this up.   At the same time, we should comb through to make sure the operators and their use of the boundaries are properly passing through the numerical flux functions specified at the simulation driver level, and that the boundary things are using it properly.

majosm:
One thing I think I'd like is if every BC explicitly implemented the flux functions that it's compatible with (using helper functions inside as needed to reduce code duplication), rather than just passing callbacks to the base class. This would prevent the possibility of, say, accidentally omitting a boundary callback when calling the base class constructor, and then having it silently apply an incorrect BC (which I have done a couple of times so far with the wall model BCs). Failing to implement a flux function in this explicit approach would instead result in a user-visible exception when the operator attempts to call it.

MTCam:
> One thing I think I'd like is if every BC explicitly implemented the flux functions that it's compatible with (using helper functions inside as needed to reduce code duplication), rather than just passing callbacks to the base class. This would prevent the possibility of, say, accidentally omitting a boundary callback when calling the base class constructor, and then having it silently apply an incorrect BC (which I have done a couple of times so far with the wall model BCs). Failing to implement a flux function in this explicit approach would instead result in a user-visible exception when the operator attempts to call it.

I like this, and agree, although I don't like it when every BC implements giant wads of repeated code. Trying to factor out all the repeated (error-prone) code was the main motivation for the boundary objects' current shape.  The execution wasn't great - but can be made way better now.

PR #970, opened 2023-09-19
Update some boundaries and initializers for mixtures

tulioricci:
- modifies some `eos` functions to not need a full cv, so we can use them for initialization (#955)
- deprecate `MixtureInitializer` and use `Uniform` instead to avoid duplicate codes.
    - updated the existing tests
- rename some arguments of initializers to keep consistent nomenclature along the different functions
- update BCs:
    - to handle mixture
    - use `MengaldoBC` as the inherited class (#576)
    - make `DummyBoundary` inherit `FluidBoundary` instead of `PrescribedBnd`; added test
- examples:
    - replace `poiseuille-local-dt` with `blasius flat plate flow`, which exercise more BCs
    - include `pulse-mixture` to exercise mixture BCs

**Questions for the review**:


examples/blasius.py
```diff
@@ -0,0 +1,554 @@
...
+        pressure = 1.0
+        mass = pressure/(eos.gas_const()*temperature)
+        velocity = self._velocity
+
+        energy = pressure/(eos.gamma() - 1.0) + 0.5*mass*np.dot(velocity, velocity)
+
+        return make_conserved(dim=self._dim,
+                              mass=mass,
+                              energy=energy,
+                              momentum=mass*velocity)
```
MTCam, line 92: Add to initializers?

examples/blasius.py
```diff
@@ -0,0 +1,554 @@
...
+
+#####################################################################
+
+    def my_write_viz(step, t, dt, state):
+        cv = state.cv
+        dv = state.dv
+        viz_fields = [("CV", cv),
+                      ("dt", dt),
+                      ("DV_U", cv.momentum[0]/cv.mass),
+                      ("DV_V", cv.momentum[1]/cv.mass),
```
MTCam, line 344: ooc, why not just "velocity", cv.velocity?

examples/blasius.py
```diff
@@ -0,0 +1,554 @@
...
+        pressure = force_evaluation(actx, dv.pressure)
+        temperature = force_evaluation(actx, dv.temperature)
+
+        if global_reduce(check_naninf_local(dcoll, "vol", pressure), op="lor"):
+            health_error = True
+            logger.info(f"{rank=}: NANs/Infs in pressure data.")
+
+        if global_reduce(check_naninf_local(dcoll, "vol", temperature), op="lor"):
+            health_error = True
+            logger.info(f"{rank=}: NANs/Infs in temperature data.")
```
MTCam, line 379: It's best to perform local checks here, and reduce once outside.

examples/blasius.py
```diff
@@ -0,0 +1,554 @@
...
+                dt = get_sim_timestep(dcoll, fluid_state, t, dt, current_cfl,
+                                      t_final, constant_cfl, local_dt)
+
+        try:
+            do_viz = check_step(step=step, interval=nviz)
+            do_restart = check_step(step=step, interval=nrestart)
+            do_health = check_step(step=step, interval=nhealth)
+
+            if do_health:
+                health_errors = global_reduce(my_health_check(cv, dv), op="lor")
```
MTCam, line 412: global reduction here will catch when any rank fails health.  inside health is the rank-local check meant to sound off on any rank-specific violation.

mirgecom/boundary.py
```diff
@@ -1848,7 +1850,7 @@ def grad_temperature_bc(self, dcoll, dd_bdry, grad_t_minus, normal, **kwargs):
         return self._adiabatic.grad_temperature_bc(grad_t_minus, normal)
 
 
-class LinearizedOutflowBoundary(PrescribedFluidBoundary):
+class LinearizedOutflowBoundary(MengaldoBoundaryCondition):
```
MTCam, line 1853: I don't recall seeing this boundary type in Mengaldo, and I think maybe if we are going to hard-code 2D boundaries, maybe they should have `2D` in the name of the boundary class.

examples/blasius.py
tulioricci: This is not a Blasius solution per-se. It is just a smoothed field to avoid crashing right away due to hot flow/cold wall. Thus, I don't think it should be in the initializers.

mirgecom/boundary.py
tulioricci: Yes, they are not described in there. But I want to apply the BCs as the paper recommends without duplicating the existing code.

mirgecom/initializers.py
```diff
@@ -634,13 +635,11 @@ class MulticomponentLump:
...
-            rho0=1.0, p0=1.0,
-            center=None, velocity=None,
-            spec_y0s=None, spec_amplitudes=None,
-            spec_centers=None, sigma=1.0
-    ):
+    def __init__(self, *, dim=1, nspecies=0,
+                 rho0=1.0, p0=1.0,
+                 center=None, velocity=None,
+                 spec_y0s=None, spec_amplitudes=None,
+                 spec_centers=None, sigma=1.0):
```
majosm, line 642: (Personal preference, I know, but ugh -- I hate the parentheses-aligned arguments thing. Shifts everything over so far.)

mirgecom/initializers.py
```diff
@@ -1001,36 +990,46 @@ def __call__(self, x_vec, cv, eos=None, **kwargs):
         if x_vec.shape != (self._dim,):
             raise ValueError(f"Expected {self._dim}-dimensional inputs.")
 
-        gamma = eos.gamma()
+        if isinstance(eos, IdealSingleGas) is False:
```
majosm, line 993:
```suggestion
        if not isinstance(eos, IdealSingleGas):
```

majosm (commented):
Added some minor comments.

mirgecom/boundary.py
```diff
@@ -1020,12 +1008,44 @@ def viscous_divergence_flux(self, dcoll, dd_bdry, gas_model, state_minus,
...
+    .. automethod:: viscous_divergence_flux
+    .. automethod:: cv_gradient_flux
+    .. automethod:: temperature_gradient_flux
+    """
 
     def __init__(self):
-        """Initialize the DummyBoundary boundary type."""
-        PrescribedFluidBoundary.__init__(self)
+        from warnings import warn
+        warn("Using dummy boundary: copies interior solution.", stacklevel=2)
```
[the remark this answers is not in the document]
tulioricci: If we are talking about an experienced user then yes, but the problem is any potential new user. I think it doesn't hurt to keep it.

mirgecom/boundary.py
```diff
@@ -1188,26 +1203,38 @@ def farfield_state(self, dcoll, dd_bdry, gas_model, state_minus, **kwargs):
...
         return make_fluid_state(cv=cv_infinity, gas_model=gas_model,
                                 temperature_seed=free_stream_temperature,
                                 smoothness_mu=state_minus.smoothness_mu,
                                 smoothness_kappa=state_minus.smoothness_kappa,
                                 smoothness_beta=state_minus.smoothness_beta)
 
+    def state_bc(self, dcoll, dd_bdry, gas_model, state_minus, **kwargs):
+        """Return BC fluid state."""
+        return self.state_plus(dcoll, dd_bdry, gas_model, state_minus, **kwargs)
+
```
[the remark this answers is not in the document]
tulioricci: Yeah, I've being thinking about this. Since this is a far-field BC, then the values should converge. But I don't see why not use an average. What are your thoughts?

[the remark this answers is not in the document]
majosm: (As discussed in Slack, this appears to be what Mengaldo 2014 prescribes for this case.)

examples/pulse.py
```diff
@@ -107,7 +104,7 @@ def main(actx_class, use_esdg=False,
         timestepper = RK4MethodBuilder("state")
     else:
         timestepper = rk4_step
-    t_final = 0.1
+    t_final = 1.0
```
MTCam, line 107: yikes :)

examples/pulse.py
```diff
@@ -140,22 +137,16 @@ def main(actx_class, use_esdg=False,
         generate_mesh = partial(generate_regular_rect_mesh,
             a=(box_ll,)*dim, b=(box_ur,)*dim,
             nelements_per_axis=(nel_1d,)*dim,
-            boundary_tag_to_face={
-                "outlet_L": ["+y"],
-                "outlet_R": ["-y", "+x"],
-                "inlet": ["-x"]})
+            periodic=(True,)*dim)
```
MTCam, line 140: Can we make it have AdiabaticSlipWalls instead?

mirgecom/boundary.py
```diff
@@ -1141,27 +1158,25 @@ class FarfieldBoundary(PrescribedFluidBoundary):
...
 
-    def __init__(self, numdim, free_stream_pressure,
-                 free_stream_velocity, free_stream_temperature,
-                 free_stream_mass_fractions=None):
+    def __init__(self, dim, free_stream_pressure, free_stream_velocity,
+                 free_stream_temperature, free_stream_mass_fractions=None):
         """Initialize the boundary condition object."""
-        if len(free_stream_velocity) != numdim:
+        if len(free_stream_velocity) != dim:
             raise ValueError("Free-stream velocity must be of ambient dimension.")
```
MTCam, line 1172:
```suggestion
        dim = len(free_stream_velocity)
```

If seems a little odd to force the user to specify `dim` for a check that happens inside the BC object.

tulioricci: https://github.com/illinois-ceesd/mirgecom/pull/970/commits/7cc91555139d7b27aaba6d7fee2d14dce230a3af

examples/pulse.py
tulioricci: https://github.com/illinois-ceesd/mirgecom/pull/970/commits/7cc91555139d7b27aaba6d7fee2d14dce230a3af

examples/pulse.py
MTCam:
```suggestion
    t_final = 0.1
```

mirgecom/boundary.py
```diff
@@ -16,14 +16,15 @@
...
 .. autoclass:: RiemannInflowBoundary
 .. autoclass:: RiemannOutflowBoundary
-.. autoclass:: PressureOutflowBoundary
 .. autoclass:: IsothermalSlipWallBoundary
 .. autoclass:: IsothermalWallBoundary
 .. autoclass:: AdiabaticSlipBoundary
 .. autoclass:: AdiabaticNoslipWallBoundary
-.. autoclass:: LinearizedOutflowBoundary
+.. autoclass:: LinearizedOutflow2DBoundary
+.. autoclass:: LinearizedInflow2DBoundary
```
MTCam, line 27: What are the chances I could get you to name this *2D, instead of embedding the 2D down inside the name?

mirgecom/eos.py
```diff
@@ -137,8 +137,9 @@ def sound_speed(self, cv: ConservedVars, temperature: DOFArray):
         """Get the gas sound speed."""
 
     @abstractmethod
-    def gas_const(self, cv: Optional[ConservedVars],
-                  temperature: Optional[DOFArray] = None):
+    def gas_const(self, cv: Optional[ConservedVars] = None,
+                  temperature: Optional[DOFArray] = None,
+                  species_mass_fractions: Optional[np.ndarray] = None) -> DOFArray:
```
MTCam, line 142: I don't understand adding the`species_mass_fractions` arg to single gas EOS.

mirgecom/eos.py
```diff
@@ -699,7 +707,7 @@ def gas_const(self, cv: ConservedVars,  # type: ignore[override]
             ($\rho$), energy ($\rho{E}$), momentum ($\rho\vec{V}$), and the vector
             of species masses, ($\rho{Y}_\alpha$).
         """
-        y = cv.species_mass_fractions
+        y = species_mass_fractions if cv is None else cv.species_mass_fractions
```
MTCam, line 710: Kind of yuck.   What about just making species_mass_fractions a required argument and passing cv.species_mass_fractions instead of CV when possible?

mirgecom/initializers.py
```diff
@@ -54,60 +54,62 @@
...
+        species_mass_fractions=None, return_fluid_state=False):
     """Create a fluid CV/state from a set of minimal input data."""
-    if gas_model is None:
-        raise ValueError("Gas model is required to create a FluidState.")
+    if gas_model is None and eos is None:
+        raise ValueError("Must provide 1 of (gas_model, eos).")
+    if eos is None:
+        eos = gas_model.eos
+    if return_fluid_state and gas_model is None:
+        raise ValueError("Must provide gas_model to create a fluid state.")
```
MTCam, line 69: What if gas_model and eos are both None?  EOS should always come from gas_model if it exists, can we do something like:
```suggestion
    if gas_model is not None:
       	if eos is not None:
       	    warn("Specifying both EOS and gas_model is overspecified")
       	 eos = gas_model.eos
    elif return_fluid_state:
        raise ValueError("Must provide gas_model to create a fluid state.")
```

mirgecom/initializers.py
```diff
@@ -54,60 +54,62 @@
 from pytools.obj_array import make_obj_array
 from mirgecom.eos import IdealSingleGas
 from mirgecom.fluid import make_conserved
+from mirgecom.gas_model import make_fluid_state
 
 
-def initialize_flow_solution(actx, dcoll, gas_model, dd_bdry=None, pressure=None,
-                             temperature=None, density=None, velocity=None,
-                             mass_fractions=None, make_fluid_state=False):
+def initialize_flow_solution(actx, nodes, gas_model=None, eos=None,
```
MTCam, line 60: I don't think I like `nodes` as an argument name.  I know the original used "nodes" but that was bad, too.   Can we just call it `coords` or some such?

mirgecom/initializers.py
```diff
@@ -54,60 +54,62 @@
...
-    if make_fluid_state:
-        from mirgecom.gas_model import make_fluid_state
+    if return_fluid_state:
         return make_fluid_state(cv=cv, gas_model=gas_model,
-            temperature_seed=temperature)
-    else:
-        return cv
+                                temperature_seed=temperature)
+
+    return cv
```
MTCam, line 112: Is there a good reason to make this do the `make_fluid_state` part?   I wonder whether this function should always just return CV and let the user make the state if they want?  Then maybe we don't need to do the song-and-dance with the EOS vs. gas_model.

mirgecom/initializers.py
```diff
@@ -1001,36 +976,46 @@ def __call__(self, x_vec, cv, eos=None, **kwargs):
         if x_vec.shape != (self._dim,):
             raise ValueError(f"Expected {self._dim}-dimensional inputs.")
 
-        gamma = eos.gamma()
+        if not isinstance(eos, IdealSingleGas):
+            temperature = eos.temperature(cv, tseed)
+            gamma = eos.gamma(cv=cv, temperature=temperature)
+        else:
+            temperature = eos.temperature(cv)
+            gamma = eos.gamma(cv=cv)
```
MTCam, line 984: I'm pretty sure this can just be:
```suggestion
        temperature = eos.temperature(cv, tseed)
        gamma = eos.gamma(cv, temperature)
```

mirgecom/eos.py
tulioricci: I added a warn to deprecate this... Since it involve BC and initializer, let's give an extra time before changing this.

mirgecom/initializers.py
tulioricci: Renamed to `coords`.

mirgecom/initializers.py
tulioricci: Since all functions in `initializers` return cv, I agree. Removed the fluid_state option.

mirgecom/initializers.py
tulioricci: Done.

mirgecom/initializers.py
tulioricci: Following comment below, this is history.

mirgecom/boundary.py
tulioricci: LinearizedInflow2DBoundary -> LinearizedInflowBoundary2D ? If yes, done.

PR #974, opened 2023-10-10
add force_compile function

matthiasdiener:
**Questions for the review**:

### DIFF
--- a/examples/ablation-workshop.py
+++ b/examples/ablation-workshop.py
@@ -1103,7 +1103,8 @@ def main(actx_class=None, use_logmgr=True, casename=None, restart_file=None):
 
             if do_garbage:
                 with gc_timer:
-                    warn("Running gc.collect() to work around memory growth issue ")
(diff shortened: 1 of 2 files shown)


majosm:
Is this mainly for measuring the compile time? If so, I wonder if we shouldn't `force_evaluate` the arguments on the inside, since that could throw off the times. Maybe it should forbid passing unevaluated inputs?

matthiasdiener:
> Is this mainly for measuring the compile time? 

Yes, another use case is debugging determinism/caching issues.

> If so, I wonder if we shouldn't `force_evaluate` the arguments on the inside, since that could throw off the times. Maybe it should forbid passing unevaluated inputs?

Hmm, not sure. For profiling compile time, I think the measurement should be done inside the function anyways (so you can choose whether to profile `.compile()`, or the function call, or both). I think moving the force_evaluation outside wouldn't really help much.

MTCam (approved):

PR #976, opened 2023-10-19
New carbon fiber model + orthotropic heat cond.

tulioricci:
Add orthotropic conduction for diffusion equation. The current implementation may not be able to handle fully anisotropic materials. The rotation of the axis can only be done in multiples of 90 degs.

Tests were added to both `test_viscous.py` and `test_diffusion.py`. An example was also added (`heat-diffusion.py`).

Updates the carbon fiber to use FiberForm values. The previous values were char-only data from TACOT and is not exactly representative of the fiber-only.

**Questions for the review**:


mirgecom/diffusion.py
```diff
@@ -169,7 +184,13 @@ def diffusion_facial_flux_harmonic(
     flux_without_penalty = np.dot(flux_tpair.avg, normal)
 
     # TODO: Verify that this is the correct form for the penalty term
-    tau = penalty_amount*kappa_harmonic_mean/lengthscales_tpair.avg
+    if isinstance(kappa_harmonic_mean, np.ndarray):
+        # if anisotropic, get the normal diffusivity absolute value
+        actx = normal[0].array_context
+        kappa_mean_normal = actx.np.abs(np.dot(kappa_harmonic_mean, normal))
```
majosm, line 190: When computing the harmonic mean of vector/tensor-valued `kappa`s, should we be taking the harmonic mean first and then dotting it with the normal, or dotting each with the normal and then taking the harmonic mean? (Or are the two equivalent? TBD)

majosm (commented):
Added a few thoughts.

mirgecom/materials/carbon_fiber.py
```diff
@@ -134,65 +143,68 @@ def void_fraction(self, tau: DOFArray) -> DOFArray:
...
-    def enthalpy(self, temperature: DOFArray, tau: Optional[DOFArray]) -> DOFArray:
+    def enthalpy(self, temperature, tau=None) -> DOFArray:
         r"""Evaluate the solid enthalpy $h_s$ of the fibers."""
         return (
-            - 1.279887694729e-11*temperature**5 + 1.491175465285e-07*temperature**4
-            - 6.994595296860e-04*temperature**3 + 1.691564018108e+00*temperature**2
-            - 3.441837408320e+01*temperature - 1.235438104496e+05)
+            - 3.37112113e-11*temperature**5 + 3.13156695e-07*temperature**4
+            - 1.17026962e-03*temperature**3 + 2.29194901e+00*temperature**2
+            - 3.62422269e+02*temperature**1 - 5.96993843e+04)
```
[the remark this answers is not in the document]
tulioricci: For the previous implementation, I was using numbers for a different kind of fiber where I made an assumption to get the numbers.. In this, Panerai gave me the actual numbers, so we should use this instead.

mirgecom/materials/carbon_fiber.py
```diff
@@ -107,7 +107,11 @@ def get_source_terms(self, temperature, tau, rhoY_o2) -> DOFArray:  # noqa N803
 
 
 class FiberEOS(PorousWallEOS):
-    """Evaluate the properties of the solid state containing only fibers.
+    r"""Evaluate the properties of the solid state containing only fibers.
+
+    .. math::
+        \tau = \frac{m}{m_0} = \frac{\rho_i \epsilon}{\rho_i \epsilon_0}
+             = \frac{r^2}{r_0^2}
```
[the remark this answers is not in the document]
tulioricci: It is all explained in more details in `wall_model.py`. Nonetheless, I added a decent comment now in here.

mirgecom/materials/carbon_fiber.py
```diff
@@ -120,10 +124,15 @@ class FiberEOS(PorousWallEOS):
...
-    def __init__(self, char_mass, virgin_mass):
+    def __init__(self, dim, normal, char_mass, virgin_mass):
         """Bulk density considering the porosity and intrinsic density."""
         self._char_mass = char_mass
         self._virgin_mass = virgin_mass
+        self._dim = dim
+        self._normal = normal
+
+        if normal > dim:
+            raise ValueError("Normal direction must be less or equal than dim.")
```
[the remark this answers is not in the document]
tulioricci: Sorry for the confusion. Normal here is not a face-normal vector. This is the anisotropic direction, which is normal to the isotropic plane. I renamed it to `anisotropic_direction` to be more clear.

mirgecom/diffusion.py
tulioricci: As we discussed offline, the normal dotting is the last step. I worked on the docs (https://github.com/illinois-ceesd/mirgecom/pull/976/commits/29d3bc2c873163e9294ffe2a9e319477931b4074) to highlight the steps we are taking to compute this case. It seems that the general anisotropic case would need some thinking, but we wont have those cases, only orthotropic materials in which the thermal conductivity is a diagonal matrix, which is easily invertible.

mirgecom/diffusion.py
```diff
@@ -74,21 +118,40 @@ def grad_facial_flux_central(kappa_tpair, u_tpair, normal):
...
+
+        F = -\frac{\kappa^- u^- + \kappa^+ u^+}{\kappa^- + \kappa^+} \hat{n},
+
+    or an orthotropic material, as
 
     .. math::
 
-        F = -\frac{\kappa^- u^- + \kappa^+ u^+}{\kappa^- + \kappa^+} \hat{n}.
+        F = -\frac{(\kappa^- \cdot \hat{n}) u^- + (\kappa^+ \cdot \hat{n}) u^+}
+            {\kappa^- \cdot \hat{n} + \kappa^+ \cdot \hat{n}} \hat{n}
```
majosm, line 133: I don't know if we have solid justification that it should look this way (though I also don't have any better suggestions). It would be good to check this with a verification case.

majosm (commented):
Looks good to me apart from minor fixes suggested below and pending verification.

mirgecom/diffusion.py
tulioricci: Matt, I tested having "multiple temperatures" (https://github.com/illinois-ceesd/mirgecom/blob/5a0ea41b6924acb7ba9dfb9b9c9957fc2e7cf48c/mirgecom/diffusion.py#L149-L150) or the product with the normal for a simple diffusion case (see `heat-diffusion.py`).

In practice, the results are the same (see below). I added a comment in the code to keep this in mind so we don't forget (https://github.com/illinois-ceesd/mirgecom/pull/976/commits/3f1dda6c9f90e5c6fa467698f95e24c7fe0addb6).

dx, dy, error, p

multiple temperature
0.03278688524590164 0.03225806451612903 0.027683064208580785 2
0.01639344262295082 0.016129032258064516 0.0030159150665184957 2

single temperature (dot normal)
0.03278688524590164 0.03225806451612903 0.02768306420858256 2
0.01639344262295082 0.016129032258064516 0.003015915066520936 2

test/test_diffusion.py
```diff
@@ -624,6 +629,115 @@ def get_u2(x, t):
...
+            exterior=u_minus)
+        grad_u_tpair = TracePair(dd_bdry,
+            interior=grad_u_minus,
+            exterior=grad_u_minus)
+        lengthscales_tpair = TracePair(
+            dd_bdry, interior=lengthscales_minus, exterior=lengthscales_minus)
+        normal = actx.thaw(dcoll.normal(dd_bdry))
+        return numerical_flux_func(
+            kappa_tpair, u_tpair, grad_u_tpair, lengthscales_tpair, normal,
+            penalty_amount=penalty_amount)
```
majosm, line 665: Maybe worth adding to `diffusion.py`?

test/test_diffusion.py
```diff
@@ -624,6 +629,115 @@ def get_u2(x, t):
...
+    flux_y = -(kappa1 + zeros)*grad_u[1]
+    err_flux_x = actx.to_numpy(op.norm(dcoll, diff_flux[0] - flux_x, np.inf))
+    err_flux_y = actx.to_numpy(op.norm(dcoll, diff_flux[1] - flux_y, np.inf))
+    assert err_flux_x < 1.e-9
+    assert err_flux_y < 1.e-9
+
+    err_rhs = actx.to_numpy(op.norm(dcoll, rhs, np.inf))
+    assert err_rhs < 1.e-8
+
+
```
majosm, line 740: What is this case testing vs. the other one with the `NeumannDiffusionBoundary`s?

majosm (approved):
A few minor comments, otherwise LGTM. 👍

test/test_diffusion.py
tulioricci: I just wanted to make sure it was working with both BCs (Dirichlet-like and Neumann)

test/test_diffusion.py
tulioricci: I don't oppose. Added.

examples/heat-diffusion.py
```diff
@@ -0,0 +1,213 @@
...
+
+        u = rk4_step(u, t, dt, compiled_rhs)
+        u = force_evaluation(actx, u)
+
+        t += dt
+        istep += 1
+
+        if logmgr:
+            set_dt(logmgr, dt)
+            logmgr.tick_after()
```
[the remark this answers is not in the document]
tulioricci: @MTCam prefers everything using the advancer. So I also changed the `heat-source` example.

test/test_diffusion.py
majosm: How about something like
```diff
     err_rhs = actx.to_numpy(op.norm(dcoll, rhs, np.inf))
     assert err_rhs < 1.e-8
 
+    def make_dirichlet_bc(btag):
+        bdry_u = op.project(dcoll, "vol", BoundaryDomainTag(btag), u)
+        return DirichletDiffusionBoundary(bdry_u)
+
     boundaries = {
-        BoundaryDomainTag("-1"): DummyDiffusionBoundary(),
-        BoundaryDomainTag("+1"): DummyDiffusionBoundary(),
-        BoundaryDomainTag("-2"): DummyDiffusionBoundary(),
-        BoundaryDomainTag("+2"): DummyDiffusionBoundary(),
+        BoundaryDomainTag("-1"): make_dirichlet_bc("-1"),
+        BoundaryDomainTag("+1"): make_dirichlet_bc("+1"),
+        BoundaryDomainTag("-2"): make_dirichlet_bc("-2"),
+        BoundaryDomainTag("+2"): make_dirichlet_bc("+2"),
     }
```
instead? I don't know if it's well posed with all dummy BCs.

tulioricci: Done

PR #993, opened 2024-01-11
Add Robin BC to diffusion operator

tulioricci:
Adds Robin BC for the diffusion operator. Tests were conducted using the added example `adv-diff-react.py` to verify convergence of the expected value at the boundary, summarized at the table below. I don't know if it would be good to add a similar analysis in one of the `test` due to extra computational time in CI for running multiple cases to assess convergence. Nonetheless, a test was added to exercise all diffusion BCs in one single case.

|N|order|error|EOC|
|-|-----|-----|---|
|50|2|0.0008709114852747195|xxx|
|100|2|0.00011054282974628119|2.977920649395854|
|200|2|1.3870643546230384e-05|2.9862097342798877|
|50|3|0.00011080654857500696|xxx|
|100|3|1.3872582248482454e-05|2.9977348843336182|
|200|3|1.735537998486721e-06|2.998258191684584|
|50|4|4.6491903854575965e-07|xxx|
|100|4|1.473154291711964e-08|4.979999060829744|
|200|4|4.619611653505728e-10|4.987496106746617|

- The implementation in diffusion is not limited to it, but rather an example on how to implement it. The same logic can be (straightforwardly) employed for the CNS equations, necessary for the McKenna burner simulations (https://github.com/illinois-ceesd/drivers_mckenna/blob/7824820813d152d5e82b21c9f3e13cee9c4864e7/McKenna_heatCoup/burner_heatFluxMeas.v2.py#L1068C1-L1076C80)

**Questions for the review**:


mirgecom/diffusion.py
```diff
@@ -430,6 +431,90 @@ def get_diffusion_flux(
...
+        Initialize the boundary condition.
+
+        Parameters
+        ----------
+        value: float or meshmode.dof_array.DOFArray
+            the reference value(s) of $u$ along the boundary
+        alpha: float or meshmode.dof_array.DOFArray
+            the weight for the variable $u$ at the boundary
+        """
+        self.value = value
```
majosm, line 480: I would maybe call this `u_ref`, or something else. `value` is kind of ambiguous for this BC.

mirgecom/diffusion.py
```diff
@@ -430,6 +431,90 @@ def get_diffusion_flux(
...
+        kappa_tpair = TracePair(dd_bdry,
+            interior=kappa_minus,
+            exterior=kappa_minus)
+        u_tpair = TracePair(dd_bdry,
+            interior=u_minus,
+            exterior=u_minus)
+        normal = actx.thaw(dcoll.normal(dd_bdry))
+        grad_u_tpair = TracePair(dd_bdry,
+            interior=grad_u_minus,
+            exterior=self.alpha*(self.value - u_minus)/kappa_minus * normal)
```
majosm, line 510:
```suggestion
        dudn_bc = self.alpha * (self.value - u_minus)/kappa_minus
        grad_u_tpair = TracePair(dd_bdry,
            interior=grad_u_minus,
            exterior=grad_u_minus + 2 * (dudn_bc - np.dot(grad_u_minus, normal)) * normal)
```

I think...

test/test_diffusion.py
```diff
@@ -43,14 +43,57 @@
...
+
+    err_grad_x = actx.to_numpy(op.norm(dcoll, grad_u[0] - 0.5, np.inf))
+    err_grad_y = actx.to_numpy(op.norm(dcoll, grad_u[1] - 0.0, np.inf))
+    assert err_grad_x < 1.e-9
+    assert err_grad_y < 1.e-9
+
+    err_rhs = actx.to_numpy(op.norm(dcoll, rhs, np.inf))
+    assert err_rhs < 1.e-8
+
+
```
majosm, line 96: Any chance you could add a case to `test_diffusion_accuracy` with a Robin BC? It would be good to check that it converges with the right order.

majosm (commented):
Looking pretty good! I added some thoughts.

examples/advection_diffusion_reaction.py
```diff
@@ -0,0 +1,347 @@
...
+logger = logging.getLogger(__name__)
+
+
+class MyRuntimeError(RuntimeError):
+    """Simple exception to kill the simulation."""
+
+    pass
+
+
+def utx(actx, nodes, t=0):
```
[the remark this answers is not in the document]
tulioricci: I copied-and-pasted from another example.. Renamed to "init_u_field"

examples/advection_diffusion_reaction.py
```diff
@@ -0,0 +1,347 @@
...
+
+    # FIXME use weak_local_d_dx instead if 1D
+    # FIXME use weal_local_div instead if dim >= 2
+    vol_flux = op.weak_local_grad(dcoll, u)[0]
+    return op.inverse_mass(dcoll, vol_flux - op.face_mass(dcoll, el_bnd_flux))
+
+
+def get_source(actx, u, x):
+    return -10.0*u
+    # return 5.0*actx.np.exp(-(x+0.9)**2/0.1**2)
```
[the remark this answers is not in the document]
tulioricci: Done.

examples/advection_diffusion_reaction.py
```diff
@@ -0,0 +1,347 @@
...
+    actx = initialize_actx(actx_class, comm)
+    queue = getattr(actx, "queue", None)
+
+    t_final = 0.1
+    nviz = 100
+
+    dim = 1
+    order = 2
+    nel_1d = 100
+
```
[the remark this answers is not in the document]
tulioricci: Done.

examples/advection_diffusion_reaction.py
```diff
@@ -0,0 +1,347 @@
...
+    # ~~~
+    visualizer = make_visualizer(dcoll)
+
+#    init_message = make_init_message(
+#        dim=dim, order=order,
+#        nelements=local_nelements, global_nelements=global_nelements,
+#        dt=current_dt, t_final=t_final, nstatus=nstatus, nviz=nviz,
+#        t_initial=current_t, cfl=current_cfl, constant_cfl=constant_cfl,
+#        initname=initname, eosname=eosname, casename=casename)
```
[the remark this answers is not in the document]
tulioricci: Fixed.

examples/advection_diffusion_reaction.py
```diff
@@ -0,0 +1,347 @@
...
+    def my_write_viz(step, t, u):
+        viz_fields = [("u", u), ("x", nodes[0])]
+        write_visfile(dcoll, viz_fields, visualizer, vizname=casename,
+                      step=step, t=t, overwrite=True, comm=comm)
+
+    def my_health_check(u):
+        health_error = False
+        if check_naninf_local(dcoll, "vol", u):
+            health_error = True
+            logger.info(f"{rank=}: Invalid pressure data found.")
```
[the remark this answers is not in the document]
tulioricci: Thanks!

examples/advection_diffusion_reaction.py
```diff
@@ -0,0 +1,347 @@
...
+        if logmgr:
+            set_dt(logmgr, dt)
+            logmgr.tick_after()
+        return state, dt
+
+    def my_rhs(t, u):
+
+        # inviscid flux
+        u_bc = op.project(dcoll, "vol", BTAG_ALL, u)
+        inv = linear_advection_operator(dcoll, u=u, u_bc=u_bc)
```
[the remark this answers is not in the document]
tulioricci: Nope, only adding to the viscous part since only one BC is necessary.

mirgecom/diffusion.py
tulioricci: Changed to `u_ref` as suggested.

mirgecom/diffusion.py
tulioricci: Implemented the suggested expression but in a different form. See https://github.com/illinois-ceesd/mirgecom/pull/993/commits/7083161b637591e9f9163b13b370381cdbd60967

test/test_diffusion.py
tulioricci: I added a test, but it is interesting/confusing that the error in test converges with "order" while the example I ran (see PR description) converges with order (for odd) and order + 1 (for even).

Thoughts/ideas?

majosm: My guess is it just comes down to computing the error on an expression that includes derivatives of the solution rather than just the solution itself. I set up another test here: https://github.com/majosm/mirgecom/commit/4a99f4e8562c6df78e21010f666cffe977b6fb4b. It checks the solution on the volume, and it gets the expected convergence rate. When I modify it to compute the error of `grad(u)`, it tends to be 1 order lower. I did notice that if I don't run for long enough, it can appear to converge at a higher order; maybe this is what you observed in your example case?

mirgecom/diffusion.py
majosm: Does that version work with orthotropic kappa, since I think in that case `dot(kappa*grad_u, normal) != dot(kappa*grad_u_normal, normal)` (in other words, kappa changes the vector direction instead of just scaling it)?

examples/heat-source.py
```diff
@@ -159,13 +163,28 @@ def my_write_viz(step, t, state):
...
-                logmgr.tick_before()
 
             if istep % 10 == 0:
-                print(istep, t, actx.to_numpy(actx.np.linalg.norm(u[0])))
+                print(istep, t, actx.to_numpy(actx.np.linalg.norm(state[0])))
+
+            health_errors = global_reduce(my_health_check(state), op="lor")
+            if health_errors:
+                if rank == 0:
+                    logger.info("Fluid solution failed health check.")
```
majosm, line 186:
```suggestion
                    logger.info("Solution failed health check.")
```

test/test_diffusion.py
```diff
@@ -319,43 +424,72 @@ def test_diffusion_accuracy(actx_factory, problem, nsteps, dt, scales, order,
...
             else:
                 quadrature_tag = DISCR_TAG_BASE
+            boundaries = p.get_boundaries(dcoll, actx, t, value=value, alpha=alpha)
             return (
-                diffusion_operator(
-                    dcoll, kappa=kappa, boundaries=p.get_boundaries(dcoll, actx, t),
-                    u=u, quadrature_tag=quadrature_tag)
-                + evaluate(sym_f, x=nodes, t=t))
+                diffusion_operator(dcoll, kappa=kappa, boundaries=boundaries,
+                                   u=u, quadrature_tag=quadrature_tag)
```
majosm, line 444: Triggering me here, lol :sweat: (not a big fan of the parentheses-aligned arguments style).

test/test_diffusion.py
```diff
@@ -319,43 +424,72 @@ def test_diffusion_accuracy(actx_factory, problem, nsteps, dt, scales, order,
...
+                + evaluate(sym_f, x=nodes, t=t)
+                + evaluate(sym_source, x=nodes, t=t, u=u)
```
majosm, line 446: FYI, this test is set up to support manufactured solutions in addition to exact solutions; if the solution specified by the problem is not an exact solution, the RHS source term will be computed symbolically (that's the `sym_f` above). So the `sym_source`/`get_source_term` stuff shouldn't be needed.

mirgecom/diffusion.py
tulioricci: Agreed that it is different. However, we only have orthotropic/anisotropic diffusivities for heat, not for species (which is the main point of this BC in our cases).

I will add a comment wrt anisotropy, but I wouldn't put effort now in a case we will likely never need.

majosm (commented):
Gave it another quick pass.

mirgecom/diffusion.py
majosm: Tempted to say the original suggested version would work though, since it preserves the perpendicular components of `grad_u` (instead of zeroing them out), and just reflects the normal component... agree/disagree?

test/test_diffusion.py
tulioricci: I added the `source`  as an actual source term from the physics of the problem. But with your test suggestion, that is not needed. Removed/aborted adding it everywhere.

mirgecom/diffusion.py
tulioricci: I see my mistake now. Agreed and kept the original suggestion in https://github.com/illinois-ceesd/mirgecom/pull/993/commits/33bf186853a614a2a3703da8fe7a64034ad0f8b2

majosm (approved):
LGTM 👍

PR #1002, opened 2024-02-21
Update Riemann and Linear BCs 

tulioricci:
- Update `species_mass_fractions` for non-mixture cases in RiemannBC.
- Allow either to prescribe or to extrapolate species in LinearOutflowBC; also, do not enforce the gradients to be zeros (over-specified).
- Extended LinearInflowBC and LinearOutflowBC to 1D

**Questions for the review**:

### DIFF
--- a/examples/blasius.py
+++ b/examples/blasius.py
@@ -246,10 +246,12 @@ def main(actx_class, use_overintegration, casename, rst_filename, use_esdg):
     velocity[0] = 0.3*np.sqrt(1.4*1.0*2.0)
     flow_init = Initializer(dim=2, velocity=velocity)
 
-    linear_outflow_bnd = LinearizedOutflow2DBoundary(free_stream_density=0.5,
+    linear_outflow_bnd = LinearizedOutflow2DBoundary(
+        dim=dim, free_stream_density=0.5,
         free_stream_pressure=1.0, free_stream_velocity=velocity)
 
-    linear_inflow_bnd = LinearizedInflow2DBoundary(free_stream_density=0.5,
+    linear_inflow_bnd = LinearizedInflow2DBoundary(
+        dim=dim, free_stream_density=0.5,
         free_stream_pressure=1.0, free_stream_velocity=velocity)
 
     boundaries = {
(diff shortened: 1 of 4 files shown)


examples/blasius.py
```diff
@@ -248,10 +248,12 @@ def get_mesh_data():
     velocity[0] = 0.3*np.sqrt(1.4*1.0*2.0)
     flow_init = Initializer(dim=2, velocity=velocity)
 
-    linear_outflow_bnd = LinearizedOutflow2DBoundary(free_stream_density=0.5,
+    linear_outflow_bnd = LinearizedOutflow2DBoundary(
+        dim=dim, free_stream_density=0.5,
```
MTCam, line 252: 🤔 2D?

mirgecom/boundary.py
```diff
@@ -1909,59 +1901,73 @@ def __init__(self, free_stream_state=None,
...
             self._spec_mass_fracs = free_stream_state.cv.species_mass_fractions
 
-        if self._ref_velocity.shape[0] != 2:
+        if dim == 3:
+            raise ValueError("This BC only supports 1 or 2-dimensional inputs.")
+
+        if self._ref_velocity.shape[0] != dim:
             vel_shape = self._ref_velocity.shape[0]
-            raise ValueError(f"Expected 2-dimensional inputs, got {vel_shape}.")
+            raise ValueError(f"Expected {dim}-dimensional inputs, got {vel_shape}.")
```
MTCam, line 1909: There may be something subtle I am missing here, but why not:

```
dim = self._ref_velocity.shape[0]
if dim == 3:
   yadda yadda
```

tulioricci: https://github.com/illinois-ceesd/mirgecom/pull/1002/commits/b5961dd4b1ce9ae0032071ea48949d3d06d3d363

examples/blasius.py
tulioricci: https://github.com/illinois-ceesd/mirgecom/pull/1002/commits/b5961dd4b1ce9ae0032071ea48949d3d06d3d363

MTCam (approved):
I trust your changes to the physics serve your purposes. 

For the interface, the only thing I would mention is that the user can still get it wrong by sending a `state_minus` that does not match the dimension of the configuration of the BC object.   It might be prudent to set self._dim = vel.shape[0], and then `assert` that it matches the input state at use time.   I leave that to your discretion.

PR #1017, opened 2024-03-08
scripts: move to FS-global / rank-local caches

matthiasdiener:
**Questions for the review**:


scripts/delta-parallel-spawner.sh
```diff
@@ -2,8 +2,8 @@
 #
 # Used to wrap the spawning of parallel mirgecom drivers on Delta.
 # unset CUDA_CACHE_DISABLE
-POCL_CACHE_ROOT=${POCL_CACHE_ROOT:-"/projects/bbkf/$USER/pocl-scratch"}
-XDG_CACHE_ROOT=${XDG_CACHE_HOME:-"/projects/bbkf/$USER/xdg-scratch"}
+POCL_CACHE_ROOT=${POCL_CACHE_ROOT:-"/scratch/bbkf/$USER/pocl-scratch"}
```
tulioricci, line 5: This is hard-coding to a specific project, which is not ideal.

matthiasdiener: Right, is there a way to 'get' the project, either via env variable or another command?

MTCam: > Right, is there a way to 'get' the project, either via env variable or another command?

I think something like this:
```
(omega.main) mtcampbe@Mikes-MacBook-Pro-2 examples % cat ../scripts/lassen-parallel-spawner.sh 
#!/bin/bash
#
# Used to wrap the spawning of parallel mirgecom drivers on Lassen
# unset CUDA_CACHE_DISABLE
POCL_CACHE_ROOT=${POCL_CACHE_ROOT:-"/tmp/$USER/pocl-scratch"}
XDG_CACHE_ROOT=${XDG_CACHE_HOME:-"/tmp/$USER/xdg-scratch"}
POCL_CACHE_DIR=${POCL_CACHE_DIR:-"${POCL_CACHE_ROOT}/rank$OMPI_COMM_WORLD_RANK"}
XDG_CACHE_HOME=${XDG_CACHE_HOME:-"${XDG_CACHE_ROOT}/rank$OMPI_COMM_WORLD_RANK"}
export POCL_CACHE_DIR
export XDG_CACHE_HOME

"$@"
```
is better.  It automatically sets it to `/tmp` which is available on all platforms, and also allows the user to override by setting `POCL_CACHE_ROOT` and `XDG_CACHE_ROOT`.

scripts/lassen-parallel-spawner.sh
```diff
@@ -2,8 +2,8 @@
 #
 # Used to wrap the spawning of parallel mirgecom drivers on Lassen
 # unset CUDA_CACHE_DISABLE
-POCL_CACHE_ROOT=${POCL_CACHE_ROOT:-"/tmp/$USER/pocl-scratch"}
-XDG_CACHE_ROOT=${XDG_CACHE_HOME:-"/tmp/$USER/xdg-scratch"}
+POCL_CACHE_ROOT=${POCL_CACHE_ROOT:-"/usr/workspace/wsa/$USER/pocl-scratch"}
+XDG_CACHE_ROOT=${XDG_CACHE_HOME:-"/usr/workspace/wsa/$USER/xdg-scratch"}
```
MTCam, line 6: I dislike this.  The default setting in /tmp is better than setting it to something some folks may not have access to.   It would be better, imo, to have the user set something that overrides POCL_CACHE_ROOT/XDG_CACHE_ROOT if they don't want node-local caching.  

If we want to invert that and make it so that the user must specify when they _want_ node-local caching, then that'd be ok too, but it should then just default to the CWD of the job or something like that, not a file path.

scripts/tioga.flux.sh
```diff
@@ -20,7 +20,7 @@ export PYOPENCL_CTX="AMD:0"
 
 run_cmd="flux run -N $nnodes -n $nproc --exclusive"
 
-export XDG_CACHE_ROOT="/tmp/$USER/xdg-scratch"
-export POCL_CACHE_ROOT="/tmp/$USER/pocl-cache"
+export XDG_CACHE_ROOT="/usr/workspace/wsa/$USER/xdg-scratch"
+export POCL_CACHE_ROOT="/usr/workspace/wsa/$USER/pocl-cache"
```
MTCam, line 24:
```suggestion
export POCL_CACHE_ROOT=${POCL_CACHE_ROOT:-"$(pwd)/pocl-scratch"}
export XDG_CACHE_HOME=${XDG_CACHE_HOME:-"$(pwd)/xdg-scratch"}
```

scripts/tioga-parallel-spawner.sh
```diff
@@ -2,8 +2,8 @@
 #
 # Used to wrap the spawning of parallel mirgecom drivers on Tioga.
 
-POCL_CACHE_ROOT=${POCL_CACHE_ROOT:-"/tmp/$USER/pocl-scratch"}
-XDG_CACHE_ROOT=${XDG_CACHE_HOME:-"/tmp/$USER/xdg-scratch"}
+POCL_CACHE_ROOT=${POCL_CACHE_ROOT:-"/usr/workspace/wsa/$USER/pocl-scratch"}
+XDG_CACHE_ROOT=${XDG_CACHE_HOME:-"/usr/workspace/wsa/$USER/xdg-scratch"}
```
MTCam, line 6:
```suggestion
POCL_CACHE_ROOT=${POCL_CACHE_ROOT:-"$(pwd)/pocl-scratch"}
XDG_CACHE_HOME=${XDG_CACHE_HOME:-"$(pwd)/xdg-scratch"}
```

tulioricci:
Could we easily check the cache path within the driver?

MTCam:
> Could we easily check the cache path within the driver?

Should a snippet be added to the mpi entry point to indicate the cache directories via stdout?  Since the cache directories potentially have an impact on performance, maybe it wouldn't be such a bad idea.

Will multiple jobs running at the same time potentially collide if they all indicate the same cache locations?

matthiasdiener:
> > Could we easily check the cache path within the driver?
> 
> Should a snippet be added to the mpi entry point to indicate the cache directories via stdout? Since the cache directories potentially have an impact on performance, maybe it wouldn't be such a bad idea.

What do you think of [cfc1af1](https://github.com/illinois-ceesd/mirgecom/pull/1017/commits/cfc1af164353b96a72bf05aec5093754072a1611)? 

The output looks like:

```
LOOPY_NO_CACHE= POCL_CACHE_DIR=foo python -m mpi4py examples/wave.py --numpy
[...]
Rank 0 disk cache config: loopy: False ; pyopencl: True (default dir); pocl: True (foo);
```
> 
> Will multiple jobs running at the same time potentially collide if they all indicate the same cache locations?

Yes, that can happen, in particular if you are running exactly the same case with the same runtime config at the same time.

matthiasdiener:
From an email by John Gyllenhaal on Mar 12, 2020:

> You can also override  the CUDA_CACHE_PATH  by setting it before running jsrun/lrun/srun but please do not set it to a NFS mounted directory (do NOT use a home directory, a workspace directory, or a gapps directory)!  

Looking at https://hpc.llnl.gov/documentation/tutorials/livermore-computing-resources-and-environment#file-systems and https://hpc.llnl.gov/documentation/tutorials/using-lc-s-sierra-systems#file-systems, probably `/p/gpfs1/$USER` may be a better location for the default disk cache storage?

MTCam:
> From an email by John Gyllenhaal on Mar 12, 2020:
> 
> > You can also override  the CUDA_CACHE_PATH  by setting it before running jsrun/lrun/srun but please do not set it to a NFS mounted directory (do NOT use a home directory, a workspace directory, or a gapps directory)!
> 
> Looking at [hpc.llnl.gov/documentation/tutorials/livermore-computing-resources-and-environment#file-systems](https://hpc.llnl.gov/documentation/tutorials/livermore-computing-resources-and-environment#file-systems) and [hpc.llnl.gov/documentation/tutorials/using-lc-s-sierra-systems#file-systems](https://hpc.llnl.gov/documentation/tutorials/using-lc-s-sierra-systems#file-systems), probably `/p/gpfs1/$USER` may be a better location for the default disk cache storage?

It is a little odd to default it opaquely to a platform-specific filesystem path.  The previous setting was `/tmp` and that's reasonable because it is platform-independent, node-local, and job-local.   When we were given space in the LC workspace, they also asked that we not install packages there that use conda because of the typical sizes of the base installs.   That includes MIRGE-Com, so we should also not be installing MIRGE-Com in those NFS places indicated above.

Where has `CUDA_CACHE_PATH` been defaulting to all this time? 

For most of us, and following system guidance, `$(pwd)` from a job script will be somewhere down in `/p/gpfs1/$USER` on lassen already, or on Lustre for quartz.   Nobody should be running parallel jobs from NFS mounted file spaces.  So I think defaulting to `$(pwd)` is also reasonable

scripts/tioga.flux.sh
```diff
@@ -20,7 +20,7 @@ export PYOPENCL_CTX="AMD:0"
 
 run_cmd="flux run -N $nnodes -n $nproc --exclusive"
 
-export XDG_CACHE_ROOT="/tmp/$USER/xdg-scratch"
-export POCL_CACHE_ROOT="/tmp/$USER/pocl-cache"
+export XDG_CACHE_ROOT=${XDG_CACHE_HOME:-"$(pwd)/xdg-scratch"}
```
matthiasdiener, line 23: Should we make `*-scratch` or `*-cache` the standard directory? Currently it's 50-50.

MTCam: I like cache myself, but meh.

matthiasdiener: renamed them to -cache in [92e227a](https://github.com/illinois-ceesd/mirgecom/pull/1017/commits/92e227a2a7aac2bd8e4c61495ec64626d518ce7d)

mirgecom/mpi.py
```diff
@@ -117,6 +117,8 @@ def _check_gpu_oversubscription() -> None:
     if size <= 1:
         return
 
+    # This may unnecessarily require pyopencl in case we run with a
+    # NumpyArrayContext or CupyArrayContext
     cl_ctx = cl.create_some_context()
     dev = cl_ctx.devices
```
inducer, line 123: Creating an OpenCL context is tremendously expensive, and doing so just to find the device to be used is not a good idea. Instead of this, there should be a central place in mirgecom that decides on the target device, and that should be used instead.

To support this, [`create_some_context` in pyopencl](https://github.com/inducer/pyopencl/blob/4e067bdd3ee06c164572c4e9ae398beee6b58663/pyopencl/__init__.py#L1555) should be refactored to separate device choice from context creation, if that device choice logic is intended to be retained.

matthiasdiener: Is https://github.com/inducer/pyopencl/pull/735 what you had in mind?

matthiasdiener: Note that #1030 goes a bit in a different direction - it centralizes actx creation in mpi_entry_point and removes all other calls to create_some_context.

scripts/delta.sbatch.sh
```diff
@@ -33,11 +33,11 @@ srun_cmd="srun -N $nnodes -n $nproc"
 # See
 # https://mirgecom.readthedocs.io/en/latest/running.html#avoiding-overheads-due-to-caching-of-kernels
 # on why this is important
-export XDG_CACHE_HOME_ROOT="/tmp/$USER/xdg-scratch/rank"
+export XDG_CACHE_HOME_ROOT="$(pwd)/xdg-cache/rank"
```
majosm, line 36: Small request: Could we make a parent directory for all the cache dirs (preferably hidden)? `.mirge-cache`?

.gitignore
```diff
@@ -48,3 +48,9 @@ test/nodal-dg
 # Emacs backups
 .\#*
 \#*
+
+
+# disk caches
+xdg-cache/
+pocl-cache/
+cuda-cache/
```
majosm, line 56: I'm forgetting if this applies to subdirs of the repo too as written, or if only the top level dir? (I sometimes run from `examples/`, so the former would be helpful.)

majosm (commented):
Couple of thoughts, otherwise looks good as far as I can tell.

scripts/delta.sbatch.sh
MTCam: I like this idea.  I would do it something like in the `lassen_parallel_spawner.sh` by defaulting `MIRGE_CACHE_ROOT=${MIRGE_CACHE_ROOT:-"$(pwd)/.mirge-cache"`.    Does that sound OK @majosm ?

.gitignore
matthiasdiener: It should apply to subdirs.

scripts/delta.sbatch.sh
matthiasdiener: Done in [dca839d](https://github.com/illinois-ceesd/mirgecom/pull/1017/commits/dca839dd58f4bff354d907f55c6226c536001c16).

scripts/tioga.flux.sh
```diff
@@ -20,7 +20,7 @@ export PYOPENCL_CTX="AMD:0"
 
 run_cmd="flux run -N $nnodes -n $nproc --exclusive"
 
-export XDG_CACHE_ROOT="/tmp/$USER/xdg-scratch"
-export POCL_CACHE_ROOT="/tmp/$USER/pocl-cache"
+export XDG_CACHE_ROOT=${XDG_CACHE_HOME:-"$(pwd)/.mirge-cache/xdg-cache"}
```
majosm, line 23: Should this be

```suggestion
export XDG_CACHE_ROOT=${XDG_CACHE_ROOT:-"$(pwd)/.mirge-cache/xdg-cache"}
```
?

scripts/delta.sbatch.sh
```diff
@@ -33,11 +33,11 @@ srun_cmd="srun -N $nnodes -n $nproc"
 # See
 # https://mirgecom.readthedocs.io/en/latest/running.html#avoiding-overheads-due-to-caching-of-kernels
 # on why this is important
-export XDG_CACHE_HOME_ROOT="/tmp/$USER/xdg-scratch/rank"
+export XDG_CACHE_HOME_ROOT="$(pwd)/.mirge-cache/xdg-cache/rank"
 
 # Fixes https://github.com/illinois-ceesd/mirgecom/issues/292
 # (each rank needs its own POCL cache dir)
-export POCL_CACHE_DIR_ROOT="/tmp/$USER/pocl-cache/rank"
+export POCL_CACHE_DIR_ROOT="$(pwd)/.mirge-cache/pocl-cache/rank"
```
majosm, line 40: Should this use `MIRGE_CACHE_ROOT`? (Same question for the other batch submitter scripts.)

matthiasdiener: These scripts are supposed to be standalone, so I think not?

majosm: Ah, gotcha.

MTCam: I think the batch script inherits the submission env, so could be a nice addtion here...

MTCam (approved):
👍

PR #1020, opened 2024-03-13
doc: fix MacOS kernel disk caching wrt XDG_CACHE_HOME

matthiasdiener:
See https://github.com/platformdirs/platformdirs/issues/4 and https://github.com/platformdirs/platformdirs/blob/7039c66b02f50fca54b965dcb2bd9de865d06cb8/src/platformdirs/macos.py#L56-L58 for context.

cc @MTCam , @inducer 

**Questions for the review**:

### DIFF
--- a/doc/running/caching.rst
+++ b/doc/running/caching.rst
@@ -1,11 +1,12 @@
 OpenCL kernel caching
 =====================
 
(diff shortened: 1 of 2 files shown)


MTCam (approved):
Thanks.  As good to know as it is irritating.

PR #1022, opened 2024-03-23
make cache path control slightly more convenient

MTCam:
This adds centralized setting of where mirgecom caches will go for Lassen, and overrides the default node-local system default setting of CUDA_CACHE_PATH.

User sets MIRGE_CACHE_ROOT to the path to locate all disk caches.  
Fine grained control is still available as before.

Co-authored-by: Mike Campbell <mtcampbe@illinois.edu>**Questions for the review**:


matthiasdiener (approved):
LGTM, thanks!

PR #1028, opened 2024-04-18
Allow limiter to optionally pass back p,T

MTCam:
Allow limiter to optionally pass back p, T.  This can break things if the limiter passes back invalid p, T , but it is a performance bottleneck to disallow it when we trust that the driver-implemented limiter is returning the correct p, T.

(This is not adding a new behavior, it is optionally allowing the old behavior)

**Questions for the review**:


anderson2981:
yup.

anderson2981 (commented):
yup.

anderson2981 (approved):

PR #1030, opened 2024-04-19
global actx init, clean up pyopencl usage

matthiasdiener:
(the other examples would need to be changed in the same way as wave)

**Questions for the review**:


mirgecom/mpi.py
```diff
@@ -287,12 +283,22 @@ def wrapped_func(*args, **kwargs) -> None:
         # exit
         from mpi4py import MPI  # noqa
 
-        _check_gpu_oversubscription()
-        _check_cache_dirs_node()
+        from mirgecom.array_context import initialize_actx
+        # This is where actual actx selection occurs
+        actx = initialize_actx(actx_class, comm=MPI.COMM_WORLD)
+        kwargs["actx"] = actx
```
matthiasdiener, line 289: This kind of worries me, but I'm not sure there is a better way to do this.

majosm: I'm getting the feeling that `mpi_entry_point` has outlived its usefulness... I wonder if it would be cleaner at this point to just split this into two function calls, e.g., add an `initialize_mpi` function and then put the rest inside `initialize_actx`?

matthiasdiener: Good idea, the only snag is that `initialize_mpi` also needs to know the actx_class, but not a big deal. What do you think of [bf9acf4](https://github.com/illinois-ceesd/mirgecom/pull/1030/commits/bf9acf4a02118cd938581f273c8694cbc3793494).

majosm: Hmm, I forgot about that. Passing `actx_class` to `initialize_mpi` seems a little odd. Would it make sense to rename it to something like `initialize_env` instead and call it unconditionally from the driver, e.g.:

```python
from mirgecom.array_context import actx_class_is_pyopencl
initialize_env(use_cl=actx_class_is_pyopencl(actx_class), use_mpi=mpi)
```

?

matthiasdiener: > ```python
> from mirgecom.array_context import actx_class_is_pyopencl
> initialize_env(use_cl=actx_class_is_pyopencl(actx_class), use_mpi=mpi)
> ```
I think this is also a bit clumsy, what do you think of [e833415](https://github.com/illinois-ceesd/mirgecom/pull/1030/commits/e8334150f37ea1b50bce12ecaceae99f84df9da9)?

matthiasdiener:
This is ready for review @MTCam @majosm . The production CI test fails due to some merge conflict.

mirgecom/mpi.py
majosm: Ehh, I dunno if I like that... it imposes hidden constraints on how the driver creator can set up their command line arguments. Better to be explicit IMO.

matthiasdiener: I dunno. The only other option that I can think of is to go back to the previous code in this location, i.e. running `cl.get_platforms()` unconditionally.

Note that the particular code here is almost completely irrelevant - the only reason for not running `cl.get_platforms()` in a numpy actx case is to avoid *pocl* from getting loaded, pyopencl will be loaded in for all actxs. There is no downside in running `cl.get_platforms()` (e.g., it will not ask for device selection or create a context). There is also no downside in loading pocl unncessarily.

majosm: How about
```diff
-    if args.mpi:
-        main_func = main
-    else:
-        import inspect
-        # run main without the mpi_entry_point wrapper
-        main_func = inspect.unwrap(main)    if args.mpi:
+    if args.mpi:
+        main_func = mpi_entry_point(using_cl=actx_class_is_pyopencl(actx_class))(main)
+    else:
+        main_func = main
```
?

matthiasdiener: I find it clumsy, and every driver needs to be updated to support it ...

majosm: [5aac0c8](https://github.com/illinois-ceesd/mirgecom/pull/1030/commits/5aac0c81bda420c4dc67b001747538ca201112ed) (unconditional loading) is OK with me if there's no real downside. @MTCam?

mirgecom/array_context.py
```diff
@@ -100,6 +110,152 @@ def actx_class_is_numpy(actx_class: Type[ArrayContext]) -> bool:
...
+                if len(all_paths) != len(set(all_paths)):
+                    hostname = MPI.Get_processor_name()
+                    dup = [path for path in set(all_paths)
+                                if all_paths.count(path) > 1]
+
+                    from warnings import warn
+                    warn(f"Multiple ranks are sharing '{var}' on node '{hostname}'. "
+                        f"Duplicate '{var}'s: {dup}.")
+
+        _check_var("XDG_CACHE_HOME")
```
majosm, line 153: Strictly speaking, don't we want `XDG_CACHE_HOME` to be unique even when not using pyopencl? Might be worth keeping the `XDG_CACHE_HOME` check inside `mpi_entry_point` and doing the rest here?

majosm (commented):
One minor thought, otherwise looks good on my end.

mirgecom/array_context.py
matthiasdiener: `XDG_CACHE_HOME` only matters for a mirgecom run if we use pyopencl, as far as I can tell.

majosm: Yeah, I wasn't sure what all goes in there. On the one hand I do see some loopy output when running with `--numpy` (clean cache):

```
% ls .cache/xdg-scratch/rank0/pytools 
pdict-v4-loopy-code-gen-cache-v3-2024.1-islpy2023.2.5-cgen2020.1-cc795034cc40d49deed34705ca594c96f407a70f-v1-py3.11.8.final.0
pdict-v4-loopy-invoker-cache-v10-2024.1-islpy2023.2.5-cgen2020.1-cc795034cc40d49deed34705ca594c96f407a70f-v1-py3.11.8.final.0
pdict-v4-loopy-memoize-cache-buffer_array-LoopyKeyBuilder.LoopyKeyBuilder-v0-2024.1-islpy2023.2.5-cgen2020.1-cc795034cc40d49deed34705ca594c96f407a70f-v1-py3.11.8.final.0
pdict-v4-loopy-memoize-cache-contract_arrays-LoopyKeyBuilder.LoopyKeyBuilder-v0-2024.1-islpy2023.2.5-cgen2020.1-cc795034cc40d49deed34705ca594c96f407a70f-v1-py3.11.8.final.0
pdict-v4-loopy-memoize-cache-fuse_same_discretization_entity_loops-LoopyKeyBuilder.LoopyKeyBuilder-v0-2024.1-islpy2023.2.5-cgen2020.1-cc795034cc40d49deed34705ca594c96f407a70f-v1-py3.11.8.final.0
pdict-v4-loopy-memoize-cache-preprocess_program-LoopyKeyBuilder.LoopyKeyBuilder-v0-2024.1-islpy2023.2.5-cgen2020.1-cc795034cc40d49deed34705ca594c96f407a70f-v1-py3.11.8.final.0
pdict-v4-loopy-schedule-cache-v4-2024.1-islpy2023.2.5-cgen2020.1-cc795034cc40d49deed34705ca594c96f407a70f-v1-py3.11.8.final.0
pdict-v4-loopy-typed-and-scheduled-cache-v1-2024.1-islpy2023.2.5-cgen2020.1-cc795034cc40d49deed34705ca594c96f407a70f-v1-py3.11.8.final.0
```
but I'm not sure if this should be happening (e.g., why is there fusion contractor stuff?).

matthiasdiener: Which case did you run?

majosm: combozzle

matthiasdiener: Hmm, any specific options you use? The cache is empty when I run locally.

matthiasdiener: Wait a sec, are there actually any files in the folders you mentioned above? If not your cache is empty, but the directory structure is created by loopy since it is still imported.

matthiasdiener: `find xdg-cache/ -type f` or so

majosm: Yeah, they're empty.

MTCam (approved):

PR #1031, opened 2024-04-23
Add more test conditions for chem kinetics

tulioricci:
- Split `test_eos` and move chemical kinetics tests to `test_chemistry`
- Add extra chemistry verification at different pressure conditions
- Add verification of species entropy and Gibbs energy (used under the hood in reversible reactions)
- Add lewis=1.0 test option to `test_transport`
-  Explicitly export the mechanism and read it back for better assessment of test coverage in `test_eos`, `test_chemistry` and `test_transport`
- Tried to remove deprecation warns from Pyrometheus but left as a comment instead. Need https://github.com/illinois-ceesd/drivers_y3-prediction/pull/53

**Questions for the review**:

### DIFF
--- a/test/test_eos.py
+++ b/test/test_eos.py
@@ -387,7 +387,7 @@ def test_pyrometheus_eos(ctx_factory, mechname, dim, y0, vel):
         yin = y0s * ones
 
         # avoid starting too far from the actual temperature
-        tguess = tempin + 100.0*np.random.random()
+        tguess = tempin + ones*100.0*np.random.random()
         assert tguess > 0.0
 
         pyro_rho = prometheus_mechanism.get_density(pin, tin, yin)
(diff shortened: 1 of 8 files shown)


test/test_eos.py
```diff
@@ -328,7 +339,7 @@ def inf_norm(x):
             assert inf_norm(conc[spec]) < 1e-14
 
 
-# TODO remove this test.. It is already covered in the other ones
+# FIXME This test is kinda redudant
```
MTCam, line 342: This comment goes from bad to worse.  Can you make it a little more clear to dummies?

tulioricci: Essentially, we have 3 tests:
1) pyrometheus vs cantera
2) mirge-com vs pyrometheus
3) mirge-com vs cantera

My original suggestion was to drop this out. But keeping it is also fine. So either I remove the comment or the test.

MTCam: I would just leave the tests, then.  It is not entirely clear there is real redundancy here without more details.   I can see motivation for making sure that our massage of Pyro into mirgecom didn't break pyro, that mirgecom gets the same answers as pyro when using mirgecom API, and that mirgecom matches Cantera when we do a dynamic problem.

tulioricci: :+1:

MTCam (approved):

PR #1035, opened 2024-05-10
replace nodal to modal transformation with an element-wise integral

anderson2981:
**Questions for the review**:


tulioricci:
Is it improving performance? I can't remember exactly but I used the modal implementation because it was better (at the time) then using what I believe to be the built-in average.

anderson2981:
> Is it improving performance? I can't remember exactly but I used the modal implementation because it was better (at the time) then using what I believe to be the built-in average.

It appears to, yes. The modal to nodal seems to invoke a large compile time penalty. This is part of the reason that we broke the filtering out of the main compiled rhs function. I haven’t tested performance yet for simplicies. For my case with TPE quads, the compile time went from 4000 s to 1000 s on my mac. The time per timestep went from 2.1 to 1.1 s.

MTCam (commented):
Any chance you could add your new limiter tests before we merge?

anderson2981:
> Any chance you could add your new limiter tests before we merge?

They should be live in y3prediction@main.

MTCam (approved):

PR #1036, opened 2024-05-12
add functionality to get temperature from enthalpy

anderson2981:
**Questions for the review**:


anderson2981:
probably could add some tests to check that enthalpy matches with temperature and energy.

Should also be able to refactor into 1 function that does both, like in pyrometheus.

MTCam:
> probably could add some tests to check that enthalpy matches with temperature and energy.
> 
> Should also be able to refactor into 1 function that does both, like in pyrometheus.

Cool, just added #1037 to align this with mixture fraction eos changes.  Unified energy/enthalpy interface and update func.

MTCam (approved):

PR #1038, opened 2024-05-13
retire POCL_CACHE_DIR

matthiasdiener:
pocl will use XDG_CACHE_HOME if it is set

**Questions for the review**:


scripts/tioga-parallel-spawner.sh
```diff
@@ -1,20 +1,13 @@
 #!/bin/bash
 #
 # Used to wrap the spawning of parallel mirgecom drivers on Tioga.
-export CUDA_CACHE_DISABLE=0
```
matthiasdiener, line 4: Tioga has AMD GPUs

MTCam (approved):
This seems good to me as long as tests pass.

PR #1041, opened 2024-05-15
Add function to compute mole fractions from mass and vice-versa

tulioricci:
**Questions for the review**:


MTCam (approved):

MTCam:
I don't think the interface is descriptive enough to be readily understood by other users.   Not a reason to block it.  The only thing I could think of to "improve" it would be:

`mass_fractions_from_mole_fractions` and `mole_fractions_from_mass_fractions`

This is readily understood at the call site, but a bit more cumbersome.   You can fix this or leave it.

PR #1051, opened 2024-08-09
Remove usage of some deprecated/undocumented functionality

majosm:
**Questions for the review**:
  - [ ] ~~The PR should have a description.~~
  - [ ] ~~The PR should have a guide if needed (e.g., an ordering).~~
- [ ] ~~Is every top-level method and class documented? Are things that should be documented actually so?~~
- [ ] ~~Is the interface understandable? (I.e. can someone figure out what stuff does?) Is it well-defined?~~
- [ ] ~~Does the implementation do what the docstring claims?~~
- [ ] ~~Do you see any immediate risks or performance disadvantages with the design? Example: what do interface normals attach to?~~

### DIFF
--- a/test/test_av.py
+++ b/test/test_av.py
@@ -36,6 +36,7 @@ from arraycontext import pytest_generate_tests_for_array_contexts
 
 from meshmode.mesh import BTAG_ALL
 from meshmode.discretization.connection import FACE_RESTR_ALL
+import grudge.geometry as geo
 import grudge.op as op
 from mirgecom.artificial_viscosity import (
     av_laplacian_operator,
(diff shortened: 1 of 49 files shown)


majosm:
Dropped the removal of https://github.com/illinois-ceesd/mirgecom/blob/03860e762c3d89cc96c856dfd2ac7885eb4d6436/mirgecom/array_context.py#L326 for now. It can't be removed until the default value changes to `True` in inducer/arraycontext#278. Will submit this as a separate PR later.

majosm:
Ready for a look. Might be easiest to read commit by commit to see the explanation for each change?

test/test_av.py
```diff
@@ -86,7 +87,7 @@ def test_tag_cells(ctx_factory, dim, order):
     queue = cl.CommandQueue(cl_ctx)
     actx = PyOpenCLArrayContext(queue)
 
-    nel_1d = 2
+    nel_1d = 1
```
MTCam, line 90: 🤔 Not complaining, just curious.

test/test_av.py
```diff
@@ -111,9 +112,9 @@ def norm_indicator(expected, dcoll, soln, **kwargs):
 
     # get meshmode polynomials
     group = dcoll.discr_from_dd("vol").groups[0]
-    basis = group.basis()  # only one group
+    basis = group.basis_obj().functions  # only one group
     unit_nodes = group.unit_nodes
-    modes = group.mode_ids()
+    modes = group.basis_obj().mode_ids
```
MTCam, line 117: Not sure if it matters here, but this `basis_obj()` change appears to fail for quadrature discretizations.

MTCam (approved):
lgtm!

test/test_av.py
majosm: The old `generate_regular_rect_mesh` argument `n` specified the number of mesh vertices along a given axis, so the test was creating a mesh with 2 vertices, or 1 element. Now it does the same thing but with `nelements_per_axis`.

Not sure whether the test was originally intended to have 1 or 2 elements. At the moment it fails if I try to use a mesh with more than 1 element (whether passing `n` or `nelements_per_axis`).

test/test_av.py
majosm: Hmm... Can you explain more? Not sure if I follow what's failing.

majosm:
Added a few more changes that I missed the first time around @MTCam (including a FIXME relating to a limiter test issue that I noticed). Previous review was up through the commit labeled "remove unused(?) TestSolution".

MTCam:
> Added a few more changes that I missed the first time around @MTCam (including a FIXME relating to a limiter test issue that I noticed). Previous review was up through the commit labeled "remove unused(?) TestSolution".

Still looks great to me.  Thanks for tackling all this.

PR #1053, opened 2024-08-14
prepare for running upstream CI

matthiasdiener:
**Questions for the review**:


examples/run_examples_short_ci.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -o nounset
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+$SCRIPT_DIR/run_examples.sh $SCRIPT_DIR/examples wave.py
```
matthiasdiener, line 7: Which examples should we run

PR #1060, opened 2024-08-26
DeviceMemoryUsage: add AMD support, use ProcessLogger for version.sh

matthiasdiener:
**Questions for the review**:

### DIFF
--- a/mirgecom/array_context.py
+++ b/mirgecom/array_context.py
@@ -5,6 +5,7 @@
 .. autofunction:: actx_class_is_eager
 .. autofunction:: actx_class_is_profiling
 .. autofunction:: actx_class_is_numpy
+.. autofunction:: actx_class_is_distributed
 .. autofunction:: initialize_actx
 """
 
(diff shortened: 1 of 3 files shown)


mirgecom/array_context.py
```diff
@@ -176,6 +180,10 @@ def _check_gpu_oversubscription(actx: ArrayContext) -> None:
     Only works with CUDA devices currently due to the use of the
     PCI_DOMAIN_ID_NV extension.
     """
+    import sys
+    if "mpi4py" not in sys.modules:
+        return
+
```
majosm, line 186: Would it make sense to add an `actx_class_is_distributed` or `actx_class_is_mpi` function and only call `_check_cache_dirs_node`/`_check_gpu_oversubscription` if it returns `True`?

mirgecom/array_context.py
```diff
@@ -227,11 +235,16 @@ def _check_gpu_oversubscription(actx: ArrayContext) -> None:
...
 
     assert isinstance(actx, (PyOpenCLArrayContext, PytatoPyOpenCLArrayContext))
 
-    rank = MPI.COMM_WORLD.Get_rank()
+    if "mpi4py" in sys.modules:
+        from mpi4py import MPI
+        rank = MPI.COMM_WORLD.Get_rank()
+    else:
+        rank = 0
+
```
majosm, line 247: Can you check if it's an MPI array context and get `actx.mpi_communicator`?

majosm (approved):
Couple of minor suggestions, otherwise LGTM.

mirgecom/array_context.py
matthiasdiener: Thanks! What do you think of [16a3a6a](https://github.com/illinois-ceesd/mirgecom/pull/1060/commits/16a3a6a9560db47c01b18bcac13ad366a10f8c04)?

mirgecom/array_context.py
matthiasdiener: Done in [16a3a6a](https://github.com/illinois-ceesd/mirgecom/pull/1060/commits/16a3a6a9560db47c01b18bcac13ad366a10f8c04)

PR #1072, opened 2024-10-14
bump Python to 3.12

matthiasdiener:
**Questions for the review**:


matthiasdiener:
Looks like coverage is our problem regarding 3.12, and it turns out 3.12 is even a bit faster than 3.11 with coverage disabled.

MTCam (approved):

PR #1102, opened 2025-04-01
make actx_class_has_fallback_args a bit broader

matthiasdiener:
**Questions for the review**:


MTCam:
> ## Pull Request Overview
> This PR broadens the detection of fallback arguments in the actx_class_has_fallback_args function. It extends the check to include keyword-only arguments in addition to positional arguments.
> 
> * Extended fallback argument detection to cover keyword-only args.
> * Updated the return logic accordingly.

Does it meet all the requirements for merge?

